PR #53, opened 2026-05-18
chore(deps): bump symphonia 0.5 → 0.6 + harden migration checksums against line-ending drift

InstaZDLL:
## Summary

Three related changes bundled together — they fell out of investigating why dependabot PR #47 (`chore: bump symphonia from 0.5 to 0.6`) was failing CI:

- **symphonia 0.5 → 0.6 migration** (3 commits, split for review clarity)
- **`db::migration_heal`** — new self-healing layer for sqlx checksum drift from `core.autocrlf` regressions on Windows
- **Volume slider** — drag was being hijacked by the WebView's native drag fallback (separate, surfaced during manual testing)

Closes #47 (dependabot PR — version bump alone, no migration). Will close that PR once this lands.

## symphonia 0.6 migration

symphonia 0.6 ships a "Redesigned all audio primitives" release. The two files that touch the decoder API needed manual changes — `SampleBuffer` was removed, `core::probe` moved under `core::formats`, `Decoder` split into `AudioDecoder` etc. Full list in the commit message of `chore(deps): bump symphonia from 0.5 to 0.6`.

Added the `all-meta` feature so probe scoring keeps ID3/APE/Vorbis tag detection (tag reading still goes through lofty).

## `db::migration_heal`

Recurring symptom on this branch:

> `migration <id> was previously applied but has been modified`

Investigation: 4 of 5 stored checksums in `app.db._sqlx_migrations` matched the LF (`.gitattributes` canonical) hash of their migration file, 1 matched the CRLF variant. The CRLF row was the most recent migration — applied during a brief window where Git for Windows' `core.autocrlf=true` (system scope) leaked CRLF on first checkout, before `.gitattributes` normalized the working tree back to LF. The stored hash never updated, so every subsequent boot panicked.

The new module runs before each `Migrator::run`. For each stored row whose checksum doesn't match the compiled-in migration, it recomputes SHA-384 over the LF and CRLF variants of the embedded SQL; if either matches, it silently rewrites the row to the canonical hash with a `tracing::warn`. A real SQL change still panics — neither normalization can rescue it.

Confirmed live: at the next boot the heal hook logged `healed=1` and `Migrator::run` proceeded without manual DB surgery.

## Volume slider fix

Without `preventDefault` on `pointerdown` plus `select-none` on the track, the WebView interprets the gesture as a text/image drag, steals the pointer-event stream, flips the cursor to "no-drop", and freezes the slider. Surfaced during manual symphonia-validation testing.

## Test plan

- [x] `cargo check --all-targets`
- [x] `cargo test` — 97 passed, 0 failed (93 pre-existing + 4 new for `migration_heal`)
- [x] `cargo clippy --all-targets --no-deps` — clean
- [x] Manual playback: MP3, FLAC, M4A (AAC), seek, crossfade, ReplayGain, spectrum visualizer, volume slider drag
- [x] Live verification of `db::migration_heal` on a real CRLF-poisoned `app.db`

## Notes de publication

* **Améliorations**
  * Amélioration de la gestion des événements pointeur dans le contrôle de volume pour une meilleure stabilité des interactions.
  * Robustesse accrue des migrations de base de données avec normalisation automatique des fins de ligne.

* **Mises à jour des dépendances**
  * Mise à jour de la bibliothèque de traitement audio vers la version 0.6.
  * Mise à jour de Vite en version 8.x.

* **Documentation**
  * Documentation moteur de lecture mise à jour.

[Review Change Stack](https://app.coderabbit.ai/change-stack/InstaZDLL/WaveFlow/pull/53?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


InstaZDLL:
@coderabbitai Findings rejected after verification against the actual installed `symphonia-core-0.6.0` source. Details below for the next reviewer.

### Finding 1 (Critical) — `Time::try_from_secs_f64` allegedly missing

Verified in `~/.cargo/registry/.../symphonia-core-0.6.0/src/units.rs:580`:

```rust
pub fn try_from_secs_f64(total_secs: f64) -> Option<Time> { ... }
```

The method exists. Additional evidence:
- `cargo check --all-targets` ✓
- `cargo clippy --all-targets --no-deps` ✓ (and clippy itself suggested `unwrap_or_default()` on this exact call — impossible if the method didn't exist)
- Seeking was manually tested on FLAC / MP3 / M4A.

### Finding 2 (Critical) — `default_track` allegedly takes no argument

Verified in `~/.cargo/registry/.../symphonia-core-0.6.0/src/formats/mod.rs:630`:

```rust
fn default_track(&self, track_type: TrackType) -> Option<&Track> { ... }
```

The method takes a `TrackType`. The current code wouldn't compile without it. The Context7-fetched canonical example (https://context7.com/pdeljanov/symphonia/llms.txt) also uses this exact form:

```rust
let track = format.default_track(TrackType::Audio).unwrap();
```

### Finding 3 — `decoder.decode` error matching pre-existing

The error-matching arms around `decoder.decode` in `analysis.rs` are byte-for-byte identical to the pre-migration 0.5 code — this PR didn't touch them. Treating IoError-at-decode as fatal during a one-shot analysis pass is defensible (unrecoverable read mid-decode) and unrelated to the symphonia 0.6 migration. Hardening the pattern is a worthwhile follow-up but out of scope here.

Both Critical findings appear to be hallucinated API shapes — possibly from a pre-release symphonia 0.6 doc snapshot. Worth flagging upstream.

PR #61, opened 2026-05-18
perf(library): prefetch inactive tabs so switching feels instant

InstaZDLL:
## Summary

User-reported pain: clicking **Artistes / Albums / Morceaux** for the first time leaves the user staring at a blank tab for 1–2 s while the backend runs the corresponding `list_*` query and the JSON crosses Tauri's IPC bridge. Subsequent visits to the same tab were already instant — each tab's state array persists across `activeTab` switches — but the cold-cache first click was painful.

## Approach

Reshape the loading logic in [LibraryView](src/components/views/LibraryView.tsx) so the user only pays the fetch cost once per cache-invalidation, regardless of which tab they click first:

1. **`tabCacheKeys` memo** — derives one stable key per tab from the inputs that legitimately invalidate its data (library set, sort, `filterNoCover`, cover-reload signal, edit signal).
2. **`loadedKeysRef`** — records which key currently lives in each tab's state array. The foreground loader short-circuits when the active tab already matches its desired key (e.g. the user clicked back to a tab that just got prefetched).
3. **Background prefetch effect** — waits 200 ms after the active tab settles, then *serially* fetches each non-active tab whose stored key is stale. Serial (not parallel) so a large `list_tracks` query doesn't fight the active tab for SQLite pool slots.
4. **`fetchTab` / `applyTab` helpers** — centralise the per-tab switch so the two paths can't drift apart on the next refactor.

## Net effect

- The very first foreground tab loads exactly as before.
- Every subsequent tab switch renders the cached data **instantly**.
- A silent refresh runs in the background only when the cache key has actually invalidated (sort change, library change, edit, etc.).
- Failures in the background path are swallowed and fall back to the foreground loader on user click — no regression on the worst case.

## Test plan

- [x] `bun run typecheck`
- [x] `bun run lint`
- [ ] Cold launch → first tab (Songs/Tracks) loads normally → click Albums within 200 ms+~query time → instant render
- [ ] Change sort on Albums → background prefetch refreshes other tabs (verify by switching back to Songs and checking the data picks up the same `editRefetch`/sort changes)
- [ ] Edit a track in Library → all tabs visible data refreshes within ~200 ms after the foreground load

## Summary by CodeRabbit

* **Refactor**
  * Optimisé le chargement des onglets de la bibliothèque avec des mécanismes de mise en cache améliorés et une pré-récupération des données en arrière-plan, offrant une navigation par onglets plus fluide et réactive.

[Review Change Stack](https://app.coderabbit.ai/change-stack/InstaZDLL/WaveFlow/pull/61?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/components/views/LibraryView.tsx
+++ b/src/components/views/LibraryView.tsx
@@ -2,6 +2,7 @@ import {
   useCallback,
   useEffect,
   useLayoutEffect,
+  useMemo,
   useRef,
   useState,
 } from "react";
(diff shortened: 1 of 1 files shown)


InstaZDLL:
Reverting — I misread the original complaint. Daisy said the tab itself opened instantly and only the artwork flickered for 1-2 s. My prefetcher kept SQLite busy in the background, so when the user clicked a new tab the foreground fetch queued behind a slow list_tracks/etc. and the perceived lag got *worse*. Closing without merging and tackling the actual issue (artwork load smoothness) in a separate PR.

PR #117, opened 2026-05-23
feat(lyrics): auto-discover local sidecar .lrc / .txt files (#115)

### LINKED ISSUE #115: feat: Better support for local Lyrics files (.Lrc / .Txt ) and auto view them.
### What problem does this solve?

Hi, 
me and many users do have local Lyrics files, in music folder , along side song file , will be great if WaveFlow will support it and view them auto ,if they are in music folders , right now they do not view... (so will not need to search on line ,if they already localey)

the files name ,same as music file name , like this:
01 song name .mp3
01 song name .lrc
01 song name .txt

image of music folder:

i know have option in WaveFlow to add menuly a lyric file (and it work fine) but very time consumin for large collaction .... to view them auto will be much more easier and user frendly .

Also , i will like to ask to add support for sub folder  name "Lyrics" / "lyrics" this will help users to keep the folders more clean and orginize (in this folder the lyrics files are name same as music file ...just keep together in 1 sub folder, i think is a great way to help users keep there music folder clean .

 
thank you.
  

### Proposed solution

.

### Alternatives considered

_No response_

### Rough scope

Small (one card / one command / one menu entry)

### Anything else?

_No response_

InstaZDLL:
Closes #115.

## Summary

Many users keep their lyrics as sidecar files next to the audio (\`01 Song.mp3\` + \`01 Song.lrc\` / \`.txt\`) or in a sibling \`Lyrics/\` subfolder — a very common K-Pop / J-Pop rip layout. Until now those files were invisible to WaveFlow unless the user pressed "Import File" per track, which doesn't scale beyond a handful of songs.

Slot a new tier between the embedded-tag read and the LRCLIB network call in the [\`fetch_lyrics\`](src-tauri/src/commands/lyrics.rs) waterfall.

### New waterfall

1. DB cache
2. Embedded tag (USLT / TXXX / Vorbis / MP4 ©lyr)
3. **NEW — Sidecar file**
4. LRCLIB

### Sidecar lookup rules

- Probes \`{audio_dir}/{stem}.{lrc|txt}\` (same folder, primary)
- Then \`{audio_dir}/Lyrics/{stem}.{lrc|txt}\` (sibling subfolder)
- Case-insensitive at every layer (subfolder name, stem, extension) so \`Song.MP3\` matches \`song.lrc\` on Linux and \`lyrics/\` matches \`Lyrics/\`
- \`.lrc\` wins over \`.txt\` at every probed directory because it carries timing info
- Same-folder hits beat \`Lyrics/\` hits when both exist
- Whitespace-only files fall through (no empty cache row pollution)

The library-wide [\`prefetch_library_lyrics\`](src-tauri/src/commands/lyrics.rs) walks the same new waterfall, so the "Prefetch lyrics" button now also surfaces local sidecars without network traffic.

### Why \`save_lyrics\` was left alone

\`save_lyrics\` keeps writing only to the embedded tag. The sidecar file stays read-only on purpose: the file's authoritative copy becomes the embedded one after any in-app edit (the new file hash invalidates the cache row, the next fetch sees the embedded tag, the sidecar is silently superseded). This avoids two confusing failure modes:

- Writing back to \`Song.lrc\` would surprise users who maintain those files with external tools
- Writing both creates a divergence problem the next time they edit one

### Backend
- [x] 9 new unit tests added in \`commands::lyrics::tests\` (all 17 tests in the module pass)
- [x] \`cargo check --all-targets\` ✅

### Frontend
- [x] \`bun run lint\` ✅
- [x] \`bun run typecheck\` ✅

### Manual
- [x] Drop a \`01 Track.lrc\` next to a \`.mp3\` track with no embedded lyrics → open the lyrics panel → synced lyrics appear without clicking Import
- [x] Replace the \`.lrc\` with a same-named \`.txt\` → "Refetch" (clears cache) → reopen panel → plain text appears
- [x] Put both at the same time → \`.lrc\` is preferred
- [x] Move the sidecar into a \`Lyrics/\` subfolder next to the audio → "Refetch" → still found
- [x] On Linux: try \`lyrics/\` (lowercase) — also found
- [x] On Linux with \`Song.MP3\` + \`song.lrc\` (case mismatch) — found
- [x] No sidecar, no embedded, online → LRCLIB still consulted as before (sidecar tier is invisible to the existing fallback)
- [x] No sidecar, no embedded, offline → returns None (sidecar runs even in offline mode since it's local I/O — confirmed)
- [x] Edit lyrics via the in-app editor → save → reopen panel → updated content shown (from embedded tag, sidecar superseded)

## Docs

- [\`CLAUDE.md\`](CLAUDE.md) playback catalogue line updated to mention sidecar auto-discovery
- [\`docs/features/integrations.md\`](docs/features/integrations.md) LRCLIB section now lists the four tiers with the sidecar rules and the save-back caveat

## Notes de version

  * Recherche et prise en charge des fichiers de paroles sidecar (.lrc/.txt) à côté des pistes ou dans un dossier Lyrics (recherche insensible à la casse, priorité .lrc > .txt, ignore les répertoires homonymes).

* **Comportement**
  * Sidecar vide/whitespace est rejeté; lecture sidecar court-circuite la requête réseau et met le cache à jour avec source=LrcFile.
  * Sauvegarde n’écrit que l’embedded; un nouvel embedded supplante silencieusement l’ancienne sidecar en cache.

* **Documentation**
  * Ajout du niveau “Sidecar” et précisions sur priorités, cache et import_lrc_file (écrasement du cache).

* **Tests**
  * Tests unitaires couvrant découverte, priorité, casse et fichiers vides.

[Review Change Stack](https://app.coderabbit.ai/change-stack/InstaZDLL/WaveFlow/pull/117?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


InstaZDLL:
Applied in 78d9af6. Real edge case caught — a directory named `Song.lrc` would shadow a valid `Song.txt` sidecar in the same folder (the directory got selected into `lrc_match`, `read_to_string` failed, and the fallback was silently lost).

Fix: `if !path.is_file() { continue; }` at the top of the loop. `is_file` follows symlinks so a symlinked sidecar (one master `lyrics/` folder linked from album directories) still works.

Added a regression test `sidecar_skips_directory_named_like_a_sidecar` that reproduces the original failure and asserts the fallback wins — all 18 `commands::lyrics` tests pass.

InstaZDLL:
Applied in 49f8d8a. Real bug — an empty `Song.lrc` (common in low-quality rips that ship stub files) would short-circuit the waterfall via the `lrc_match.or(txt_match)` shortcut, the trimmed read returned None, and a valid `Song.txt` next to it was silently lost (so we'd fall through to LRCLIB unnecessarily).

Fix: try each candidate individually, fall back to `.txt` only if the `.lrc` read is empty. Extracted the read+trim+is_empty dance into a `read_non_empty_file` helper so both candidates share the same logic and stay symmetrical.

New regression test `sidecar_empty_lrc_falls_back_to_txt` reproduces the failure and asserts the txt wins. All 19 `commands::lyrics` tests pass.

PR #169, opened 2026-05-26
refactor(workspace): extract waveflow-core crate (Phase 1.a)

### LINKED ISSUE #128: Phase 1.a — Extract waveflow-core crate
**Tracking issue for sub-phase 1.a of [RFC-001](../blob/main/docs/rfcs/RFC-001-waveflow-server.md).**

Extract a `waveflow-core` workspace crate from `src-tauri/src/` so the same business logic can be reused by `waveflow-server` (Phase 1.b+) without forking the codebase. Pure refactor — zero behavior change for the desktop app.

## Scope

- Convert `src-tauri/` into a Cargo workspace (`src-tauri/Cargo.toml` becomes workspace manifest).
- Create `src-tauri/crates/core/` with:
  - Scanner pipeline (BLAKE3 hashing, lofty tag parsing, multi-artist split, album grouping)
  - Metadata clients (Deezer, Last.fm) — already use `reqwest` + `rustls-tls`
  - Audio analysis (ReplayGain, BPM, key detection)
  - DSD → PCM converter ([src-tauri/src/audio/dsd/](../tree/main/src-tauri/src/audio/dsd))
  - Domain types (Track, Album, Artist, Playlist, etc.)
- Define repository traits (`trait TrackRepository`, `trait PlaylistRepository`, …) parameterized over a connection type so SQLite (desktop) and Postgres (server) can plug in their own impls later.
- Move desktop-specific code (Tauri commands, media controls, DLNA, mini-player) to `src-tauri/crates/app/`.
- Run `cargo check --all-targets` + `cargo test` clean from the workspace root.
- Ship a tag-bumped desktop release to validate nothing regressed.

## Out of scope

- Postgres impl of the repository traits (that's 1.b).
- Splitting `waveflow-core` into its own repo (deferred until API stabilizes; see RFC §5).

## Acceptance criteria

- [ ] `cargo build --release` produces the same desktop binary footprint (± 5%) and the app boots normally.
- [ ] Existing test suite passes from the workspace root.
- [ ] CI updated to run `cargo check`/`cargo test` against the workspace, not just the legacy crate.
- [ ] Documentation: `CLAUDE.md` + [docs/architecture/](../tree/main/docs/architecture) updated to describe the new layout.

## Estimate

~3 weeks. Treat as research-grade work — if the extraction reveals tight coupling that needs a larger refactor, scope-creep is acceptable. Everything downstream depends on this being clean (see RFC §9 Risks).

InstaZDLL:
## Summary

Phase 1.a of [RFC-001](../issues/128): turn the monolithic `src-tauri/` package into a Cargo workspace with two members and lift every portable bit of business logic into `waveflow-core` so the future `waveflow-server` can consume the exact same code without going through the Tauri runtime.

- **`crates/core/` (`waveflow-core`)** — domain DTOs, repository traits + SQLite impls, scanner helpers + upserts, smart-playlist engine, audio analysis, DSD parsers + 1-bit→PCM conversion, artwork pipeline (thumbnails + shared metadata cache), HTTP clients (Deezer / Last.fm / LRCLIB), `CoreError`. **Zero `tauri` / `cpal` deps.**
- **`crates/app/` (`waveflow`)** — Tauri 2 application. Holds every `#[tauri::command]` (now thin wrappers over the core traits), the real-time `cpal` + `rtrb` audio engine, the DLNA / OS-media-controls / Discord-RPC integrations, the fs watcher, the tray, the profile pool wiring, paths.

Closes #128.

## How it's organised

15 commits, one logical step per commit so the diff stays reviewable:

1. `chore(workspace)` — virtual workspace skeleton
2. `refactor(core)` — domain types
3. `refactor(core)` — generic error type
4. `refactor(core)` — HTTP clients (deezer/lastfm/lrclib)
5. `refactor(core)` — `ProfileRepository`
6. `refactor(core)` — `LibraryRepository`
7. `refactor(core)` — `PlaylistRepository`
8. `refactor(core)` — `TrackRepository`
9. `refactor(core)` — artwork pipeline
10. `refactor(core)` — audio analysis
11. `refactor(core)` — smart-playlist engine (+ `PathsContext`)
12. `refactor(core)` — scanner pure helpers + upserts
13. `refactor(core)` — DSD pipeline
14. `docs(architecture)` — `docs/architecture/crates.md` + CLAUDE.md + bulk path updates + `--workspace` flag in CI + `.coderabbit.yaml` retargeted at the new layout
15. `fix(review)` — 3 CodeRabbit findings introduced or owned by this PR

Stats: **158 files changed, +5 657 / -4 296**. Bulk of the diff is `git mv` + one-line re-export shims to keep `crate::*` imports resolving in the app crate.

## What stayed in `crates/app/`

Anything tied to Tauri, cpal, the desktop OS or the real-time audio path. Specifically: `#[tauri::command]` handlers (even when the body is a thin call into core), the cpal callback + WASAPI exclusive thread (hot path, no allocs/locks/logs), souvlaki bridge, Discord RPC IPC client, native notifications, the DLNA worker, the fs watcher, `AppPaths`, the tray, the single-instance lock, the updater, the mini-player WebviewWindow.

Full split rules in `docs/architecture/crates.md`.

## Validation

- `cargo check --workspace --all-targets --manifest-path src-tauri/Cargo.toml` clean
- `cargo test --workspace --manifest-path src-tauri/Cargo.toml` → **111 passed / 0 failed** (66 in app + 45 in core)
- `bun run typecheck` / `bun run lint` clean (frontend untouched, sanity check only)

**Caveat:** validated on Linux only. Windows / macOS bundling (`bun run tauri build`) and the per-target audio paths (WASAPI exclusive, MediaRemote) are unchanged in shape so should be unaffected, but a green CI matrix on this PR is the real signal.

## Documented follow-ups (intentionally not in this PR)

These were called out by CodeRabbit and verified valid, but they're pre-existing in code the refactor merely moved; bundling them here would balloon the diff and risk regressions in paths Phase 1.a is supposed to leave alone:

- `scan_folder_inner` Tauri decoupling via a `ScannerEventSink` trait (so the orchestrator can move to core too)
- `ArtworkRepository` (artwork table CRUD)
- `AppError` legacy variants cleanup (collapse duplication once core owns Database/Io)
- Move `src-tauri/migrations/**` to `crates/core/migrations/**`
- Per-file `file_hash` recompute in `edit::update_track_tags` / `update_track_cover` / `save_lyrics` / `set_track_rating` (CLAUDE.md rule already exists)
- Offline-mode guards on the Spotify command surface
- DLNA `cds.rs` BrowseMetadata column-alias bug + `serve_art` backslash traversal check
- Misc perf wins (batched DELETEs, blocking-pool offloads, JOIN-not-N+1)

If any of those should block the merge, flag them and I'll add commits on the same branch.

## Test plan

- [ ] CI matrix green on Linux / Windows / macOS
- [ ] `bun run tauri dev` boots, library scan completes, playback works, profile switch works
- [ ] DLNA opt-in still bound (Settings → Integrations)
- [ ] Smart playlist regen ("Régénérer" on Home) materialises Daily Mix slots + On Repeat
- [ ] Lyrics waterfall (cache → embedded → sidecar → LRCLIB) on a previously-unlooked-up track

## Summary by CodeRabbit

* **New Features**
  * Scan de la bibliothèque plus fiable : import parallèle, progression en temps réel, prise en charge DSD et rescan des images d’artistes locales.

* **Improvements**
  * Recherches et consultation de pistes enrichies (filtres avancés, tri, likes, notes).
  * Playlists intelligentes stabilisées et régénération améliorée.
  * Gestion des profils améliorée (création, renommage, suppression, export/import).

* **Documentation**
  * Docs et guides mis à jour pour la nouvelle organisation du projet.

* **Chores**
  * Wrapper CLI Tauri, ajustements CI/workflows, mises à jour de dépendances et nettoyages d’artefacts.

[Review Change Stack](https://app.coderabbit.ai/change-stack/InstaZDLL/WaveFlow/pull/169?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


InstaZDLL:
Other findings memo :
Fix the following issues. The issues can be from different files or can overlap on same lines in one file.

- Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

In @src-tauri/crates/app/Cargo.toml around lines 208 - 219, Update the outdated comment that says "zip 4.x" to reflect the actual dependency version "zip 8.x" next to the zip = { version = "8", default-features = false, features = ["deflate", "time"] } declaration; edit the comment text around that dependency in Cargo.toml so it accurately says "zip 8.x" (or "zip 8") and retains the note about deflate and pure-Rust backend.

- Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

In @src-tauri/crates/app/src/audio/analytics.rs around lines 209 - 227, The code issues two separate SQL queries to fetch album_id for current_id and track.id; replace them with one query using WHERE id IN (?, ?) against the same pool and bind both IDs in order, fetch the two results (e.g., with query_and_then or fetch_all) and assign to current_album and next_album (Option<i64>) before computing same_album; update the block that currently sets current_album/next_album and same_album so it executes one sqlx query on pool, extracts the two optional album_ids mapped to the correct IDs, and preserves the existing same_album comparison logic.

- Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

In @src-tauri/crates/app/src/audio/engine.rs around lines 395 - 495, Both set_output_device and set_wasapi_exclusive duplicate the async resume logic (DB lookup of file_path/duration, fetch_replay_gain_db, then send AudioCmd::LoadAndPlay); extract this into a helper like resume_playback_async(app: tauri::AppHandle, cmd_tx: Sender<AudioCmd>, track_id: i64, position_ms: u64) that performs the SQL query, calls crate::commands::player::fetch_replay_gain_db, constructs the LoadAndPlay payload and sends it; replace the duplicated tauri::async_runtime::spawn blocks in set_output_device and set_wasapi_exclusive with a single call to spawn(resume_playback_async(...)).

- Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

In @src-tauri/crates/app/src/commands/browse.rs around lines 753 - 770, The SQL ranking in the AlbumTrackRaw query (CTE named ranked used by tracks_raw) currently orders by t.bit_depth DESC which penalizes DSD tracks (bit_depth = 1); change the ORDER BY to treat DSD as highest quality by mapping or prioritizing bit_depth = 1 first (e.g. add (t.bit_depth = 1) DESC before t.bit_depth DESC or replace t.bit_depth with a CASE that maps bit_depth = 1 to a very large value), keeping the existing secondary sort keys (t.sample_rate DESC, t.file_size DESC, t.id ASC) so the collapse keeps DSD variants for mixed DSD/PCM albums.

- Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

In @src-tauri/crates/app/src/commands/browse.rs around lines 1381 - 1402, The computed start_ms is wrong because r.bucket is in localtime but you reconstruct the midnight as UTC via and_utc(); replace that logic that builds start_ms to construct a local midnight using chrono::Local (e.g. Local.with_ymd_and_hms or Local.from_local_datetime) for year/month from r.bucket and handle the chrono::LocalResult variants (Single, Ambiguous, None) deterministically (pick one of the ambiguous datetimes or return an error) before calling timestamp_millis() so start_ms matches the SQL bucket timezone.

- Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

In @src-tauri/crates/app/src/commands/deezer.rs around lines 531 - 538, The UPDATE that sets artwork_id for an album currently ignores the query result so an invalid album_id can return Ok(()) without modifying anything; after calling sqlx::query(...).execute(&pool).await, capture the result (e.g., let res = ...), call res.rows_affected() and return an Err (or propagate a not-found error) when rows_affected() == 0 so callers see a failure; apply the same fix pattern for the analogous block around lines 564-571. Ensure you reference the upsert_artwork_row call and the album_id bind to locate the UPDATE statement to modify.

- Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

In @src-tauri/crates/app/src/commands/deezer.rs around lines 781 - 831, batch_fetch_missing_album_covers currently calls enrich_album_inner but never persists the new artwork relationship, so albums stay with artwork_id NULL; after a successful enrich_album_inner return (in batch_fetch_missing_album_covers) check the returned enrich object for the new artwork identifier (e.g. enrich.artwork_id or derive it from enrich.cover_path) and execute an UPDATE on the album row to set album.artwork_id to that artwork's id (or insert the artwork and use its id), ensuring the DB column read by list_albums/get_album_detail is updated; update the success counting only after the DB update succeeds and reference the functions enrich_album_inner and batch_fetch_missing_album_covers and the album.artwork_id column when making the change.

- 

- Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

In @src-tauri/crates/app/src/commands/dlna.rs around lines 44 - 52, The function build_resources is declared pub but appears to be used only within this crate/module; change its visibility to pub(crate) by updating the function signature for build_resources (which returns DlnaResources and takes &AppState) to restrict export to the crate, and run cargo check to ensure no external callers break—if any external uses exist, either keep it pub or update those call sites accordingly.

- Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

In @src-tauri/crates/app/src/commands/duplicates.rs around lines 66 - 71, The correlated subquery using GROUP_CONCAT to build artist_name for each track (SELECT GROUP_CONCAT(name, ', ') ... FROM track_artist ta JOIN artist ar ...) can be very slow on large libraries; change the query to avoid per-row subqueries by LEFT JOINing track_artist and artist to the main tracks result and aggregating artists in a single GROUP BY, or alternatively fetch tracks first and batch-query/articulate artist lists in Rust; update references to the artist_name projection, the track_artist and artist joins, and any code consuming artist_name (e.g., the alias artist_name and query using t.id) to match the new shape so duplicates.rs produces the same artist list without per-row GROUP_CONCAT subqueries.

- Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

In @src-tauri/crates/app/src/commands/duplicates.rs around lines 155 - 162, The current loop in the duplicates removal path iterates over track_ids and issues a separate DELETE per id (see the for id in track_ids loop, sqlx::query("DELETE FROM track WHERE id = ?"), tx and deleted variable), which is inefficient; replace it with a single batched DELETE executed once: either construct a parametrized DELETE ... WHERE id IN (?,?,...) with placeholders for all ids (or if you prefer SQLite-specific approach use DELETE FROM track WHERE id IN (SELECT value FROM json_each(?)) and bind a JSON array of ids), execute that single sqlx::query(...).bind(...).execute(&mut *tx).await, and set deleted from the returned rows_affected; also short-circuit when track_ids is empty to avoid invalid SQL.

- Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

In @src-tauri/crates/app/src/commands/edit.rs around lines 184 - 196, La fonction update_tracks_batch ne met pas à jour file_hash après l'appel à write_tags_to_file; applique la même correction que dans update_track_tags: après un write_tags_to_file(&path, &edit) réussi, recalculer/obtenir le nouveau file_hash (même logique que utilisée dans update_track_tags) et affecter-le à edit.file_hash (ou au champ approprié) avant d'appeler sync_db(&pool, *track_id, &edit). Assure-toi d'attraper et propager toute erreur éventuelle lors du recalcul du hash, et de conserver l'ajout d'erreurs à summary.errors pour track_id en cas d'échec.

- Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

In @src-tauri/crates/app/src/commands/edit.rs around lines 547 - 558, The code currently computes a blake3 hash of the image bytes and uses that for the artwork cache, but it never updates the track's file_hash after write_cover_to_file; compute the hash of the modified audio file (the file at file_path) after calling write_cover_to_file and assign it to track.file_hash (and persist it via the same updater you use elsewhere), leaving the image hashing logic (blake3::hash(&bytes) -> artwork cache) intact; update references to file_path, write_cover_to_file, track.file_hash and any save/update function used to persist the track so the stored file_hash reflects the new audio file contents rather than the image bytes.

- Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

In @src-tauri/crates/app/src/commands/edit.rs around lines 95 - 101, After write_tags_to_file(&path, &edit) completes, recompute the file's BLAKE3 hash and update track.file_hash on the edit object before calling sync_db(&pool, track_id, &edit). Specifically, after write_tags_to_file return, read the file at path and compute a BLAKE3 digest (use blake3::Hasher with streaming for large files), set edit.track.file_hash (or the equivalent field inside edit) to the new hex hash, then proceed to sync_db so the DB and scanner fast-path (mtime,size) remain consistent with the modified file.

- Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

In @src-tauri/crates/app/src/commands/integration.rs around lines 173 - 221, Before making the HTTP call via LastfmClient::auth_get_mobile_session in lastfm_login, call offline::is_offline() and short-circuit the flow when offline (do not call auth_get_mobile_session); return an appropriate LastfmStatus (e.g., configured flag based on local config presence but connected: false and username: None) or an AppResult error according to project convention so callers get a clear offline response instead of a network timeout. Ensure the check occurs before creating/using LastfmClient or invoking auth_get_mobile_session and keep existing DB write logic skipped when offline.

- Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

In @src-tauri/crates/app/src/commands/lyrics.rs around lines 1107 - 1115, The current code always pauses playback when payload.write_to_file is true (via engine.send(crate::audio::AudioCmd::Pause)), which must be restricted to Windows only; change the pause logic to run only on Windows (e.g., cfg!(windows) or cfg attribute) and keep behavior unchanged on macOS/Linux; after writing lyrics/files (in save_lyrics / edit::update_track_tags / set_track_rating paths) re-compute the file hash with blake3 and assign it to track.file_hash so the scanner’s (mtime,size) fast path remains valid, and preserve the existing small delay (tokio::time::sleep) only when the Windows-only pause is executed.

- Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

In @src-tauri/crates/app/src/commands/maintenance.rs around lines 34 - 72, The regen_in_dir function performs blocking filesystem operations (std::fs::read_dir, entry.path/is_file, file_stem) in a synchronous function, which can block the Tokio runtime; convert it into an async function (e.g., async fn regen_in_dir) and move blocking work into tokio::task::spawn_blocking or replace with async equivalents from tokio::fs (tokio::fs::read_dir, DirEntry::path/metadata) to avoid blocking; keep the same AppResult<u32> return semantics, preserve the existing error mapping to AppError::Io and tracing::warn calls, and continue to call crate::thumbnails::generate_thumbnails from the appropriate context (if generate_thumbnails is blocking, spawn that call on spawn_blocking as well) so behavior and logging remain unchanged.

- Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

In @src-tauri/crates/app/src/commands/player.rs around lines 754 - 762, The A-B loop handling currently only clears when both a_ms and b_ms are None; change it so passing None clears only that endpoint: when a_ms is Some store the value into shared.loop_a_ms (Ordering::Release) and when a_ms is None reset/clear the A endpoint (e.g., call the shared method that clears A or store a sentinel/reset value), do the same for b_ms with shared.loop_b_ms, and keep shared.clear_ab_loop() only for the case where both are None; update the logic in the function (e.g., player_set_ab_loop) and/or adjust the doc comment if the original behavior was intended.

- Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

In @src-tauri/crates/app/src/commands/playlist.rs around lines 396 - 419, canonical_path_key currently forces to_lowercase for all platforms which breaks case-sensitive filesystems; update the function so the UNC/verbatim prefix stripping and to_lowercase() are only applied on Windows (use cfg!(windows) or #[cfg(windows)] logic). Keep the canonicalize call and the s/canon variables, but if not on Windows simply return s (or s.as_ref()) without calling to_lowercase or performing the has_verbatim/has_unc checks; on Windows preserve the existing has_verbatim/has_unc handling and to_lowercase behavior. Ensure you reference canonical_path_key, canon, s, has_verbatim and has_unc when making the change so the platform-specific behavior is localized.

- Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

In @src-tauri/crates/app/src/commands/profile.rs around lines 242 - 249, The code currently calls the blocking std::fs::remove_dir_all on dir, which can block the Tokio runtime; replace that call with the async tokio::fs::remove_dir_all(&dir).await and keep the same error handling (tracing::warn!) so the directory removal is non‑blocking; if you cannot change the enclosing function to async, instead move the blocking call into tokio::task::spawn_blocking(|| std::fs::remove_dir_all(dir)) and .await that JoinHandle, preserving the existing error logging.

- Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

In @src-tauri/crates/app/src/commands/scan.rs around lines 153 - 159, The code currently hardcodes album_artist: None and is_compilation: false for DSD tracks, which loses Album Artist/compilation metadata and breaks album grouping; update the DSD metadata path to propagate the reader-extracted Album Artist and compilation flag into the record instead of setting those neutral defaults: locate where album_artist and is_compilation are set in the scan handling (symbols album_artist and is_compilation in the DSD branch of the scan logic) and replace the hardcoded values with the values returned from the DSD reader/parser (or map the reader's Album Artist tag and compilation flag into album_artist and is_compilation), ensuring the grouping key logic (canonical_title + album_artist_id and merge_implicit_compilations behavior) receives the actual metadata.

- Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

In @src-tauri/crates/app/src/commands/similar.rs around lines 506 - 511, The comparison uses to_lowercase() causing mismatches for names like "AC/DC" or "P!nk"; import canonical_name (from crate::commands::scan or waveflow_core::scanner) and use it instead of to_lowercase() when building canon and when comparing hit names inside the closure (the hits.into_iter().find(|h| ... ) block) so both source_name and h.name are canonicalized the same way before equality check; update the imports to include canonical_name and replace the two to_lowercase() calls with canonical_name(...) calls.

- Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

In @src-tauri/crates/app/src/commands/smart_playlists.rs around lines 115 - 133, Wrap the playlist INSERT and the subsequent custom::materialize call in a single database transaction so they succeed or fail together: start a transaction with pool.begin(), run the INSERT using the transaction (use sqlx::query(...).execute(&mut tx).await?), get the playlist_id from the insert result, then call custom::materialize using the same transaction (adjust custom::materialize to accept &mut Transaction if it currently takes &Pool or &PoolConnection), and only commit tx if materialize returns Ok (rollback on error). Apply the same transactional change to the other occurrence (the code around custom::materialize at the second location).

- Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

In @src-tauri/crates/app/src/commands/spotify.rs around lines 54 - 149, All Spotify command paths (spotify_login, spotify_get_access_token, spotify_list_playlists, spotify_get_playlist_tracks, spotify_get_queue, spotify_search) currently perform network ops or open a browser without checking offline mode; update each exported command to call offline::is_offline() at the start and short-circuit when true: for read/list/search calls return the appropriate empty payload types (e.g., empty Vec or default SpotifySearchResults) or a clear AppError/Result consistent with other services, and for spotify_login prevent opening the browser/awaiting callback and return an AppError or a SpotifyStatus indicating offline; ensure you use the same return shapes as each function (SpotifyAccessToken, Vec<...>, SpotifySearchResults, SpotifyStatus, spotify::SpotifyQueueSnapshot) so callers still deserialize correctly.

- Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

In @src-tauri/crates/app/src/commands/track.rs around lines 454 - 495, After writing the rating via write_rating_to_file in set_track_rating, recompute the file's blake3 hash on a blocking thread and persist it to the DB so track.file_hash stays in sync: inside the Some(path_str) branch (after write_rating_to_file returns), spawn a tokio::task::spawn_blocking that reads the file and computes the blake3 hash, then call the repository update to set the new file_hash (mirror the pattern used in save_lyrics), and ensure any errors from hashing or updating are logged (non-fatal) while still allowing repo.set_rating(track_id, rating).await? to run as before.

- Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

In @src-tauri/crates/app/src/db/profile_db.rs around lines 44 - 55, The ATTACH DATABASE SQL is built via format! using app_db_path_owned inside the after_connect closure which looks unsafe at first glance; add a clear code comment next to the after_connect/ app_db_path_owned usage stating that the path is not user-controlled and documenting its origin (e.g. initialized from AppState.paths.app_db at startup from a system-derived path), and note why escaping apostrophes is applied and that SQLite does not support parameter binding for ATTACH; reference the after_connect closure, app_db_path_owned variable, and AppState.paths.app_db so reviewers can verify the guarantee.

- Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

In @src-tauri/crates/app/src/discord_presence.rs around lines 357 - 397, The resolve_cover_url function currently calls crate::commands::deezer::enrich_album_inner which performs outbound HTTP; before invoking enrich_album_inner (and after resolving album_id), check offline::is_offline() and if it returns true skip the enrichment and return None (or keep existing cached result) to avoid network calls in offline mode; reference resolve_cover_url, enrich_album_inner, and offline::is_offline() when making the conditional guard.

- Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

In @src-tauri/crates/app/src/dlna/cds.rs around lines 380 - 390, The SQL selected column names do not match the fields expected by TrackRow in BrowseMetadata: the query selects t.primary_artist and t.bit_rate while sqlx::query_as::<_, TrackRow> expects artist_name and bitrate; update the SELECT to alias those columns to the struct field names (e.g. t.primary_artist AS artist_name and t.bit_rate AS bitrate) or alternatively rename TrackRow fields to match the SQL, ensuring the projection in the query and the TrackRow field names (used in the sqlx::query_as call) are identical.

- Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

In @src-tauri/crates/app/src/dlna/description.rs around lines 85 - 98, The xml_escape function currently preallocates String::with_capacity(input.len()), which underestimates worst-case growth when characters are escaped; update xml_escape to use a more conservative hint such as String::with_capacity(input.len() * 2) or String::with_capacity(input.len() + 32) so fewer reallocations occur when replacing chars like '&', '"', '\'' with multi-byte entities.

- Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

In @src-tauri/crates/app/src/dlna/description.rs around lines 126 - 131, Le test existant descriptor_escapes_xml_metachars ne couvre que <, > et &; ajoute dans la même fonction (ou un nouveau test) un input comme device_descriptor("Test's \"quoted\"", "http://x") et vérifie que la sortie contient &apos; pour l'apostrophe et &quot; pour les guillemets doubles; referencie la fonction device_descriptor lors de l'assertion pour s'assurer que les entités XML &apos; et &quot; sont bien produites.

- Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

In @src-tauri/crates/app/src/dlna/http.rs around lines 232 - 258, The path-traversal check in serve_art (checking hash_with_ext.contains('/') || hash_with_ext.contains("..")) misses Windows backslashes; update the validation for hash_with_ext to reject any path-separator or parent-dir tokens (e.g., check for '/' and '\\' and "..", or better use Path::new(&hash_with_ext).components().any(Component::ParentDir) and also ensure no separator characters) before joining against ctx.profile_artwork_dir / ctx.metadata_artwork_dir so that malicious inputs like "..\\..." are refused; keep the rest of the serve_art flow (mime_for_art, tokio::fs::read) unchanged.

- Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

In @src-tauri/crates/app/src/dlna/mod.rs around lines 274 - 284, pick_lan_ip() currently returns the first non-loopback IPv4 which can pick Docker/VPN addresses; update it to prefer RFC1918 LAN addresses by filtering for 10.0.0.0/8, 172.16.0.0/12 and 192.168.0.0/16 first (return the first match), falling back to the first non-loopback IPv4 if none match; reference the pick_lan_ip function to implement this prioritized filtering so the SSDP-announced URL is more likely to be a real LAN address (you can keep the existing fallback behavior used elsewhere intact).

- Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

In @src-tauri/crates/app/src/paths.rs around lines 36 - 39, The conversion currently drops the original error from handle.path().app_data_dir() by using .map_err(|_| AppError::MissingAppDataDir); instead, preserve context by mapping the error into AppError with the source (e.g., change AppError::MissingAppDataDir to hold a Box<dyn Error> or implement From<anyhow::Error> and do .map_err(|e| AppError::MissingAppDataDir(e.into()))), or if changing the enum shape isn’t possible, log the original error before mapping (e.g., capture e in .map_err(|e| { log::error!("app_data_dir failed: {:?}", e); AppError::MissingAppDataDir })). Ensure you reference handle.path().app_data_dir() and AppError::MissingAppDataDir when making the change.

- Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

In @src-tauri/crates/app/src/queue.rs around lines 937 - 975, The preshuffle snapshot currently serializes only track IDs, losing per-item identity and metadata so write_queue_order recreates rows as source_type='manual'/source_id=NULL and unshuffle uses track_id to find the current slot; instead, change the preshuffle/unshuffle flow to serialize the full queue item identity and metadata (the queue row primary key/ID plus source_type, source_id and any other columns used by append_to_user_queue) in preshuffle_json in preshuffle (where preshuffle_json is created) and restore by re-inserting/restoring those exact queue items (or calling a new helper like write_queue_order_from_items) so write_queue_order preserves original source_type/source_id and you locate the current index by the queue item ID (not track_id) when computing new_index in unshuffle/current_track; update any call sites (including append_to_user_queue logic) to rely on the stable queue item ID for boundaries.

- Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

In @src-tauri/crates/app/src/scrobbler.rs around lines 213 - 241, The current filter_map in fetch_due (inside run_once) drops rows with a NULL/empty artist but leaves those scrobble_queue entries in the DB so they are retried forever; change the logic to collect the IDs of dropped/poison rows while iterating (use the existing row tuple with the id and the artist variable), return None for those entries, and after fetch_due returns (in run_once) execute a single deletion of those IDs from scrobble_queue (e.g. DELETE ... WHERE id IN (...)) so the poison rows are removed immediately; reference the PendingScrobble construction, the artist variable, fetch_due and run_once and the scrobble_queue table when locating where to add the ID collection and deletion.

- Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

In @src-tauri/crates/app/src/spotify.rs around lines 423 - 444, La fonction list_playlists ne pagine que la première page (limit=50) et coupe les playlists au-delà de 50 ; changez-la pour itérer sur toutes les pages en appelant get_json successivement (ou en suivant la Page.next URL) jusqu'à ce que plus de pages n'existent, en accumulant les items Option<PlaylistResponse> comme actuellement, en aplatissant les None, en convertissant chaque PlaylistResponse via SpotifyPlaylistLite::from, et en retournant la collection complète; réutilisez les types Page, PlaylistResponse, SpotifyPlaylistLite et la constante API_BASE pour construire/follow la requête de pagination.

- Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

In @src-tauri/crates/app/src/spotify.rs around lines 466 - 503, La boucle de pagination silencieusement tronque les playlists quand MAX_PAGES (const MAX_PAGES) est atteint; remplacez le cas qui casse silencieusement par un retour d'erreur explicite: dans le match sur page.next, au lieu de tomber dans le _ => break quand pages >= MAX_PAGES et page.next.is_some(), renvoyez un Err décrivant que la pagination a été interrompue par le cap MAX_PAGES (inclure playlist_id, pages et éventuellement item_count) afin que l'appelant sache que la liste est incomplète; conservez le comportement actuel pour les autres chemins et utilisez les mêmes types d'erreur/context que les autres appels (p.ex. ceux autour de get_json/PlaylistTrackItem).

- Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

In @src-tauri/crates/app/src/spotify.rs around lines 290 - 326, Both exchange_code and refresh_token call the network even when the app is in offline mode; add an early check using offline::is_offline() at the top of each function (exchange_code and refresh_token) and short-circuit instead of making the reqwest POST: return a cached TokenResponse if available or an empty/default TokenResponse (Ok(TokenResponse::default())) so the UI stays consistent with app_setting['network.offline_mode'] (persisted offline state). Ensure you consult any existing cache retrieval logic before returning the empty response so cached tokens are used when present.

- Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

In @src-tauri/crates/app/src/state.rs around lines 176 - 180, The code calls previous.pool.close().await while holding the write lock obtained from self.profile.write(), which can block; change the flow to take the previous ActiveProfile out of the write guard, drop/release the guard, then perform previous.pool.close().await outside the lock, and finally re-acquire the write lock to set *guard = Some(ActiveProfile { profile_id, pool }) if necessary. Concretely: use guard.take() to extract previous, immediately drop the write guard (let guard go out of scope), await previous.pool.close().await if Some, then re-open self.profile.write().await and assign the new ActiveProfile.

- Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

In @src-tauri/crates/app/tauri.conf.json around lines 40 - 50, La configuration actuelle met "csp": null, ce qui désactive la protection XSS pour le renderer ; remplacez la valeur null de la clé csp dans tauri.conf.json par une politique CSP restrictive (par exemple restreindre default-src, script-src et style-src à 'self' et autorisations minimales) afin de protéger le WebView, tout en conservant assetProtocol.enable et les scopes existants ; assurez-vous que la politique couvre les besoins légitimes de votre UI (ajoutez nonces/hashes ou hôtes explicitement autorisés si nécessaire) et documentez tout assouplissement dans le repo.

- Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

In @src-tauri/crates/core/src/artwork/thumbnails.rs around lines 47 - 49, The current pre-check "if out.exists() { continue; }" in generate_thumbnails is vulnerable to TOCTOU when multiple threads/processes handle the same hash; replace it with either a per-hash lock/mutex keyed by the thumbnail hash (acquire before existence check and release after write) or perform atomic writes by writing to a temp file in the same directory and then rename() to `out` (atomic replace) so concurrent callers cannot corrupt the file; locate the check in function generate_thumbnails and protect the `out` path/variable accordingly.

- Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

In @src-tauri/crates/core/src/audio_format/dsd/parser.rs at line 357, Replace the test uses of std::iter::repeat(value).take(count) with the stabilized std::iter::repeat_n(value, count) in the DSD parser tests: change the call inside the out.extend(...) where payload_bytes is used (the expression out.extend(std::iter::repeat(0xAA).take(payload_bytes as usize))) and the analogous site near the other test (the second repeat/take at the other test). Use repeat_n(0xAA, payload_bytes as usize) (and the equivalent for the other case) to keep the same behavior and casts while using the modern API.

- Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

In @src-tauri/crates/core/src/metadata/deezer.rs around lines 73 - 81, Add a Default implementation for DeezerClient that delegates to the existing constructor: implement std::default::Default for DeezerClient with fn default() -> Self { DeezerClient::new() }; this keeps the current reqwest client construction in DeezerClient::new() and makes DeezerClient usable with Default::default() and derive-based APIs while changing no construction logic.

- Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

In @src-tauri/crates/core/src/metadata/lrclib.rs around lines 56 - 81, The get method in Lrclib client makes an outbound HTTP call without checking offline mode; add an early check at the top of pub async fn get to call offline::is_offline() and short-circuit (e.g., return Ok(None) or load from cache) so no request is sent when offline; update any callers/tests if they expect a different empty payload and keep references to symbols: offline::is_offline, get, LrclibResponse, BASE_URL, and self.http to locate where to add the guard.

- Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

In @src-tauri/crates/core/src/repository/sqlite/playlist.rs around lines 191 - 223, The current append_tracks() and create_with_tracks() implementations hold a single long-running transaction while inserting thousands of playlist_track rows; change them to process track_ids in chunks (e.g., batch size 200) using repeated transactions obtained from self.pool.begin(), run the per-row INSERT OR IGNORE (or call the existing upsert helpers that accept &mut sqlx::SqliteConnection) inside each transaction, commit after each batch, and then update playlist.updated_at (in a final short transaction or the last batch) so the write lock and WAL growth are bounded; ensure next_position is preserved across batches and use the upsert helper signatures that accept &mut SqliteConnection to participate in the opened transaction.

- Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

In @src-tauri/crates/core/src/repository/sqlite/playlist.rs around lines 233 - 264, The SELECT that reads removed_position must be executed inside the same transaction as the DELETE/UPDATE to avoid races: move the sqlx::query_scalar("SELECT position FROM playlist_track ...").bind(playlist_id).bind(track_id).fetch_optional(...) so it runs after let mut tx = self.pool.begin().await? and use fetch_optional(&mut *tx) (not &self.pool), then pattern-match the Option (let Some(pos) = removed_position) while still holding tx before performing the DELETE, the position-shifting UPDATE and the playlist updated_at UPDATE, and finally commit the tx; update variable references (removed_position/pos/tx) accordingly.

- Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

In @src-tauri/crates/core/src/repository/track.rs around lines 86 - 90, The FTS5 query built by the caller can contain unescaped special characters (parentheses, boolean operators like AND/OR/NOT, quotes, etc.) which produces invalid FTS5 syntax and crashes search_fts; update the code so search_fts validates and/or escapes FTS5 metacharacters before binding or catches FTS5 syntax errors and returns a clear CoreResult error. Concretely, in search_fts (and/or the caller that appends * to words) implement a sanitizer that removes/escapes characters like ()"'+-*/^:[]{}<> and reserved words, or run a lightweight validator that rejects malformed tokens and returns a user-friendly error; additionally wrap the sqlx query in error handling to map FTS5 syntax errors into a recoverable CoreResult with guidance. Ensure references to search_fts and the caller that composes fts_query are updated accordingly.

- Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

In @src-tauri/crates/core/src/scanner/extract.rs around lines 120 - 125, file_type_label currently maps FileType::Mp4 to "AAC" which mislabels ALAC; update file_type_label to inspect the Mp4 variant's codec (e.g., match FileType::Mp4(codec) or use the Mp4Codec accessor) and return the specific codec label (e.g., "ALAC" for Mp4Codec::Alac, "AAC" for Mp4Codec::Aac) and fall back to a generic container label like "MP4" for unknown/other codecs; modify the match arm for FileType::Mp4 in file_type_label to perform this nested match against Mp4Codec instead of unconditionally returning "AAC".

- Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

In @src-tauri/crates/core/src/scanner/upserts.rs around lines 29 - 57, La doc pour split_artist_name dit qu'on ne doit scinder que sur les séquences exactes ", " et "; ", donc ajuste l'implémentation de split_artist_name pour respecter ça (ou mets à jour la doc si tu préfères garder le comportement actuel) ; par exemple, remplace le split sur caractères par une chaîne en deux étapes : d'abord raw.split(", ").flat_map(|s| s.split("; ")).map(|s| s.trim()).filter(|s| !s.is_empty()).map(|s| s.to_string()).collect() afin de ne scinder que quand la virgule/point-virgule est suivie d'un espace, en conservant le nom de fonction split_artist_name et le comportement de trimmer/filtrer.

- Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

In @src-tauri/crates/core/src/smart_playlists/cover.rs around lines 260 - 264, Remove the dead NonZeroU32 construction: delete the creation of src_w_nz, src_h_nz, dst_w_nz, dst_h_nz and the unused tuple binding (_ = (src_w_nz, src_h_nz, dst_w_nz, dst_h_nz)); since the v6 crate API accepts u32 directly, keep using crop_w, crop_h, dst_w, dst_h as u32 and remove the expect! checks to avoid confusing dead code (verify no other code uses src_w_nz/src_h_nz/dst_w_nz/dst_h_nz before removing).

- Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

In @src-tauri/crates/core/src/smart_playlists/generator.rs around lines 197 - 200, When tracks.is_empty() you currently delete_existing_slot(pool, bucket.slot()).await? and return Ok(0), which causes a misleading playlist id to be pushed into created; change the function to return an Option<PlaylistId> (return Ok(None) when tracks.is_empty() after calling delete_existing_slot) and update the caller that collects IDs (the code that pushes into created around the current call site) to only push Some(id) values (or filter out None), so no spurious 0 id is added; keep references to delete_existing_slot, bucket.slot(), tracks.is_empty(), and the created vector when making the changes.

- Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

In @src-tauri/crates/core/src/smart_playlists/mod.rs around lines 68 - 70, The to_json method on SmartPlaylistRules currently calls serde_json::to_string(...).expect(...) which can panic; change the signature of pub fn to_json(&self) to return Result<String, serde_json::Error> and propagate the serialization error by returning serde_json::to_string(self) instead of unwrapping; update any call sites that assumed a String to handle the Result (e.g., map_err/unwrap_or_else or ? where appropriate), and ensure this covers cases where custom::CustomRules may contain non-serializable data.

---

Verify each finding against current code. Fix only still-valid issues, skip the
rest with a brief reason, keep changes minimal, and validate.

Inline comments:
In `@docs/features/smart-playlists.md`:
- Line 3: Update the incorrect link target for the smart playlists engine in the
document: replace the path fragment pointing to "smart_playlists/" under the app
crate with the core crate path used elsewhere so the link targets the extracted
engine (i.e. change the link in the first paragraph that references
smart_playlists to the same "../../src-tauri/crates/core/src/smart_playlists"
form used in the other occurrences); ensure this single-line link is consistent
with the other references on lines 19, 61, 73, 99, 101, 137.

In `@scripts/tauri.mjs`:
- Around line 15-29: Rename the top-level constants CONFIG_PATH and CONFIG_AWARE
to camelCase (e.g., configPath and configAware) and update all usages in this
file (including the args construction that references CONFIG_AWARE and
CONFIG_PATH) to the new names; ensure you only change the identifier names
(preserve string value 'src-tauri/crates/app/tauri.conf.json' and the Set
contents), and run a quick search in this module for any remaining references to
CONFIG_PATH/CONFIG_AWARE to update them to configPath/configAware.

In `@src-tauri/crates/app/src/commands/edit.rs`:
- Around line 23-25: La fonction update_track_tags ne recalcule pas le hash du
fichier ni ne met à jour track.file_hash après avoir réécrit les tags; modifiez
update_track_tags (dans src-tauri/crates/app/src/commands/edit.rs) pour, si la
piste éditée est la piste courante, suspendre la lecture, ré-hasher le fichier
modifié avec blake3 et assigner le nouveau hash à track.file_hash, puis
persister cette mise à jour (et mettre à jour tout cache/lookup dépendant du
hash) avant de reprendre la lecture; suivez la même séquence de
pause-rehash-update/persist pour les autres commandes similaires (save_lyrics,
set_track_rating) si applicable.

In `@src-tauri/crates/app/src/commands/integration.rs`:
- Line 14: The Last.fm login path (lastfm_login) currently calls
LastfmClient::auth_get_mobile_session unconditionally and must short-circuit
when offline; add a guard at the start of lastfm_login that checks
offline::is_offline() (and the persisted app_setting['network.offline_mode']
flag) and immediately return the offline-appropriate result (e.g., an empty
payload or a LastfmError::Offline/consistent local error) instead of calling
LastfmClient::auth_get_mobile_session so no network call is attempted while
offline.

In `@src-tauri/crates/app/src/commands/playlist.rs`:
- Around line 619-628: The current import builds a PlaylistDraft and calls
SqlitePlaylistRepository::create_with_tracks(&draft, &matched) but matched
contains only existing track_ids so M3U imports lose tracks on new machines;
instead pass the M3U entries through the scanner's upsert logic first and use
the resulting track IDs for playlist creation. Update the flow around
PlaylistDraft creation to call the scanner upsert/upsert_tracks function (the
scanner module or service used elsewhere in import) with the raw M3U entries,
collect the returned/created track IDs, and then call
SqlitePlaylistRepository::create_with_tracks(&draft, &upserted_ids) so playlists
always round-trip via the scanner.

In `@src-tauri/crates/app/src/commands/profile.rs`:
- Around line 69-79: Après l'INSERT du profil via profile_repo().insert(...) et
set_data_dir(...), envelopper l'initialisation des répertoires/BD
(ensure_profile_dirs() et profile_db::open()) de sorte qu'en cas d'erreur on
supprime le profil créé pour ne pas laisser une entrée cassée; concrètement,
après avoir calculé rel_dir et appelé repo.set_data_dir(profile_id,
&rel_dir).await?, exécuter ensure_profile_dirs(profile_id) et l'ouverture
profile_db::open(...); si l'une de ces opérations échoue, appeler le repository
pour retirer le profil créé (par ex. repo.delete(profile_id) ou
repo.remove(profile_id) selon l'API existante) et ré-émmettre l'erreur; assurer
que cette suppression est effectuée même si set_data_dir a réussi afin de garder
la DB cohérente.

In `@src-tauri/crates/app/src/commands/track.rs`:
- Around line 490-494: The DB is left with an outdated track.file_hash after
set_track_rating mutates the audio file; update the flow in set_track_rating
(and analogous commands edit::update_track_tags, save_lyrics) to 1) if the
edited track is the current playback track, pause playback (call the existing
playback pause helper), 2) after the file write succeeds re-compute the BLAKE3
hash of the file (use the project's blake3 hashing helper) and 3) persist the
new hash to the DB (either via repo.update_file_hash(track_id, new_hash) or
include file_hash in the same DB update that sets the rating) before emitting
"track:updated".

In `@src-tauri/crates/core/src/repository/playlist.rs`:
- Around line 1-6: Mettre à jour le doc-comment en tête de playlist.rs pour
refléter que la logique de smart-playlists a été déplacée dans crates/core
(répertoire src/smart_playlists) : remplacez la phrase indiquant que la logique
reste dans `crates/app` par une mention claire que le moteur smart-playlists vit
désormais dans `crates/core::smart_playlists`, qu’il doit consommer
`PathsContext` et rester portable côté serveur; conservez le reste du
commentaire sur la responsabilité du repository (ne pas évaluer les règles) et
mentionnez explicitement le chemin `src-tauri/crates/core/src/smart_playlists`
et l’exigence `PathsContext` pour aider la navigation future.

In `@src-tauri/crates/core/src/repository/sqlite/library.rs`:
- Around line 39-43: The artist_count currently uses COUNT(DISTINCT
primary_artist) on the track table which misses normalized multi-artist data;
change the query that builds the library totals (the SELECT starting with
"SELECT library_id, COUNT(*) AS track_count, ... FROM track") to join the
track_artist table (e.g., INNER JOIN track_artist ON track.id =
track_artist.track_id) and compute artist_count as COUNT(DISTINCT
track_artist.artist_id) (filtering by the same library scope) so distinct
artists are counted from canonical rows; also ensure any display-string
rebuilding elsewhere follows the GROUP_CONCAT ordering-by-position rule against
track_artist rows.

In `@src-tauri/crates/core/src/repository/sqlite/track.rs`:
- Around line 164-179: The search_fts function currently binds the incoming
limit directly, which allows limit ≤ 0 (and negative values) to produce
unbounded results; before building or executing the query in search_fts, enforce
a minimum bound on the limit (e.g., replace or shadow the incoming limit with a
clamped value such as 1 when limit <= 0) and then bind that sanitized limit to
the SQL query so TrackRow fetch_all on self.pool cannot return unexpectedly
massive results.

In `@src-tauri/crates/core/src/scanner/upserts.rs`:
- Around line 52-56: La méthode actuelle utilise raw.split([',', ';']) qui casse
des noms contenant ces caractères; changez la logique pour ne splitter que sur
les délimiteurs exacts ", " et "; " (par exemple en recherchant les séquences ",
" et "; " ou en utilisant une regex qui matche ces séquences), conservez l'ordre
en produisant des éléments avec leur position/index et insérez chaque artiste
individuel dans les lignes track_artist en enregistrant la position (p.ex. champ
position lié au track_id), puis assurez-vous que la reconstruction du display
utilise GROUP_CONCAT ORDER BY position pour reconstituer l'affichage dans le bon
ordre.

PR #202, opened 2026-06-04
feat(share): phase 1.g.3 desktop — canonical_id plumbing + share modal

InstaZDLL:
## Summary

End-to-end public sharing for desktop-created playlists. Closes the Phase 1.g loop alongside server PRs #25 / #26 / #27 + web PR #20.

## Backend (Rust / Tauri)

- Migration \`20260605000000_profile_canonical_id.sql\` adds \`profile.canonical_id TEXT\` with a partial UNIQUE index on non-NULL.
- \`db::profile_meta\` ships three helpers: \`backfill_canonical_ids\` (startup hook, UUIDs every NULL row), \`ensure_canonical_id\` (called from \`commands::profile::create_profile\` so a fresh profile never has a NULL canonical), and \`canonical_id_for\` (read-only lookup for the drain task).
- \`sync::drain\` injects the active profile's canonical id into every outbound \`SyncOpInBody.profile_canonical_id\` and gates the push when the canonical is still NULL (backfill pending) — better to defer than silently drop.
- \`commands::share\` gains \`share_link_mint\` / \`share_link_revoke\` / \`share_link_status\`. Mint / revoke hit waveflow-server's by-canonical surface (PR #27); status is local-only via a per-playlist \`share.token.<canonical>\` cache in \`profile_setting\`.

## Frontend (React / TypeScript)

- \`ShareModal\` component with QR code (qrcode lib), URL + copy button, mint / revoke CTAs. Opens against cached status for instant feedback.
- \"Partager\" button on PlaylistView header.
- \`lib/tauri/share.ts\` adds the three new wrappers next to \`saveShareImage\`.

## i18n

13 new keys propagated across all 17 locales via \`scripts/add_share_i18n.py\` (kept in tree for the next time we add a feature crossing the locale matrix).

## Validation

- \`cargo check --workspace --all-targets\` clean
- \`cargo test --workspace --lib --bins\` → 45/45
- \`bun run typecheck\` + \`bun run lint\` clean

## Test plan

- [x] CI confirms across the workspace
- [x] Manual: sign in to waveflow-server, create a desktop playlist, click Partager → QR + URL render, browser opens \`/p/<token>\` with the playlist preview
- [x] Manual: revoke clears the public URL (404 on web)
- [x] Manual: profile.canonical_id backfilled at first boot post-migration (check via SQL on app.db)

## Follow-ups (out of scope)

- \`playlist_track\` server materialisation (needs desktop emitting file_hash refs instead of local BIGINT track_ids).
- Server-side track-metadata sync so liked / rating ops can JOIN by file_hash.

## Summary by CodeRabbit

* **Nouvelles Fonctionnalités**
  * Partage de playlists via lien public (générer, consulter le statut, révoquer)
  * Modal de partage avec affichage de l’URL, copie en un clic et feedback “copied”
  * QR code généré automatiquement pour chaque lien

* **Internationalisation**
  * Chaînes traduites et disponibles dans 16+ langues pour l’interface de partage


InstaZDLL:
All seven findings addressed in d1151c9.

**1. `@types/qrcode` → devDependencies** — moved, type-only package.

**2. Neutral canonical_id error** — "restart the app" was misleading since the backfill already runs at `db::app_db::open`. New message: `profile {id}.canonical_id is NULL; this should normally be backfilled at startup` — states the unexpected state without prescribing.

**3 + 4. German typographic quote in subtitle** — `\"` → `"` in both `de.json` and `scripts/add_share_i18n.py` (so a re-run idempotently lands the correct character).

**5. i18n script fail-fast** — collects missing locales up-front and raises `FileNotFoundError` with the names instead of silently skipping. CI catches incomplete coverage now.

**6. ShareModal race** — sharp catch. The race was real: open → triggers `shareLinkStatus` (bumps session to N) → user clicks Mint quickly → handler reads `sessionRef.current` (still N) → mint completes (setLink to new value) → status resolves with N === sessionRef.current === N → `setLink(null)` overwrites the minted value. Fixed by pre-bumping `sessionRef` inside `handleMint` / `handleRevoke` (`const session = ++sessionRef.current`) so any in-flight status check hits a stale session and short-circuits.

**7. ShareModal stays open across playlist swap** — confirmed (no parent keying, the route just rebinds props). Reset via React-19's "adjust state during render" pattern:
```tsx
const [shareOpenForId, setShareOpenForId] = useState(playlistId);
if (shareOpenForId !== playlistId) {
  setShareOpenForId(playlistId);
  setIsShareOpen(false);
}
```
Avoids the `react-hooks/set-state-in-effect` lint and ships the reset in the same render pass.

Local: `bun run typecheck` + `bun run lint` + `cargo check --workspace --all-targets` all clean.

Issue #34, opened 2026-05-16
bug: Dark mode crash


Hashimini:
### WaveFlow version

1.0.0

### Operating system

Linux (AppImage on other distro)

### How did you install WaveFlow?

GitHub Releases (.AppImage)

### What happened?

Trying to change it to dark mode results in full app crash.

### Steps to reproduce

1. Open WaveFlow via appimage
2. Try changing to darkmode via de top right icon
3. Crash

### Relevant logs / errors

```shell

```

### Anything else?

No errors on the app log.

<img width="1451" height="940" alt="Image" src="https://github.com/user-attachments/assets/046227fd-bd3d-4d7f-b584-a69efc0351c1" />

_System information:

OS: Arch Linux x86_64
Kernel: Linux 6.19.10-1-cachyos

Desktop Environment: KDE Plasma (KWin)
Display Server: Wayland
XWayland/X.Org Version: 24.1.9

CPU: Intel Core i7-3770 @ 3.40GHz
Cores/Threads: 4C/8T

GPU: NVIDIA GeForce RTX 2060 SUPER (8GB VRAM)
Driver Version: NVIDIA 595.58.03
OpenGL Version: 4.6

Audio Device: Realtek ALC662

RAM: 16 GB

Monitor:

Resolution: 1920x1080
Refresh Rate: 119 Hz

Locale:

LANG=pt_BR.UTF-8_

InstaZDLL:
Hi @Hashimini, thanks for the report!

I’ll look into the AppImage issue. In the meantime, since you’re on CachyOS, I recommend installing WaveFlow through your system package manager via the AUR. That way updates are handled by the same tool as the rest of your system:

```bash
yay -S waveflow-bin
# or paru -S waveflow-bin
```

Hashimini:
Hi! I appreciate the response but I tried using the AUR version and its an instant crash to me. The app image works just fine (except for the dark mode thingy)

PR #209, opened 2026-06-08
feat(core): project track.album_id in postgres list + get + insert + update

InstaZDLL:
## Summary

Closes the asymmetry documented in the server's CLAUDE.md (and in the desktop \`project_sprint_4_web_progress.md\` follow-ups note).

Phase 4.d.0.1 (server PR #34) added the \`album\` table + the nullable \`track.album_id\` FK. Phase 4.d.0.4 (server PR #36) added the album drill-down endpoint which projects the real value. The legacy \`/api/v1/profiles/{p}/libraries/{l}/tracks\` collection still NULL-projected the column because the SELECT lives here in waveflow-core and predates the album surface — so the desktop / web client saw \`album_id = null\` on the per-library track list even when the row was materialised by the apply pipeline.

This PR swaps every \`NULL::bigint AS album_id\` for \`t.album_id\` (or \`album_id\` in the RETURNING blocks) in 4 sites:

- \`PostgresTrackRepository::list_for_library\` (SELECT)
- \`PostgresTrackRepository::get_for_library\` (SELECT)
- \`PostgresTrackRepository::insert_for_library\` (RETURNING)
- \`PostgresTrackRepository::update_for_library\` (RETURNING)

INSERT doesn't set \`album_id\` (the column list deliberately excludes it — track→album materialisation owns that field via the sync apply pipeline, not the CRUD write path), so RETURNING projects \`null\` for freshly inserted rows. UPDATE doesn't touch it either, so RETURNING projects whatever's stored. Both correct.

The module-level doc comment now calls out that \`album_id\` is the projected-real case; the remaining NULL casts (\`album_title\`, \`artist_*\`, \`artwork_*\`) stay until their materialisation paths land server-side.

## Test plan

- [x] \`cargo check --manifest-path src-tauri/Cargo.toml -p waveflow-core --features postgres --no-default-features\` clean
- [x] \`cargo check --manifest-path src-tauri/Cargo.toml --workspace --all-targets\` clean (covers the sqlite consumer in \`crates/app\` — no impact since it doesn't go through these methods)
- [x] \`cargo test -p waveflow-core --features postgres --lib --no-run\` builds clean
- [x] \`cargo fmt --all\` clean
- [x] CI runs the full suite

No desktop behavior change — \`crates/app\` consumes \`waveflow-core\` via the \`sqlite\` feature flag which routes through the sqlite repos, not the postgres ones touched here.

## Follow-up

A small PR on waveflow-server bumps the \`waveflow-core\` git rev in \`Cargo.toml\` to pick this up once this lands on \`main\`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Correction : les identifiants d'album sont désormais retournés correctement lors de la récupération et de l'énumération des pistes depuis la bibliothèque, au lieu de valeurs vides.

### DIFF
--- a/src-tauri/crates/core/src/repository/postgres/track.rs
+++ b/src-tauri/crates/core/src/repository/postgres/track.rs
@@ -281,7 +281,7 @@ impl PostgresTrackRepository {
           RETURNING id,
                     library_id,
                     title,
-                    NULL::bigint  AS album_id,
+                    album_id,
                     NULL::text    AS album_title,
                     NULL::bigint  AS artist_id,
                     NULL::text    AS artist_name,
--- a/src-tauri/crates/core/src/repository/postgres/track.rs
+++ b/src-tauri/crates/core/src/repository/postgres/track.rs
@@ -13,11 +13,15 @@
 //! waveflow-server side): a thin row carrying the columns that
 //! actually exist on disk for every file. The album / artist /
 //! artwork joins from the desktop's `TrackRow` projection still
-//! materialise in the response shape — they're projected as `NULL`
-//! casts of the right Postgres type until the album / artist /
-//! artwork tables ship on the server. Same pattern as
-//! `library.track_count = 0::bigint`: keep the wire field so the
-//! client doesn't need to adapt when the joins become real.
+//! materialise in the response shape — most are projected as `NULL`
+//! casts of the right Postgres type until the joined tables ship.
+//! `album_id` IS projected real since phase 4.d.0.1 added the
+//! `album` table + the nullable `track.album_id` FK; the album
+//! drill-down endpoint already relies on it, and the legacy
+//! `/tracks` collection now matches (closing the asymmetry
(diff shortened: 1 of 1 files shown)


InstaZDLL:
@coderabbitai resolved `album_id` in 60550e1 — missed the `update_for_library` RETURNING site in the previous commit (indent mismatch: 19 spaces vs 20 in the other RETURNING block, my `replace_all` only caught the matching indent). All 4 sites now project `album_id` consistently.

`album_title` skipped: the other 3 methods (list / get / insert RETURNING) also project `NULL::text AS album_title` because the SELECT doesn't join `album` to fetch the title — the wire shape stays uniform at `{album_id: Option<i64>, album_title: None}` across the 4 paths, matching the module doc that calls out "the remaining NULL casts (album_title, artist_*, artwork_*) stay until their materialisation paths land server-side." Projecting the real title here alone would re-introduce the asymmetry this PR exists to close.

PR #265, opened 2026-06-19
fix(plugins): refuse uninstall on bundled plugins + hide unavailable web-radio surface

InstaZDLL:
## Context

The Web Radio plugin is bundled with WaveFlow and re-seeded at every boot by [`ensure_bundled_plugins`](src-tauri/crates/app/src/state.rs). Today, three rough edges around its lifecycle:

1. **Uninstall has no effect after restart.** Clicking Uninstall in Settings → Plugins wipes `<plugins_root>/web-radio/` and drops the `app_setting` row — then the next launch's boot extractor finds the install dir missing and re-copies the plugin. The user sees the plugin "come back" with no UI explanation; reads as a bug.
2. **Disable leaves stale UI.** The sidebar "Web Radio" entry is hardcoded, so disabling the plugin leaves the entry visible. Clicking it lands on `WebRadioView` with the raw `"plugin disabled: web-radio"` error banner and an empty grid below.
3. **Uninstalling a plugin during an active stream orphans it.** Once a stream URL has been minted, the cpal engine doesn't know which plugin it came from. Removing the install dir while audio is playing leaves a stream with no owner.

### Backend (`src-tauri/crates/app/src/`)

- [`state.rs`](src-tauri/crates/app/src/state.rs) — exposes `is_bundled_plugin(id)` over the existing `BUNDLED_PLUGINS` constant.
- [`commands/plugins.rs`](src-tauri/crates/app/src/commands/plugins.rs)
  - `PluginInfo` gains `bundled: bool` so the UI can render the right surface.
  - `uninstall_plugin` refuses bundled ids upfront with a clear error (`plugin {id} is bundled with WaveFlow and cannot be uninstalled (disable it instead)`). The frontend hides the button entirely on bundled rows; the backend-side gate is the second line of defence.
  - `uninstall_plugin` stops the engine when `engine.shared().current_track_id < 0` (the negative-id sentinel `player_play_url` mints for every URL stream) before `remove_dir_all`. A library track is left alone. Plugin id isn't currently round-tripped to the engine, so this stops *any* URL stream — accepted trade-off vs orphaning an active stream whose source has just been removed.

### Frontend

- New [`src/hooks/usePluginAvailability.ts`](src/hooks/usePluginAvailability.ts) — resolves `installed && enabled` for a given plugin id, refreshes on the `waveflow:plugin-availability-changed` DOM event. Same lightweight bus the Spotify visibility toggle already uses.
- [`Sidebar.tsx`](src/components/layout/Sidebar.tsx) — conditions the Web Radio nav entry on `usePluginAvailability("web-radio")`.
- [`WebRadioView.tsx`](src/components/views/WebRadioView.tsx) — `isPluginUnavailableError()` sniffs the backend error for `"plugin disabled" | "plugin not installed" | "manifest id mismatch"`. When that matches OR the availability hook reports false, the view renders a Puzzle empty state pointing the user at Settings → Plugins instead of stacking a red banner on a stale grid.
- [`PluginsCard.tsx`](src/components/views/settings/PluginsCard.tsx) — renders a passive `Built-in` badge on bundled rows in place of the Uninstall button; fires the availability event on toggle + uninstall so consumers refresh.
- [`lib/tauri/plugins.ts`](src/lib/tauri/plugins.ts) — `PluginInfo` mirrors the new `bundled` field.

### i18n

Four new keys (`settings.plugins.bundled`, `settings.plugins.bundledHint`, `webRadio.unavailableTitle`, `webRadio.unavailableHint`) propagated to all 17 locales via [`scripts/add_bundled_plugin_i18n.py`](scripts/add_bundled_plugin_i18n.py) (idempotent — safe to re-run).

## Test plan

- [x] **Bundled refuses uninstall (backend):** patch `BUNDLED_PLUGINS` test-side, call `uninstall_plugin("web-radio")`, expect the typed error.
- [x] **Bundled badge replaces Uninstall (UI):** Settings → Plugins shows "Built-in" on the Web Radio row, no Uninstall button.
- [x] **Disable hides sidebar entry:** disable Web Radio in Settings → sidebar entry vanishes within a tick; re-enable → entry reappears.
- [x] **Disable empty-state in view:** open Web Radio view → disable plugin from Settings in another tab → view auto-switches to the Puzzle empty-state (no red banner).
- [x] **Backend stop-on-uninstall (manual third-party):** sideload a non-bundled plugin (or temporarily blank `BUNDLED_PLUGINS`), start its stream, uninstall mid-playback → engine stops cleanly, no orphan stream.
- [x] **i18n parity:** spot-check 3 locales (e.g. ja, ar, de) to confirm the bundled hint + webRadio.unavailableHint render and respect RTL on Arabic.
- [x] `bun run typecheck` ✅
- [x] `bun run lint` (excluding `.agents/` and `.claude/` skill scripts that ship pre-existing no-undef warnings) ✅
- [x] `cargo check --manifest-path src-tauri/Cargo.toml --workspace --all-targets` ✅

## Notes de Version

* **Nouvelles Fonctionnalités**
  * Les plugins intégrés (“bundled”) sont désormais identifiés : ils peuvent être désactivés mais restent non désinstallables.
  * Mise à jour en temps réel de la disponibilité des plugins dans toute l’interface.
  * Page d’état dédiée pour Web Radio lorsqu’elle est désactivée (avec titre et conseil pour l’activer).

* **Améliorations**
  * Navigation et contenu ajustés automatiquement selon l’état du plugin (entrée web radio masquée si indisponible).
  * Tracés d’interface et libellés i18n mis à jour pour Web Radio et les plugins intégrés (17 langues).

### DIFF
--- a/scripts/add_bundled_plugin_i18n.py
+++ b/scripts/add_bundled_plugin_i18n.py
@@ -1,0 +1,183 @@
+"""One-shot script: propagate i18n strings introduced by the
+"bundled plugin uninstall guard" PR (Web Radio polish).
+
+Adds:
+  - `settings.plugins.bundled` / `bundledHint`
+  - `webRadio.unavailableTitle` / `unavailableHint`
+
+across all 17 locales. Idempotent — re-running on a locale that
+already has the keys is a no-op.
+
+Run from repo root:
+    python scripts/add_bundled_plugin_i18n.py
+"""
+
+import json
+import sys
+from pathlib import Path
+
+PLUGINS_BUNDLED_KEY = "bundled"
+PLUGINS_BUNDLED_HINT_KEY = "bundledHint"
+WEBRADIO_UNAVAILABLE_TITLE_KEY = "unavailableTitle"
+WEBRADIO_UNAVAILABLE_HINT_KEY = "unavailableHint"
+
+TRANSLATIONS = {
+    "fr": {
+        "bundled": "Inclus",
+        "bundledHint": "Livré avec WaveFlow — désactivez-le pour le masquer.",
+        "unavailableTitle": "Web Radio désactivée",
+        "unavailableHint": "Activez le plugin Web Radio dans Réglages → Plugins pour parcourir plus de 30 000 stations.",
+    },
+    "en": {
+        "bundled": "Built-in",
+        "bundledHint": "Ships with WaveFlow — disable it to hide it.",
+        "unavailableTitle": "Web Radio is disabled",
+        "unavailableHint": "Enable the Web Radio plugin in Settings → Plugins to browse 30,000+ stations.",
+    },
+    "es": {
+        "bundled": "Integrado",
+        "bundledHint": "Incluido con WaveFlow — desactívalo para ocultarlo.",
+        "unavailableTitle": "Web Radio está desactivada",
+        "unavailableHint": "Activa el plugin de Web Radio en Ajustes → Plugins para explorar más de 30 000 estaciones.",
+    },
+    "de": {
+        "bundled": "Mitgeliefert",
+        "bundledHint": "Mit WaveFlow ausgeliefert — deaktivieren, um es auszublenden.",
+        "unavailableTitle": "Web Radio ist deaktiviert",
+        "unavailableHint": "Aktivieren Sie das Web Radio-Plugin in Einstellungen → Plugins, um über 30 000 Sender zu durchsuchen.",
+    },
+    "it": {
+        "bundled": "Integrato",
+        "bundledHint": "Incluso in WaveFlow — disattivalo per nasconderlo.",
+        "unavailableTitle": "Web Radio è disattivata",
+        "unavailableHint": "Attiva il plugin Web Radio in Impostazioni → Plugin per sfogliare oltre 30 000 stazioni.",
+    },
+    "nl": {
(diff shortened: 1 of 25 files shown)


InstaZDLL:
@coderabbitai both findings verified and fixed in 4b064a0:

**Finding 1 (`usePluginAvailability`):** Race confirmed — `cancelled` only covered unmount, two refreshes inside the same effect lifetime could swap. Added a per-effect token ref; the `.then` continuation bails when `token !== reqRef.current`.

**Finding 2 (`WebRadioView` `[]` deps):** Confirmed — landing with the plugin disabled parked the error and never re-fetched on re-enable. Made the effect depend on `pluginAvailable`; on flip OFF it resets every fetched-state field (entries, tracks, activeEntry, searchActive, error, loading) so a flip ON gives a clean fetch instead of briefly flashing the previous category list under the loading state.

InstaZDLL:
@coderabbitai verified — the rule does fire on the reset block (one error at the first setState). Suppressed in c84e5dd with a targeted `eslint-disable-next-line` + a short rationale (React 19 batches the 6 calls into one re-render, so the cascading-renders concern doesn't apply; a useReducer-with-RESET refactor would sidestep the rule but is heavier than this earns). Lint clean.

PR #267, opened 2026-06-19
feat(lyrics): word-mode editor auto-advances to next line on last-word capture

InstaZDLL:
## Problem

In the synced lyrics editor's word-by-word mode, the natural Space-Space-Space rhythm stalled at every line boundary. After the last word of a row was stamped, the user had to click the next row (or press Enter explicitly) before they could keep going.

User report: \"quand on arrive au dernier mot il faudrait que l'on passe automatiquement à la ligne suivante sans devoir cliquer manuellement sur la ligne\".

## Fix

[\`LyricsEditorModal.tsx::captureWord\`](src/components/common/LyricsEditorModal.tsx) now raises a \`shouldAdvanceToNext\` flag inside the \`setSyncedRows\` updater when:

1. The freshly-stamped word was the last word of the active row (\`wordCursor >= words.length\` after the stamp)
2. There is an existing next row to land on (\`idx + 1 < rows.length\`)

A follow-up \`setActiveRow(i => i + 1)\` lives outside the updater so the activeRow change lands as its own idempotent state mutation — safe under React's strict-mode double-invoke (the updater's flag write is idempotent; the outer setter runs once per keystroke).

The auto-advance **only skips between existing rows**. We deliberately don't auto-append a fresh empty row on the very last line — Enter still owns that explicit \"add a new line\" semantic. The change is invisible to a user who's at the end of a finished lyric, and game-changing to anyone capturing a 60-line song.

## Test plan

- [x] Open a track with multiple lines, switch to Word-by-word
- [x] Hit Space until you stamp the last word of line 1 → activeRow auto-bumps to line 2, next Space starts capturing line 2's first word
- [x] Hit Space on the last word of the LAST row → activeRow stays put (no auto-append). Enter still appends a fresh empty line.
- [x] Backspace on a row that was auto-advanced INTO → un-stamps line 2's first word (no surprise back-jump)
- [x] \`bun run typecheck\` ✅

## Summary by CodeRabbit

* **Améliorations**
  * Amélioration de l’édition de paroles synchronisées en mode capture « par mot » : l’avance automatique du focus vers la ligne suivante est désormais déclenchée après la mise à jour des lignes synchronisées, pour une meilleure cohérence.
  * Une action au-delà de la fin de la ligne n’entraîne plus de capture : le passage à la ligne suivante ne se produit que si elle existe.
  * Après la capture du dernier mot d’une ligne, l’éditeur passe automatiquement à la ligne suivante sans créer de ligne vide implicite.

### DIFF
--- a/src/components/common/LyricsEditorModal.tsx
+++ b/src/components/common/LyricsEditorModal.tsx
@@ -288,39 +288,55 @@ export function LyricsEditorModal({
   // stamped, the next press advances to the next line (and stamps the
   // line's own timeMs if it's still -1, like line mode).
   const captureWord = useCallback(() => {
-    // Signal lifted out of the state updater so the follow-up
-    // `setActiveRow` lands as a separate, idempotent state change.
-    // Set inside the updater is safe under strict-mode double-invoke
-    // because both passes write the same boolean. The follow-up
-    // setter is outside the updater so it fires exactly once per
-    // user keystroke.
-    let shouldAdvanceToNext = false;
+    if (syncedRows.length === 0) return;
+    const idx = Math.min(activeRow, syncedRows.length - 1);
+    const currentRow = syncedRows[idx];
+    const currentWords = currentRow.words ?? tokenize(currentRow.text);
+    const currentCursor = currentRow.wordCursor ?? 0;
+    const hasNextRow = idx + 1 < syncedRows.length;
+
+    // Decide auto-advance BEFORE mutating state. The previous version
+    // set a `shouldAdvance` flag inside the `setSyncedRows` updater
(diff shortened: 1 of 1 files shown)


InstaZDLL:
@coderabbitai verified — this finding was already resolved by 3fb89b8 (third commit on this PR, pushed after the initial review). The current code at lines 290–358 computes `shouldAdvance` **before** the `setSyncedRows` call, against the live `syncedRows[idx]` snapshot from the closure. The `if (shouldAdvance)` at line 355 reads a `const` set synchronously, so the React-18 race the review described (updater running asynchronously / under strict-mode double-invoke) no longer applies. No further change needed.

PR #268, opened 2026-06-19
fix(locales): repeat-button aria labels describe current state, not next click

InstaZDLL:
## Context

The translation validation pass surfaced a logical bug inherited across all 17 locales: the `player.controls.repeatOff/All/One` accessible labels were describing the action the **next click** would perform, not the button's current state.

Concrete impact: a screen-reader user with the player in **Repeat All** mode heard `"Repeat one"` (the next state in the cycle), in **Repeat One** heard `"Disable repeat"`. The wrong state, every time. The validator AI flagged this in 6+ locale reports independently (it, hi, fr, es, pt-BR, …) because each translation faithfully mirrored the English bug.

## Fix

Per ARIA toggle-button convention, the accessible name should reflect the **current** state. Relabelled every locale:

| Key | Was | Now |
|---|---|---|
| `repeatOff` | "Enable repeat" | "Repeat off" |
| `repeatAll` | "Repeat one" | "Repeat all" |
| `repeatOne` | "Disable repeat" | "Repeat one" |

Applied via [`scripts/fix_repeat_aria_labels.py`](scripts/fix_repeat_aria_labels.py) (idempotent) — 17 locales updated atomically with locale-appropriate phrasing.

## Brand tokens

[`CLAUDE.md`](CLAUDE.md) now formalises `Daily Mix` and `On Repeat` as untranslated brand-like tokens. Spotify keeps both verbatim in every market and Apple Music does the same with `New Music Mix` — these are explicit feature names, not generic UI strings. The validator was overzealous flagging them as "untranslated" in the per-locale reports; documenting the convention prevents future PRs from churning them.

`Web Radio` and `Now Playing` remain generic UI terms that should still be translated — those are handled in the per-locale follow-up PRs (Cluster 3 of the translation-validation sprint).

## Test plan

- [x] `bun run typecheck` ✅ clean
- [x] Read aria-label on the repeat button across the 3 modes in EN — confirm matches the current state
- [x] Spot check 3 locales (de, ja, ar) for natural phrasing in the new strings

## Notes de version

* **Correctifs**
  * Correction des libellés des modes de répétition du lecteur (désactivée, tout, une piste) afin qu’ils reflètent correctement l’état actuel.
  * Harmonisation des libellés ARIA pour une meilleure cohérence de l’accessibilité.
* **Documentation**
  * Mise à jour des consignes de localisation pour préciser la convention de repli, la couverture de toutes les clés, le maintien des éléments non traduits et la conservation des placeholders.


InstaZDLL:
@coderabbitai partially valid — fixed in 4aac021 with a tweak: the literal "exit non-zero when skipped is non-empty" suggestion would have broken idempotent re-runs (a second run on an already-patched repo lands every locale in `skipped` and would exit 1 spuriously, which would fail any CI / pre-commit hook calling this script unconditionally).

Split `missing` out of `skipped` instead — `missing` now means "locale file not on disk" (real out-of-sync working copy, fatal → exit 1, stderr message) while `skipped` keeps its existing benign meaning ("locale already at target values, no-op"). Idempotent runs stay exit-0; missing-file runs now surface clearly. Same fix preemptively applied to #269's twin script for consistency (b9fcc95 over there).

PR #269, opened 2026-06-19
fix(locales): de-franglais skin descriptions + CJK punctuation (cluster 2)

InstaZDLL:
## Context

Second translation-validation cleanup wave. The AI validator flagged consistent franglais / machine-translation artefacts in [\`settings.appearance.skin.subtitle\`](src/i18n/locales/en.json) and the five [\`settings.appearance.skins.<id>.description\`](src/i18n/locales/en.json) strings across 14 of the 16 non-English locales — the new 1.6.0 skin family ships UI-design jargon (\`pills\`, \`hairline rules\`, \`springy motion\`, \`pill capsules\`) that doesn't survive a literal pass.

## Findings, by failure mode

**\`pills\` / \`pill capsules\` mistranslated as medication**
- DE: \`Mono-Pills\` / \`Pillen-Kapseln\` — sounds medical
- KO: \`알약\` — literal medicine pills
- ZH: \`胶囊药丸\` — literal medicine capsules
- TR: \`mono pill'ler\`, \`kapsül haplar\` — \`hap\` is medication
- IT: \`pillole mono\`, \`capsule pillola\` — medication / redundant
- RU: \`моно-пилюли\` — medication
- PT-BR: \`cápsulas em pílula\` — redundant, nonsensical

**\`springy motion\` rendered as literal-spring constructions**
- IT: \`animazioni a molla\`, ES: \`animaciones tipo resorte\`, ID: \`animasi pegas\`, PT-BR: \`animações com mola\` — all literally "spring animations"

**\`hairline rules\` / \`soft grain\` misread as hair / dots**
- AR: \`خطوط شعرة\` (hair lines)
- ID: \`bintik halus\` (fine dots instead of paper grain)

**CJK punctuation**
- ZH-CN + ZH-TW used half-width \`,\` mid-sentence where CJK conventions require full-width \`，\`

## Fix

Patches verbatim from the per-locale "Suggested target" column of [\`secrets/translation-validation-reports/2026-06-19T13-58-13-349Z\`](https://github.com/InstaZDLL/WaveFlow). Applied via [\`scripts/fix_skin_descriptions_i18n.py\`](scripts/fix_skin_descriptions_i18n.py) — idempotent, safe to re-run.

14 locales touched: \`fr, de, es, it, nl, pt, pt-BR, ru, tr, id, ko, zh-CN, zh-TW, ar\`. \`ja, hi\` had no skin-description issues flagged.

## Notes

[...]

PR #270, opened 2026-06-19
fix(locales): fr _zero plurals + zh-CN/zh-TW CJK punctuation cleanup

InstaZDLL:
## Context

Third translation-validation pass, this time targeting the systemic mechanical bugs that pass-2 surfaced after #268 + #269 landed. Two clusters:

### 1. French `_zero` keys at singular instead of plural

French grammar requires a plural noun after `0` — `0 titres`, not `0 titre`. The original JSON authoring pass borrowed English plural rules where `0 track` is singular. The validator caught ~30 keys reading wrong.

Fixed every leaf string whose key ends in \`_zero\` and whose value matched a known singular form (\`titre\`, \`album\`, \`artiste\`, \`genre\`, \`dossier\`, \`écoute\`, \`minute de musique\`, \`jour d'affilée\`, \`sélectionné\`). Idiomatic alternative phrasings (\`Aucun résultat\`, \`Rien à supprimer\`, \`À suivre\`, \`vide\`) are deliberately left alone — they're not "0 X" patterns and don't need pluralisation.

### 2. zh-CN + zh-TW half-width punctuation inside CJK text

Both Chinese locales had drifted to ASCII \`,\` \`:\` \`?\` \`(\` \`)\` between Chinese characters where CJK conventions require full-width \`，\` \`：\` \`？\` \`（\` \`）\`. The validator flagged this across many strings.

Context-sensitive walk:
- \`,\` \`:\` \`?\` \`!\` → full-width **only** when surrounded by CJK characters
- \`(content)\` → \`（content）\` **only** when the content is purely CJK

The second rule preserves intentional half-width Latin annotations like \`(A → Z)\` and \`(A = {{a}})\`, which the CJK style convention keeps half-width when the parenthetical wraps foreign script. Otherwise the script would over-convert and introduce regressions.

## Script

[\`scripts/fix_locales_mechanical.py\`](scripts/fix_locales_mechanical.py) — idempotent, exit non-zero only on missing locale files (same convention as \`fix_repeat_aria_labels.py\` and \`fix_skin_descriptions_i18n.py\`).

**Touched:** 30 fr keys, 22 zh-CN strings, 21 zh-TW strings.

## What's NOT in this PR

- **id pronoun consistency** (Anda/kamu mixing), **ar audio terms** (resample/downmix/pitch), **pt-BR strip vs track confusion**, **zh terminology drift** (Library/Profile/Genre/Cover) — these need per-locale judgment calls, not mechanical sweeps. Deferred to per-locale follow-ups or 1.6.0 community-contribution work.
- Brand tokens (\`Daily Mix\`, \`On Repeat\`, etc.) stay verbatim per the convention formalised in #268.

## Test plan

- [x] Switch through fr in Settings → confirm "0 titres", "0 albums" etc. render naturally
- [x] Switch through zh-CN → confirm `（分钟）`, `？` between CJK, etc. look right
- [x] zh-TW same spot check
- [x] Re-run \`bun secrets/validate-translations.mjs\` — fr plural bug + zh CJK punctuation findings should be gone

## Summary by CodeRabbit

* **Localization**
  * Correction des libellés de pluralisation en français pour les compteurs (notamment le cas “0” dans les barres, sections Bibliothèque et vues détaillées).
  * Harmonisation de la ponctuation et de l’espacement en chinois (simplifié et traditionnel), avec ajustements des guillemets, séparateurs et parenthèses pour une typographie plus cohérente.
* **Chores**
  * Ajout d’un script autonome pour corriger automatiquement et de façon idempotente certaines dérives de traductions, avec rapport de traitement (modifiés/non modifiés).

### DIFF
--- a/scripts/fix_locales_mechanical.py
+++ b/scripts/fix_locales_mechanical.py
@@ -1,0 +1,189 @@
+"""Mechanical i18n cleanup — cluster 3 (post-validator pass-2).
+
+Two systemic bugs the AI validator surfaced after #268 + #269 landed:
+
+1. **fr `_zero` keys at singular instead of plural.** French grammar
+   requires a plural noun after `0` (unlike English, where `0 track`
+   is singular). The first JSON authoring pass borrowed the English
+   plural rules and parked every `_zero` form at singular, which the
+   validator caught across ~25 keys.
+
+2. **zh-CN + zh-TW half-width ASCII punctuation inside CJK text.**
+   CJK conventions use full-width `，` `：` `（` `）` between Chinese
+   characters; the source JSON drifted to half-width `,` `:` `(` `)`
+   in many places. The validator flagged this in both Chinese
+   locales.
+
+Idempotent — re-running on already-clean files is a no-op. Returns
+non-zero only when a locale file the script expected isn't on disk.
+
+Run from repo root:
+    python scripts/fix_locales_mechanical.py
+"""
+
+import json
+import re
+import sys
+from pathlib import Path
+
+
+# --- Section 1 — French `_zero` pluralisation ------------------------
+
+# Map of singular → plural noun stems. The script walks every leaf
+# string whose key ends with `_zero` and, when the value starts with
+# the singular stem, rewrites it to the plural. Stems that the
+# validator did NOT flag (idiomatic alternatives like `"Aucun résultat"`
+# or `"Rien à supprimer"`) are deliberately absent from this map so
+# the script leaves them alone.
+FR_ZERO_FIXES = {
+    "0 titre": "0 titres",
+    "0 album": "0 albums",
+    "0 artiste": "0 artistes",
+    "0 genre": "0 genres",
+    "0 dossier": "0 dossiers",
+    "0 écoute": "0 écoutes",
+    "0 minute de musique": "0 minutes de musique",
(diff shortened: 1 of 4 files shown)


InstaZDLL:
@coderabbitai both findings verified and fixed in ab8674f:

1. **JSON parse handling** — confirmed `patch_locale` only did `with open(...) as fh: json.load(fh)` with no error handling. Wrapped in try/except, return sentinel `-2` so `main()` collects unparseable files into a separate list alongside `missing` and exits non-zero with a clear "parse error in <path>: line X col Y: msg" message on stderr.

2. **FW parens around non-CJK content** — confirmed 10 pre-existing strings in each of zh-CN and zh-TW carried `（…）` around mixed content (template vars `{{version}}`, Latin acronyms `MP3、FLAC、…`, product names `Rust`, etc.). Added the symmetric reverse rule: FW→HW when content has any Latin / digit / template-variable character, paired with the existing HW→FW for pure CJK. The two rules together are idempotent (a pure-CJK `（分钟）` stays FW, a mixed `（{{version}}）` flips to HW, and re-running on a clean file is a no-op). 13 strings touched per locale by the reverse pass.

Re-ran the script — exit 0, idempotent on subsequent runs.

PR #273, opened 2026-06-20
fix(lyrics): graceful fallback when external providers error after LRCLIB miss

InstaZDLL:
## Summary

External lyrics-search errors (megalobiz HTTP 500, network timeouts, etc.) were propagating all the way to the LyricsPanel + FullscreenLyrics as a raw error string painted into the lyrics-content surface.

Reported visually:

> `d: http request failed: HTTP status server error (500 Internal Server Error) for url (https://www.megalobiz.com/search/all?qry=ROCKSTAR+-+TAK+Remix+Instrumental+TAK%2C+DokeV%2C+GALACTIKA*%2C+LUENA&searchButton.x=0&searchButton.y=0)`

## Root cause

Two layers of the same omission — the existing graceful-fallback pattern wasn't applied consistently.

**Backend** — [fetch_lyrics](src-tauri/crates/app/src/commands/lyrics.rs) calls the post-LRCLIB fallback chain `[NetEase, Megalobiz, Genius]` at two sites and `?`s the result naively:

```rust
if let Some(result) = external_lyrics_search(...).await? { ... }
```

But the syncedlyrics client intentionally returns `Err(last_error)` when at least one provider errored AND no provider produced acceptable content — this is correct cache-poisoning protection. The naive `?` turns that "transient blip on one provider" into a full RPC failure.

The Musixmatch-enhanced tier (lines 808-824) and the prefetch path (lines 1336-1339) already handle this gracefully with `Err(err) => tracing::debug!(...)`. The two post-LRCLIB sites were the only ones still naively `?`'ing.

**Frontend** — [LyricsPanel.tsx](src/components/layout/LyricsPanel.tsx) and [FullscreenLyrics.tsx](src/components/player/FullscreenLyrics.tsx) treated the resulting error as a display message:

```tsx
) : error ? (
  <EmptyState text={error} />
```

→ the raw `AppError::Other(format!("external lyrics search failed: {err}"))` was painted verbatim where the lyrics should have been.

## Fix

**Backend** ([6b64b8b](6b64b8b)): both post-LRCLIB fallback sites now `match` on the result. `Ok(Some(result))` caches as before; `Ok(None)` + `Err(_)` both fall through to the existing "cache empty + return empty payload" path. LRCLIB has already given us a definitive verdict at that point (404 or empty row), so a fallback-chain error is non-fatal.

**Frontend** ([92a7ba5](92a7ba5)): the `error` branch is collapsed into the same `lyrics.notFound` empty state used when `payload` is null. The `console.error(...)` in the catch handler keeps the technical detail for debugging; the user gets a clean "no lyrics found" panel with the music icon and the import/refetch controls still available. Defense in depth in case other transient errors still bubble up.

## Notes

- Cache-poisoning guard preserved: only LRCLIB-confirmed misses get cached as empty. If LRCLIB itself errors (timeout, DNS, refused), we still surface that to the caller — only fallback-chain errors degrade silently.
- Megalobiz reliability is a separate concern (their search endpoint 500s often on queries with special chars like the `*` in `GALACTIKA*`). Could demote or drop it from `external_fallback_providers()` in a follow-up.

## Test plan

- [x] Play a track for which LRCLIB has no entry AND megalobiz returns 500 (e.g. obscure track with `*` in artist name)
- [x] LyricsPanel + FullscreenLyrics both show the music icon + `No lyrics found for this track. You can import a .lrc file.` instead of the raw HTTP error
- [x] Backend logs the megalobiz failure at debug level (visible with `RUST_LOG=waveflow_lib=debug`)
- [x] The empty miss is cached: re-opening the panel doesn't re-hit the network
- [x] Refetch button still re-runs the waterfall (drops cache + re-walks providers)
- [x] Tracks with valid LRCLIB entries still resolve normally

## Bug Fixes
* Affichage des paroles plus fiable : les erreurs “transitoires” de récupération sont désormais distinguées de l’absence de paroles, sans afficher le texte d’erreur brut.
* En cas d’échec côté fournisseurs (ou de réponse LRCLIB vide), l’interface continue et propose une recherche de repli, en mettant en cache le résultat afin d’éviter que le chargement soit interrompu.
* Harmonisation du rendu entre le panneau et l’écran plein écran, avec un message d’erreur centré et une icône d’alerte.

## Internationalisation
* Ajout de la clé de traduction `lyrics.fetchError` dans les langues disponibles.


InstaZDLL:
@coderabbitai applied all 5 findings in 6d61afa:

**Findings 1+2 (`handleRefetch` / `handleImport` don't clear error)** — valid. Both handlers now call `setError(null)` immediately after `setPayload(next)` in the success path. Without it the stale error from the initial fetch would shadow the freshly-loaded payload behind the conditional below.

**Finding 3 (separate error from notFound in LyricsPanel)** — valid in tandem with 1+2. Once the error state is properly cleared on success, conflating it with notFound throws away actionable signal. Split the branches: `error` paints `AlertCircle` + new `lyrics.fetchError` key ("Couldn't reach lyrics providers. Try again later."); `!payload || empty` keeps the existing `Music2` + `lyrics.notFound` copy.

**Finding 4 (FullscreenLyrics inherits the same bug)** — valid, fixed symmetrically. The state lives in the parent so clearing error in `LyricsPanel` resolves the prop-drilling side, and the same icon/message split was applied to `FullscreenLyrics`'s conditional ladder for visual consistency.

**Finding 5 (extract `try_external_fallback_or_miss` helper)** — valid. The two LRCLIB branches (404 + empty-row) carried near-identical 30-line copies of the fallback-chain match + cache-empty + payload-build block. Extracted to a single helper just below `cache_external_lyrics`; both call sites collapse to a 3-line `return …or_miss(&pool, track_id, &meta).await.map(Some)`. Error-handling policy now lives in one place.

**i18n** — new `lyrics.fetchError` key added to all 17 locales (`fr` source of truth, per-locale register matched). Inserted right after `lyrics.notFound` to keep semantic ordering.

Validated: `bun run typecheck` clean, `bunx eslint` clean on touched files, `cargo check -p waveflow` compiles.

PR #274, opened 2026-06-20
fix(settings): portal language dropdown to escape Lounge/Liquid stacking trap

InstaZDLL:
## Repro

Open **Settings → Library tab** on either the **Lounge** or **Liquid Glass** skin → click the language picker. The dropdown options paint behind the cards below it instead of floating above them.

| Lounge | Liquid Glass |
|---|---|
| (red glow theme) | (teal glass theme) |

Both show the same symptom: "Deutsch" peeks out, then the UI Zoom card stacks on top of the rest of the list.

## Cause

Both skins paint Settings cards through `backdrop-filter`:
- [`lounge.css:175-176`](src/styles/skins/lounge.css#L175-L176), [201-202](src/styles/skins/lounge.css#L201-L202), [230-231](src/styles/skins/lounge.css#L230-L231)
- [`liquid.css:308-309`](src/styles/skins/liquid.css#L308-L309), [358-359](src/styles/skins/liquid.css#L358-L359), [434-435](src/styles/skins/liquid.css#L434-L435)

Per CSS spec, **any element with `backdrop-filter` creates a new stacking context**. The dropdown's `z-50` therefore only stacked WITHIN its parent card's context — every subsequent card's own `backdrop-filter` stacking context then painted on top of it.

## Fix

Render the listbox via `createPortal` to `document.body` using fixed positioning derived from a snapshot of the trigger button's bounding rect. The portaled node escapes every ancestor stacking context, so the listbox now stacks against the document root rather than being trapped inside a `backdrop-filter` card.

Side touches required:
- `triggerRef` on the button so we can measure its rect on open
- `listboxRef` on the `<ul>` so click-outside whitelists the portaled subtree (without it, mousedown inside the dropdown trips the close-on-outside handler and dismisses the menu before the option's `onClick` fires)
- Close on scroll/resize since the fixed coords are a snapshot — matches native `<select>` behaviour and avoids re-measuring per frame

## Test plan

- [x] On Lounge skin: language dropdown options fully visible above all cards
- [x] On Liquid Glass skin: same
- [x] On Studio / Editorial / Pulse: no visual regression (dropdown still anchored correctly to the trigger)
- [x] Click an option → language changes + menu closes
- [x] Click outside the menu → closes
- [x] Escape → closes
- [x] Arrow Up/Down/Home/End keyboard nav still works
- [x] Scroll the Settings page while menu is open → menu closes (no detached floating list)
- [x] Resize window while menu is open → menu closes

## Follow-up (not in this PR)

Same stacking-context trap likely affects other dropdowns under Lounge/Liquid:
- [`PlaylistView.tsx:1327`](src/components/views/PlaylistView.tsx#L1327) — track-row context menu
- [`LibraryView.tsx:940`](src/components/views/LibraryView.tsx#L940) — track-row context menu
- [`EqualizerCard.tsx:197`](src/components/views/settings/EqualizerCard.tsx#L197) — preset picker

Worth porting the same pattern if reported. TopBar dropdowns sit at the top of the app so they're less likely to overlap backdrop-filter cards below.

## Améliorations

* **Améliorations**
  * Le sélecteur de langue s’affiche et se positionne de façon plus fiable, même dans des contextes complexes, grâce à un rendu plus adapté.
  * Il se ferme désormais aussi lors du défilement et du redimensionnement de la fenêtre.
  * Le comportement clavier et la gestion du focus sont renforcés : fermeture et restauration plus cohérentes après une sélection.

### DIFF
--- a/src/components/views/SettingsView.tsx
+++ b/src/components/views/SettingsView.tsx
@@ -318,6 +318,10 @@ function LanguageDropdown({ currentCode, onSelect }: LanguageDropdownProps) {
     const handleKey = (event: KeyboardEvent) => {
       if (event.key === "Escape") {
         setIsOpen(false);
+        // Return focus to the trigger so keyboard users stay anchored
+        // to where they opened the listbox from. Matches WAI-ARIA APG
+        // for combobox/listbox dismissal via Escape.
+        triggerRef.current?.focus();
       }
     };
 
@@ -405,6 +409,13 @@ function LanguageDropdown({ currentCode, onSelect }: LanguageDropdownProps) {
   const handleSelect = (code: string) => {
     onSelect(code);
     setIsOpen(false);
+    // Same restoration as the Escape branch: a deliberate option pick
+    // (mouse click or Enter/Space) lands the keyboard caret back on
+    // the trigger so the user can keep tabbing forward from where
+    // they were. Click-outside is intentionally NOT mirrored here —
+    // there the user explicitly aimed at something else, and stealing
+    // focus back to the trigger would override that intent.
+    triggerRef.current?.focus();
   };
 
   return (
--- a/src/components/views/SettingsView.tsx
+++ b/src/components/views/SettingsView.tsx
@@ -403,6 +403,17 @@ function LanguageDropdown({ currentCode, onSelect }: LanguageDropdownProps) {
     } else if (event.key === "Enter" || event.key === " ") {
       event.preventDefault();
       handleSelect(SUPPORTED_LANGUAGES[index].code);
+    } else if (event.key === "Tab") {
+      // Dismiss the listbox and let the browser's default Tab action
+      // continue from the trigger's position in tab order. Focus is
+      // moved synchronously to the trigger before this handler
+      // returns; the un-prevented Tab default then advances from
+      // there, so Tab lands on the next focusable element AFTER the
+      // trigger and Shift+Tab lands on the previous one. Without this
+      // shim, Tab from a portaled option escapes through document.body
+      // and the keyboard caret falls off the end of tab order.
+      setIsOpen(false);
+      triggerRef.current?.focus();
     }
   };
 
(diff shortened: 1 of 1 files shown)


InstaZDLL:
@coderabbitai applied in 898ac1f.

**Finding** (focus management on close) — **valid**. `handleSelect` and the Escape branch now both call `triggerRef.current?.focus()` after `setIsOpen(false)`, matching the WAI-ARIA APG combobox/listbox dismissal pattern: a deliberate option pick (mouse or keyboard) and Escape both land the caret back on the trigger so the user can keep tabbing forward from where they opened the menu.

**Click-outside** — **skipped on purpose**. The original suggestion asked to mirror the restoration there too, but ARIA APG explicitly carves out outside-click: the user aimed at something else (e.g. a toggle on the next card), so stealing focus back to the trigger would override that intent. Browsers leave focus on whatever the user clicked, and so do native `<select>` elements. The skip is documented inline.

`triggerRef` already existed for the portal positioning (the snapshot rect in `handleTriggerClick`), so this is a 2-line change with no new state.

Validated: `bun run typecheck` clean, `bunx eslint` clean on the touched file.

InstaZDLL:
@coderabbitai applied in 9864cf5.

**Tab keydown handler** — **valid**. Without it, Tab from a portaled option lands in `document.body` (the portal's mount point) and the caret falls off the end of tab order — no element receives focus, the menu stays open visually, and the user has to mouse to recover.

Added a Tab branch to `handleOptionKeyDown` that closes the listbox and focuses the trigger synchronously, *without* `preventDefault`. The browser's default Tab action then advances from the trigger's position in tab order: Tab → next focusable after the trigger, Shift+Tab → previous. Matches WAI-ARIA APG combobox dismissal: "Tab selects the current option and moves focus to the next interactive widget."

**Scroll/resize gated on focus location** — **skipped on purpose**. The reviewer's suggested change would leave the listbox visually detached from the trigger after a scroll (the trigger has moved with the document, but the listbox is anchored by a stale `position: fixed` snapshot). Closing on any layout shift is exactly what the snapshot-positioning contract requires, and it matches what native `<select>` and the radix-ui / headless-ui reference implementations do — they all close on scroll/resize regardless of focus.

Validated: `bun run typecheck` clean, `bunx eslint` clean on the touched file.

Issue #321, opened 2026-06-27
topbar search only surfaces tracks — albums and artists are unreachable


InstaZDLL:
## Summary

The global search bar in the `TopBar` advertises itself as searching titles, artists, and albums (placeholder: *"Rechercher un titre, un artiste, un album..."*) but the dropdown only ever renders a flat list of **tracks**. There is no way to land on an album page or an artist page from the search, even though the underlying FTS index already covers all three.

Reported by a user via email.

## Reproduction

1. Type an album name (e.g. *"Dark Side of the Moon"*) into the top search bar.
2. The dropdown shows only individual tracks from that album.
3. There is no "Albums" section, no "Artists" section, and clicking any row plays the track — there is no way to navigate to the album detail page or the artist detail page.

## Expected behavior

Spotify / Apple Music style sectioned results:

- **Top result** (best match across all entities)
- **Artists** — clicking opens `ArtistDetailView`
- **Albums** — clicking opens `AlbumDetailView`
- **Titles** — clicking plays the track (current behavior)

## Root cause

The backend is fine — `track_fts` already indexes `title + album_title + artist_name` ([`migrations/profile/20260411120000_initial.sql:322`](src-tauri/migrations/profile/20260411120000_initial.sql)), so typing an album or artist name already returns matching rows. The limitation is purely in the frontend:

- [`src/components/layout/TopBar.tsx`](src/components/layout/TopBar.tsx) stores a single `Track[]` (line 73) populated exclusively from `searchTracks` / `searchTracksAdvanced` ([`src/lib/tauri/track.ts`](src/lib/tauri/track.ts)).
- The dropdown (lines 533–562) renders one section, and the only click handler is `playTracks(...)` (line 205) — no `setActiveView('album-detail' / 'artist-detail')` path exists.
- The placeholder string ([`src/i18n/locales/fr.json:213`](src/i18n/locales/fr.json) and equivalents in the other 16 locales) therefore misrepresents what the feature actually does.

## Suggested fix

[...]

PR #330, opened 2026-06-28
feat(artist): offline manual override of bio + similar artists

### LINKED ISSUE #323: offline-first: allow manual override of similar artists and artist bio (no internet required)
## Summary

Reported by a user via email. WaveFlow's offline-first audience (people leaving Spotify specifically to escape cloud-managed metadata) wants the same level of manual control over **artist similar lists** and **artist biographies** that they already have over **artist pictures** (which can be overridden via a local `artist.jpg` / `<artist_name>.jpg` sidecar — see [`scanner::extract_artist_image`](src-tauri/crates/core/src/scanner/extract.rs)).

Today, those two fields are populated exclusively from Deezer + Last.fm + TheAudioDB:
- [`commands/similar.rs`](src-tauri/crates/app/src/commands/similar.rs) calls Last.fm's `artist.getSimilar` and backfills pictures via Deezer
- [`commands/deezer.rs`](src-tauri/crates/app/src/commands/deezer.rs) + [`metadata::lastfm`](src-tauri/crates/core/src/metadata/lastfm.rs) + [`metadata::theaudiodb`](src-tauri/crates/core/src/metadata/theaudiodb.rs) fetch the bio

Two complaints follow:
1. For long-tail repertoires (e.g. Italian artists), Deezer's metadata is sparse or outdated — pictures are bad, similar lists return mostly artists not in the user's library
2. Users who run with **offline mode** on get nothing at all for those fields

## Proposed UX

Per-artist override panel (Artist Detail view → "Edit metadata"):
- **Custom bio** — free-text editor, Markdown-aware. When set, replaces the fetched Last.fm/TheAudioDB bio. A "Reset to online" button drops the override.
- **Similar artists** — pick from artists already in the user's library (autocomplete on `artist.canonical_name`). When set, the list shown in the Artist Detail view becomes user-curated and is filtered/scoped to library-only by design.

Both overrides are per-profile, persist in the per-profile DB, and survive enrichment passes (the Last.fm/Deezer pull paths must check `artist.custom_bio IS NOT NULL` / `artist_similar_custom` before overwriting).

## Sketch

- DB: add `artist.custom_bio TEXT NULL` + new table `artist_similar_custom (artist_id, similar_artist_id, position)` — both per-profile, in a new dated migration under `migrations/profile/`.
- Backend: extend [`enrich_artist_deezer`](src-tauri/crates/app/src/commands/deezer.rs) and the similar-artist read path to short-circuit when overrides exist. New commands `set_artist_bio_override(artist_id, text|null)` + `set_artist_similar_override(artist_id, similar_ids|null)`.
- Frontend: small modal triggered from the existing artist-context-menu in `ArtistDetailView`. Reuse the autocomplete pattern from the playlist "Add tracks" flow for similar-artist selection.

## Out of scope

- Sidecar-file approach for bios (sidecar `artist.txt` next to the audio file). Could be a follow-up, but rapid manual editing is the more common ask — the local-image sidecar exists mostly because covers usually ship with a rip.

InstaZDLL:
Closes #323.

Offline-first users now get the same per-artist manual control over the **biography** and **similar list** they already have over the **picture** (local `artist.jpg` sidecar) — for long-tail repertoires where Deezer/Last.fm metadata is sparse, and for offline-mode users who otherwise get nothing for those fields.

## Backend (per-profile)
New migration `20260628000000_artist_metadata_overrides`:
- `artist.custom_bio TEXT` + table `artist_similar_custom(artist_id, similar_artist_id, position)`.
- [`commands/artist_overrides.rs`](src-tauri/crates/app/src/commands/artist_overrides.rs): `get_artist_overrides` (pre-fill), `set_artist_bio_override`, `set_artist_similar_override` (normalises: drops self + dupes, caps at 50, preserves order).
- [`enrich_artist_deezer`](src-tauri/crates/app/src/commands/deezer.rs) reads `custom_bio` up-front and swaps it onto the returned payload on **every** path (cache hit, offline short-circuit, fresh fetch). The inner path still fetches + caches the online bio so the **shared cross-profile** `metadata_artist` cache stays correct for profiles *without* an override — only the returned value is swapped (avoids clobbering other profiles' cached bio with NULL).
- [`get_similar_artists`](src-tauri/crates/app/src/commands/similar.rs) short-circuits to the curated, library-scoped list (source `"custom"`) before any cache/network — works fully offline; every entry is in-library by construction.

## Frontend
- [`ArtistMetadataEditorModal`](src/components/common/ArtistMetadataEditorModal.tsx), opened from **Artist Detail → "Edit info"**: bio textarea with a reset-to-online affordance, library autocomplete (reuses `search_artists` from #321) + removable chips for the similar list. Saving blank/empty drops the respective override. `onSuccess` resets the bio state and re-runs the enrichment + similar effects so the view reflects the change immediately.

## Notes
- Overrides survive enrichment passes (pull paths write the shared `app.metadata_artist` cache, never the per-profile `artist` row / override table).
- i18n across all 17 locales (`artistDetail.editMetadata`, `artistMetadataEditor.*`).
- **Markdown rendering** of the bio is left as a follow-up — the custom bio renders through the existing plaintext (`whitespace-pre-line`) path, avoiding a new markdown dependency + XSS surface for this PR.
- Validated: `bun run typecheck` ✅ · `bun run lint` ✅ · `cargo clippy -p waveflow --all-targets` ✅.

## Summary by CodeRabbit

* **New Features**
  * Ajout d’un éditeur des métadonnées d’artiste : personnalisation de la bio et des “artistes similaires”.
  * Les modifications sont enregistrées **par profil** et fonctionnent **hors ligne**.
  * Mise en place d’un bouton d’édition depuis la page de détail artiste, avec réinitialisation vers la version en ligne si besoin.
* **Bug Fixes**
  * Les infos bio personnalisées prennent désormais le dessus sur l’enrichissement automatique.
  * Les “similaires” personnalisés s’affichent immédiatement, sans rafraîchissement automatique.
* **Documentation**
  * Mise à jour de la documentation des intégrations pour les overrides manuels.
  * Ajout de traductions d’interface pour l’édition (multi-langues).


InstaZDLL:
@coderabbitai Addressed the artist-override findings in commit 09d53e4:
- **deezer**: pool resolved once in the wrapper and passed into `enrich_artist_deezer_inner` — custom_bio lookup + enrichment now share one profile scope.
- **similar**: curated list capped to `RESULT_LIMIT` before building DTOs.
- **save atomicity**: the two setters are collapsed into one transactional `set_artist_metadata_overrides`; the modal calls it once.
- **modal load**: bio + chips are cleared up-front on open.
- **selection cap**: UI blocks at 50 (mirrors backend) with a visible "limit reached" notice.
- **roles**: dropped `listbox`/`option` from the results — it's now a plain list of action buttons.
- **id.json**: `editMetadata`/`title` localized to the existing "Sunting" convention.

Re: the **stats.rs export snapshot** finding on this PR — that file is only in this diff because this branch was cut off the `fix/export-stats-profile-snapshot` branch (#329), which shares the `64db15e` commit. The atomic-snapshot fix for it lands in **#329** (commit e5318bf); once #329 merges, the stats.rs changes drop out of this PR's diff entirely. Not duplicating it here to avoid a merge conflict between the two branches.

InstaZDLL:
@coderabbitai Done in commit 537a431 — added a shared `isBusy = isLoading || isSaving` that short-circuits the search effect and disables the bio textarea, the reset button, the search input, add/remove, and the results list while either `getArtistOverrides` or `setArtistMetadataOverrides` is in flight. Typed input can no longer be clobbered by a late initial read or submitted as stale state mid-save.

InstaZDLL:
@coderabbitai Both addressed in commit 42b27f7:
- **addArtist**: the `MAX_SIMILAR` cap + dedup now run *inside* the `setSelected` updater, validating against the latest `prev` — rapid clicks can't overshoot the cap or double-add. The busy guard stays outside the updater.
- **search**: results are cleared the moment a new request starts (right after `setIsSearching(true)`) and again in the `catch` path, so stale entries are never shown or clickable mid-request or after a failure.

InstaZDLL:
@coderabbitai Done in commit d4a8a2b — `requestIdRef` is now bumped in the effect body the moment `query` changes, before the debounce timeout. The captured `requestId` is still checked inside the `searchArtists` promise handlers, so a response from a previous query that resolves during the debounce window now fails the check and can't apply stale hits. The short-query branch and the start/catch result clears are unchanged.

PR #333, opened 2026-06-28
feat(ui): merged immersive view with tabbed control panel + native fullscreen (#328)

### LINKED ISSUE #328: immersive view: merge lyrics into now-playing as a side column + use native window fullscreen
## Summary

Reported by a user via email. The immersive view has three related shortcomings that all point at the same underlying decision:

1. **Not actually fullscreen.** The overlay is a `fixed inset-0 z-100` element inside the Tauri window ([`FullscreenNowPlaying.tsx:142`](src/components/player/FullscreenNowPlaying.tsx)). The OS window chrome (title bar, taskbar) is still visible. There is no `getCurrentWindow().setFullscreen(true)` call. This breaks the immersion the view is named after.
2. **Now Playing and Lyrics are mutually exclusive overlays.** [`PlayerContext.tsx:197`](src/contexts/PlayerContext.tsx) declares `fullscreenOverlay: "nowPlaying" | "lyrics" | null` as a single-slot enum — opening one closes the other. To go from one to the other the user has to toggle. Issue #109 (closed) added a direct lyrics → now-playing button, but the underlying mutex remains: you can't *see both at the same time*.
3. **Lots of empty space on wide displays.** Now Playing centers a `max-w-[min(45vh,26rem)]` cover with `max-w-3xl mx-auto` controls below ([`FullscreenNowPlaying.tsx:245-246`](src/components/player/FullscreenNowPlaying.tsx)). On a 27" 2K/4K display, roughly 30% of the horizontal real estate on each side is dead space — visually unbalanced and a perfect home for the lyrics column.

The user explicitly proposed merging: *"I think it would be better if they were merged together — switch tracks while reading lyrics, instead of having to toggle each time."*

## Proposed UX

**One immersive view, two columns.** Spotify desktop / Apple Music TV-style:

- **Left column** (≈ 60% width): cover hero + track metadata + transport controls + visualizer — basically today's `FullscreenNowPlaying` layout, just no longer centered horizontally.
- **Right column** (≈ 40% width): the synced lyrics scroller from today's `FullscreenLyrics`, with auto-scroll + active-line highlight.
- **Single Escape close, single open trigger.** The "Lyrics" button in the player bar opens the same merged view; the layout doesn't change based on entry point.
- **Graceful fallback.** When no lyrics exist for the current track, the right column shrinks to a discreet "No lyrics — Import / Fetch" CTA instead of being hidden (avoids a jarring full-width relayout every time the track changes).
- **Narrow window collapse.** Below a threshold (~900 px wide), fall back to the current single-view-with-toggle behavior — the dual layout only makes sense at desktop widths.

## Native fullscreen

Wire `getCurrentWindow().setFullscreen(true)` on overlay open / `setFullscreen(false)` on close. This needs to be reversible if the user already had the window maximized (restore that state on exit, not just "unset fullscreen"). Should also be opt-out via a Settings → Appearance toggle for users on multi-monitor setups who explicitly *want* to see the rest of their OS while the immersive view runs on one screen — the previous overlay-only behavior is a legit use case, not just a bug.

## Files in scope

- [`src/components/player/FullscreenNowPlaying.tsx`](src/components/player/FullscreenNowPlaying.tsx) — split into a left-column component
- [`src/components/player/FullscreenLyrics.tsx`](src/components/player/FullscreenLyrics.tsx) — extract the scroller into a right-column component reusable by the merged view
- New parent `ImmersiveView.tsx` orchestrating the two columns + the narrow-window fallback
- [`src/contexts/PlayerContext.tsx`](src/contexts/PlayerContext.tsx) — collapse `fullscreenOverlay` enum to a boolean `immersiveOpen` (keep the two old `openFullscreenNowPlaying` / `openFullscreenLyrics` exports as aliases so callsites don't break)
- Tauri side: add an opt-in `app_setting['immersive.use_native_fullscreen']` (default ON) and a one-time prior-window-state cache for clean restore.

## Out of scope

- Karaoke / word-level lyric animation behavior — unchanged, the right-column scroller reuses the existing component.
- Issue #109's lyrics → now-playing direct button — becomes obsolete with the merge (the user no longer needs to switch).

InstaZDLL:
## Summary

Closes #328. Reworks the immersive view to address the reporter's two points (not actually fullscreen; now-playing and lyrics are mutually-exclusive overlays with dead side space) — but ships them as **opt-in, per-profile options** so the out-of-the-box experience is unchanged.

Both toggles live in **Settings → Appearance** (`ImmersiveViewCard`), persist in `profile_setting`, and **default OFF**:

| Toggle | Off (default) | On |
|---|---|---|
| **Lyrics side column** (`immersive.merged_lyrics`) | Classic immersive — now-playing fullscreen with a Mic2 button flipping to a lyrics-only fullscreen (pre-#328 behaviour, intact) | Merged 2-column view: now-playing left + a tabbed control panel (Lyrics / Queue) right, `PanelRight` toggle to hide/show |
| **Native fullscreen** (`immersive.use_native_fullscreen`) | In-window overlay (as before) | Real OS fullscreen (`setFullscreen(true)`), restoring the prior maximized state on close |

Narrow windows (< 900 px) force the classic interface even when merged is on.

## Architecture

- **`useTrackLyrics`** — extracts the full lyrics lifecycle (three-tier fetch + parse + active-line/word tracking + import/refetch/clear, with every mid-flight staleness guard) from `LyricsPanel` into a shared hook. Both `LyricsPanel` and the immersive view consume it, so there's no second fetch path and the old `createPortal`-from-panel coupling is gone.
- Components: `ImmersiveView` (orchestrator + reversible native-fullscreen effect + narrow fallback) → `ImmersiveNowPlaying`, `ImmersiveSidePanel` (tabs) → `ImmersiveLyricsColumn` + `ImmersiveQueueTab`, `ImmersiveShareButton`.
- `ImmersiveQueueTab` reuses `player_get_queue` / `player_jump_to_index` (white-on-dark styled); the docked `QueuePanel` keeps its drag-reorder.
- `PlayerContext`: `fullscreenOverlay` enum → `immersiveOpen` bool + `immersiveInitialTab`; old `openFullscreenNowPlaying` / `openFullscreenLyrics` / `close*` kept as aliases.
- Native fullscreen needs `core:window:allow-set-fullscreen` / `allow-is-fullscreen` / `allow-is-maximized` / `allow-maximize` — added to `capabilities/default.json` (compiled into the binary, so it needs a Rust rebuild).
- Removes `FullscreenNowPlaying` + `FullscreenLyrics`.

## Notes

- i18n propagated to all 17 locales; docs (CLAUDE.md, `docs/features/ui.md`, `integrations.md`) synced.
- Onboarding "done" step gets a discreet tip about the immersive view.

## Validation

`bun run typecheck` ✅ · `bun run lint` ✅ · `bun run build` ✅

## Summary by CodeRabbit

  * Nouvelle **vue immersive plein écran** fusionnant “Now Playing” et les paroles en **deux colonnes** (responsive) avec onglets **Paroles / File**.
  * Option **plein écran natif** (ouverture/retour automatique).
  * **Carte de partage** Now Playing dédiée à la vue immersive (enregistrer ou copier).
* **Améliorations**
  * Paroles : cycle de vie unifié via une logique partagée (import / rechargement / effacement / seek), avec synchronisation et mise en évidence de la ligne active.
  * File immersive mise à jour en temps réel avec rafraîchissements mieux contrôlés.
* **Documentation**
  * Mise à jour de l’UI immersive, du rendu “word-level lyrics” et des textes d’onboarding.


InstaZDLL:
@coderabbitai thanks — all findings were valid, fixed in 37767a7:

- **useTrackLyrics**: `applyPayload` now clears `error` when a non-null payload is applied (mirrors `importLyrics`); the `trackId == null` branch clears `isFetching` too; `clear()` got the same `requestedTrackId !== trackIdRef.current` staleness guard as `importLyrics`/`refetch`.
- **useImmersivePrefs**: the effect now depends on the active profile id (re-reads per-profile prefs on switch) and a monotonic `seqRef` drops out-of-order async reads before `setPrefs`.
- **ImmersiveView**: after `setFullscreen(true)` resolves, if the cleanup already ran (`cancelled`) we undo the fullscreen + restore the prior maximized state instead of leaving the window stuck fullscreen.
- **ImmersiveQueueTab**: the fetch `catch` only logs now (keeps the last queue snapshot instead of blanking it); the `listen` setup tracks a `disposed` flag and tears the subscription down immediately if it resolves after unmount.
- **ImmersiveShareButton**: `type="button"` on both menu actions.
- **PlayerBar + PlayerContext**: dropped the `currentTrack` mount guard so the view unmounts only via `closeImmersive`, and the profile-reset effect now sets `immersiveOpen = false` so it can't linger through the `currentTrack` null window and reopen on the next track.
- **docs**: the `ui.cover_action` `immersive` row no longer promises a guaranteed merged/fullscreen view — it states the merged columns + native fullscreen are opt-in, classic by default.

typecheck / lint / build all green.

InstaZDLL:
@coderabbitai both valid, fixed in 3d804be:

- **useImmersivePrefs**: the effect now resets `prefs` to `{ ...DEFAULTS, loaded: false }` up front, so a profile switch never surfaces the previous profile's values during the async re-read. It's a no-op on first mount (already DEFAULTS) and isn't run on the custom `IMMERSIVE_PREFS_EVENT` (that calls `refresh` directly), so toggling a setting doesn't flash the card to defaults.
- **useTrackLyrics** `clear()`: after a successful `clearLyrics` + the staleness guard, it now also `setError(null)` so a prior fetch error can't mask the empty state. Also switched the call to `clearLyrics(requestedTrackId)` for consistency with the captured id.

typecheck / lint green.

InstaZDLL:
@coderabbitai valid, fixed in 98e3b8c — the `refresh` catch now publishes `setPrefs({ ...DEFAULTS, loaded: true })` (guarded by `cancelled`/`seq`) so a failed Tauri read still resolves `loaded` instead of leaving it stuck at the pre-read `false` reset. typecheck + lint green.

InstaZDLL:
@coderabbitai valid, fixed in the latest push — the cover-click button now uses `t("playerBar.nowPlaying")` ("Vue en cours de lecture" / "Now Playing view") instead of `playerBar.openFullscreen`, since `onCoverClick={toggleClassicLyrics}` flips back to the now-playing view rather than opening fullscreen. Reused the existing key (present in all 17 locales) so no new i18n. typecheck + lint green.

PR #334, opened 2026-06-28
fix(ui): immersive view skin neutrality + long-title marquee

InstaZDLL:
## Summary

Polish pass on the immersive view + skins, reported visually during testing.

### Immersive view was wrecked by the skins
The immersive view (`role="dialog"` + a `shadow-2xl` cover) was captured by every skin's modal / surface / text-colour chrome:
- **Liquid (light)** + **Editorial** repainted it as a light glass slab with dark, invisible text and washed-out (grey-on-grey) controls.
- **Lounge / Pulse** flattened the cover backdrop.

Fix: the root is tagged `data-immersive` + a nested `dark` context (so the shared `PlaybackControls` / `VolumeControl` / `ProgressBar` render their dark-theme variants over the always-dark backdrop), and each skin's dialog / surface / text-colour rules carry `:not([data-immersive]):not([data-immersive] *)`. The view now renders identically across all skins.

### Long titles scroll instead of truncating
New [`MarqueeText`](../../src/components/common/MarqueeText.tsx) — measures overflow via `ResizeObserver` and, only when the text actually overflows, glides it end-to-end and back (ping-pong with a pause at each end, `prefers-reduced-motion` respected). Wired into the now-playing title + the bottom `PlayerBar` title.

- Per-profile **Settings → Appearance** toggle (`ui.scroll_long_titles`, default ON) via `useScrollLongTitles` + `ScrollTitlesCard`; off = ellipsis truncation.
- Also fixes descenders (`g` / `y` / `p`) being clipped in the immersive title.

### Notes
- i18n propagated to all 17 locales; `docs/features/ui.md` updated.
- Validated visually across Liquid (light + dark), Editorial, Lounge, Pulse.

## Validation
`bun run typecheck` ✅ · `bun run lint` ✅ · `bun run build` ✅

## Summary by CodeRabbit

  * Les titres longs peuvent défiler en mode “marquee” dans la vue immersive et la barre du lecteur.
  * Nouveau réglage par profil pour activer/désactiver ce défilement (activé par défaut).
* **Améliorations**
  * Le défilement ne démarre que si le titre déborde réellement, et il respecte “réduire les animations”.
  * La vue immersive est mieux isolée des styles de thème (modales/surfaces/couleurs) pour un rendu plus cohérent.
* **Documentation**
  * Mise à jour des notes sur la vue immersive et le défilement des titres.
* **Traductions**
  * Nouvelles chaînes ajoutées pour l’option dans les langues disponibles.


InstaZDLL:
@coderabbitai both valid, fixed:

- **MarqueeText**: the render branch now gates on `prefers-reduced-motion` (via framer-motion's `useReducedMotion`) alongside `enabled` + `shift` — when motion is reduced, `scroll` is false so it takes the static `truncate` path (ellipsis + inherited centering) instead of the scrolling layout whose CSS animation was disabled but whose left-aligned nowrap layout remained.
- **useScrollLongTitles**: `setEnabled` now bumps a monotonic `writeSeqRef` and captures the rollback target from an `enabledRef` mirror; only the latest write applies its success (dispatch) / failure (rollback) side effects, so an older overlapping response can't clobber the newer intent.

typecheck + lint + build green.

InstaZDLL:
@coderabbitai valid, fixed — the Tauri writes are now serialized on a `writeChainRef` promise chain (each `setProfileSetting` runs after the previous resolves, `.catch` keeps the chain alive on failure), so the backend applies toggles in issue order and the last write wins. On success the written value is recorded in `confirmedEnabledRef`; a failed write now rolls back to that confirmed value instead of the optimistic `enabledRef`. The read path (`refresh`) seeds both refs too. typecheck + lint green.

InstaZDLL:
@coderabbitai valid, fixed in 104cd0f — `setEnabled` now captures `activeProfileIdRef.current` at dispatch time, and the queued chain task aborts (skips the `setProfileSetting` + the `confirmedEnabledRef` update) if the active profile changed while it waited, so a toggle issued on profile A can't persist onto profile B after a switch. The refresh effect already re-reads per-profile on `activeProfile?.id`, so the new profile's value is reflected correctly. typecheck + lint green.

InstaZDLL:
@coderabbitai valid, fixed in cf89452 — added a profile re-check after the `setProfileSetting` await (before confirming the value) and at the top of both the success and catch paths, so a profile switch mid-write can no longer confirm the old profile's value nor roll back / dispatch onto the new profile. Only the profile the toggle was issued for applies side effects. typecheck + lint green.

Issue #175, opened 2026-05-28
bug(audio): cpal DeviceNotAvailable mid-stream causes channel-count flip-flop


InstaZDLL:
## Symptom

Mid-stream the cpal callback returned `DeviceNotAvailable` without any user-driven device switch. The engine reopened the output and the new stream came up with **different format parameters** than the original, then flipped back a second later, interrupting the same track three times in ~14 seconds:

```
20:26:30  cpal output stream opened sample_rate=44100 channels=8 sample_format=F32
…
20:29:01  WARN cpal stream error err=DeviceNotAvailable
20:29:13  cpal output stream opened sample_rate=48000 channels=2 sample_format=F32   ← different SR + channels
20:29:13  play_track interrupted finished_track_id=215 listened_ms=113557
20:29:13  decoder picked up new ring producer
20:29:14  cpal output stream opened sample_rate=44100 channels=8 sample_format=F32   ← flipped back
20:29:14  play_track interrupted finished_track_id=215 listened_ms=280               ← interrupted again
20:29:14  decoder picked up new ring producer
```

User reported no manual device change — Windows audio session reset itself.

## Root cause hypothesis

Two layers:

1. **Trigger**: Windows can issue a session-invalidation event when a default device changes / a driver restarts / a Bluetooth source flaps. cpal surfaces this as `DeviceNotAvailable` on the stream's error callback.

2. **Response**: [`audio/output.rs`](https://github.com/InstaZDLL/WaveFlow/blob/main/src-tauri/crates/app/src/audio/output.rs) handles the error by rebuilding the stream. The rebuild path **re-queries the host default device**, but the user-pinned device name (if any) and the previous channel layout aren't preserved across the rebuild — so when the rebuild lands on a different default than before, the channel count and SR change unexpectedly.

3. **Second flip**: a few ms later something rebuilds again. Possibly the engine's own `set_output_device` path firing because the persisted `audio.output_device` setting points at the original device which has come back online. That second `LoadAndPlay` interrupts the track barely 280 ms in.

## Fix sketch

In the cpal error handler in `audio/output.rs`:

1. Capture the original `(device_name_pinned, sample_rate, channels)` triple at first stream open
2. On `DeviceNotAvailable`:
   - Wait a short backoff (250-500 ms) to give the OS time to settle
   - Attempt to reopen the **same** pinned device first; only fall back to "OS default" after N retries
[...]

PR #346, opened 2026-07-04
fix(audio): recover from WASAPI exclusive device-reset flap storm (#322)

### LINKED ISSUE #322: wasapi exclusive: engine stops responding to play when toggled on, no fallback to shared mode
## Summary

Reported by a user via email. On Windows, enabling **WASAPI Exclusive** in Settings → Playback puts the player into a frozen state: pressing Play has no effect, the track stays at 00:00, and no `player:error` is emitted. The cpal shared-mode fallback that [`audio/wasapi_exclusive.rs`](src-tauri/crates/app/src/audio/wasapi_exclusive.rs) is supposed to trigger on init failure does not appear to fire either.

Reproduces on the user's setup; works correctly on the same DAC in MusicBee, so it's not a hardware/driver issue per se.

## Reproduction (to confirm with reporter)

1. Settings → Playback → enable **WASAPI Exclusive**
2. Pick any track
3. Press Play
4. Expected: audio plays bit-perfect, or the engine falls back to shared mode + surfaces a `player:error` toast
5. Observed: nothing — the track stays at 00:00, no error, no audio

## Notes

- Init is synchronous in [`spawn_exclusive_output_thread`](src-tauri/crates/app/src/audio/wasapi_exclusive.rs) and surfaces failures back through `init_rx`, so a genuine init failure should already trigger the fallback path. If the engine is frozen without an error it likely means **init succeeded but the audio thread is blocked** somewhere downstream — possibly:
  - mix-format negotiation picked a `WaveFormat` the DAC accepts on paper but never drains the SPSC ring
  - event-driven `IAudioClient::SetEventHandle` is fired but the buffer-fill loop misses the first event
  - the decoder thread is blocked waiting for the producer to drain (RING_CAPACITY full)
- Worth adding more aggressive `tracing::warn!` around the first N callback invocations to localize the stall.

## Info needed from reporter

- Windows version (winver)
- DAC make/model + driver source (manufacturer vs Windows Update)
- What the Windows Sound control panel shows for that device's default format (e.g. `24-bit, 96000 Hz`)
- Whether the lib has a mix of sample rates / bit depths or is uniform
- `%APPDATA%\waveflow\logs\` excerpt covering the moment WASAPI Exclusive was toggled and Play pressed

InstaZDLL:
## What & why

Fixes the WASAPI Exclusive freeze in #322. New reporter logs pinned the root cause.

**Root cause** — some onboard codecs (Realtek HD Audio here) reset the device right after an exclusive grab, firing cpal `DeviceNotAvailable` ~300 ms later. The same-device auto-rebuild (#175) then re-opened exclusive while the **old exclusive client still owned the device**, so:

- the re-init failed with `AUDCLNT_E_DEVICE_IN_USE` (`0x8889000A`), and
- even the shared fallback failed with *"device no longer available"*.

Re-engaging exclusive re-armed the reset → infinite loop. Symptom: playback stalls, and turning the toggle off takes many clicks.

```
wasapi exclusive stream opened … format="S24_4LE"   ← engages
rebuilding cpal output after DeviceNotAvailable exclusive=true
wasapi exclusive init failed … HRESULT(0x8889000A)  ← DEVICE_IN_USE (old client still held it)
WASAPI Exclusive init failed, falling back to shared mode
auto-rebuild after DeviceNotAvailable failed … build_output_stream: The requested device is no longer available
```

## The fix (`audio/engine.rs`)

1. **Release-before-reopen** — in `force_rebuild_output`, when the *old* stream is exclusive, stop it **before** spawning the new one. A WASAPI exclusive client locks the device against *every* other client (exclusive or shared), so the old handle must go first. Fixes both error lines above. Only runs after a device error (old stream already dead); shared-old keeps the spawn-first order so a failed spawn can still roll back.
2. **Flap-storm suppression** — count exclusive `DeviceNotAvailable` rebuilds; past **3 within 12 s**, disable exclusive for the rest of the session and settle on shared so the device stops thrashing. Session-only: the persisted preference is untouched (a fresh launch retries exclusive), and an explicit device switch or exclusive re-toggle clears it.

Context: `rustmusic` (our audio reference) sidesteps this whole class by never using WASAPI exclusive — pure cpal shared. So there was nothing to borrow; the fix is ours. Their design does validate fix #2 (falling back to shared is what every other player effectively does).

## ⚠️ Needs hardware testing before merge

I **cannot reproduce WASAPI exclusive locally**, so this is reasoned from the logs + code, not verified end-to-end. `cargo check` + `clippy` are clean, but please don't merge until the reporter (or someone with a device that flaps on exclusive) confirms a test build:

- turn WASAPI Exclusive **on**, press play → audio should either play in exclusive, or fall back to shared and **keep playing** (no stall);
- if it can't hold exclusive, after a few seconds it should log `WASAPI exclusive disabled for this session after repeated device flaps` and stay on shared;
- toggling exclusive **off** should be immediate (no multi-click).

## Docs

No user-facing behaviour/setting change (the existing "transparent fallback to cpal shared" still holds), so docs are untouched. Will add a one-liner to `docs/features/playback.md` at merge if we want to note the flap-suppression.

Closes #322 (pending reporter confirmation).

## Summary by CodeRabbit

  * Ajout d’une protection contre les basculements répétés du mode audio exclusif lors des reconnections du périphérique (WASAPI exclusif).

* **Correctifs**
  * Réduction des erreurs « périphérique déjà utilisé » pendant le hot-swap du périphérique.
  * Si le mode exclusif échoue trop souvent sur la session, l’application bascule temporairement en mode partagé pour préserver la lecture.
  * Réinitialisation plus fiable de la suppression du mode exclusif lors d’un changement de périphérique ou d’un réglage utilisateur.

### DIFF
--- a/src-tauri/crates/app/src/audio/engine.rs
+++ b/src-tauri/crates/app/src/audio/engine.rs
@@ -166,6 +166,26 @@ struct FlapWindow {
     window_start: Option<Instant>,
 }
 
+impl FlapWindow {
+    /// Record one flap at `now` and report whether it trips the storm
+    /// threshold ([`EXCLUSIVE_FLAP_THRESHOLD`] flaps within
+    /// [`EXCLUSIVE_FLAP_WINDOW`]). A flap outside the current window
+    /// restarts the count at 1. Clock is injected so the logic is
+    /// deterministic under test.
+    fn record(&mut self, now: Instant) -> bool {
+        let within_window = self
+            .window_start
+            .is_some_and(|start| now.duration_since(start) <= EXCLUSIVE_FLAP_WINDOW);
+        if within_window {
+            self.count += 1;
+        } else {
+            self.window_start = Some(now);
+            self.count = 1;
+        }
+        self.count >= EXCLUSIVE_FLAP_THRESHOLD
+    }
+}
+
 /// Handle stored in Tauri state. Cloning an `Arc<AudioEngine>` is cheap.
 ///
 /// The cpal `Stream` is NOT stored here — it lives on a dedicated output
@@ -481,25 +501,15 @@ impl AudioEngine {
         let mut exclusive =
             pref_exclusive && !self.exclusive_suppressed.load(Ordering::Relaxed);
         if exclusive {
-            let mut flaps = self
+            let tripped = self
                 .exclusive_flaps
                 .lock()
-                .unwrap_or_else(|poisoned| poisoned.into_inner());
-            let now = Instant::now();
-            let within_window = flaps
-                .window_start
-                .is_some_and(|start| now.duration_since(start) <= EXCLUSIVE_FLAP_WINDOW);
-            if within_window {
-                flaps.count += 1;
-            } else {
-                flaps.window_start = Some(now);
-                flaps.count = 1;
-            }
-            if flaps.count >= EXCLUSIVE_FLAP_THRESHOLD {
+                .unwrap_or_else(|poisoned| poisoned.into_inner())
+                .record(Instant::now());
+            if tripped {
                 self.exclusive_suppressed.store(true, Ordering::Relaxed);
                 exclusive = false;
                 tracing::warn!(
-                    flaps = flaps.count,
                     "WASAPI exclusive disabled for this session after repeated device \
                      flaps; staying on shared mode. Re-enable it in Settings to retry."
                 );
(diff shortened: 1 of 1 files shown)


InstaZDLL:
@coderabbitai valid — fixed in fb53be0.

Extracted the window/count update into `FlapWindow::record(&mut self, now: Instant) -> bool` (clock injected for determinism). `try_rebuild_after_device_error` now just calls `record()` and keeps only the suppression + warning decision. Added `flap_window_tests` covering: threshold trip within the window, reset when a flap lands outside the window, and no trip for spaced-out flaps.

`cargo check` + `clippy --tests` clean and the tests compile; they can't run locally on Windows (`STATUS_ENTRYPOINT_NOT_FOUND` — the known Tauri-DLL harness issue) but run on CI.

PR #392, opened 2026-07-18
fix(ui): portal context menus and raise the TopBar layer

### LINKED ISSUE #390: bug: Theme-specific overlay issues with UI elements
### WaveFlow version

1.6.0

### Operating system

macOS (Apple Silicon)

### How did you install WaveFlow?

GitHub Releases (.dmg)

### What happened?

The overlay order of some UI elements is incorrect in certain themes.

Expected:
Menus (such as the track context menu and profile menu) should always be rendered above other UI elements.

Actual:
- In the Pulse and Liquid themes, the playback bar is rendered above the track context menu, hiding the lower menu items (such as "Properties").
- In the Studio theme, the date header in the "Recently Played" playlist is rendered above the profile menu.

### Steps to reproduce

1. Open WaveFlow.
2. Switch to the Pulse theme.
3. Open any playlist.
4. Right-click a track near the bottom of the list.
5. Observe that the playback bar overlaps the context menu, making the lower menu items inaccessible.

Additional observations:
- The same issue occurs with the Liquid theme.
- With the Studio theme, opening the profile menu while viewing the "Recently Played" playlist causes the date header to overlap the menu.

### Relevant logs / errors

```shell
No relevant errors were found in the application logs.
This appears to be a UI rendering / overlay issue.
```

### Anything else?

Image
Image
Image

InstaZDLL:
Closes #390.

Two symptoms reported on different skins turned out to have **one root cause**: `backdrop-filter` creates a **stacking context**, and a `z-index` is only ever compared inside its own context. `position: fixed` escapes layout flow but *not* the stacking context — so a nested overlay at `z-100` can still paint under a `z-50` sibling of its trapping ancestor.

## Symptom 1 — context menu hidden behind the PlayerBar (Pulse + Liquid)

`ContextMenu` is `position: fixed` at `z-100`, but it was rendered inline in the track-list subtree. The Pulse / Liquid skins put `backdrop-filter` on chrome containers (e.g. [`pulse.css`](src/styles/skins/pulse.css) `:root[data-skin="pulse"] footer { backdrop-filter: blur(20px) }` — that `footer` *is* the PlayerBar, `z-50`), which trapped the menu and made the bottom items ("Properties") unreachable. Studio has no skin CSS, which is exactly why it was unaffected.

**Fix:** `ContextMenu` now renders through `createPortal(…, document.body)`, so it no longer depends on any ancestor's stacking context. It was already `position: fixed` with viewport coordinates, so the positioning math is unchanged.

Skin styling is preserved — every skin rule is rooted at `:root[data-skin="…"] :where(…)`, and `body` is still a descendant of `:root`, so the portalled subtree keeps matching. The `[data-context-menu]` click-outside guard also still works, since it uses `closest()` on the real DOM.

## Symptom 2 — Recently Played date header over the profile menu (Studio)

[`TopBar.tsx`](src/components/layout/TopBar.tsx) carries `backdrop-blur-md`, so the header is itself a stacking context and **clamps every dropdown inside it** (profile menu, notifications, search results) to the header's own z-index — regardless of their `z-50`. The header sat at `z-10`, while the content-level sticky group header in [`HistoryView.tsx`](src/components/views/HistoryView.tsx) is `z-20`. So `20 > 10` and the date header painted over the open profile menu.

**Fix:** header raised to `z-30` — above content chrome, still below the PlayerBar (`z-50`).

No collision introduced: the other `z-30` users are either inside portalled modals (`CreatePlaylistModal`) or vertically centred on the right edge (`NowPlayingChevronTab`, `LibraryView` scrubber), so they never overlap the header strip.

## Prevention

Rather than leaving this as two spot-fixes, the layer scale is now documented at the top of [`src/app.css`](src/app.css):

```
z-10   in-panel sticky headers
z-20   content-level sticky group headers
z-30   TopBar (a stacking context — its dropdowns are clamped here)
z-50   PlayerBar (a stacking context under Pulse/Liquid)
z-100  overlays: context menus, dropdown popovers
z-101  nested submenus
```

with the rule that **anything at z-100+ must be portalled to `document.body`**, plus a matching cross-cutting entry in `CLAUDE.md` so the next overlay portals instead of escalating numbers.

## Checks
- `bun run typecheck` ✅
- `bun run lint` ✅
- `bun run build` ✅

> **Not visually verified in-app.** This is a GUI interaction bug (right-click near the bottom of a list in Pulse/Liquid; open the profile menu over Recently Played in Studio) and I couldn't drive the desktop window from my environment. The diagnosis is traced to exact lines and the reasoning is deterministic, but a quick manual pass on the three skins before merge would be worthwhile — ideally @ukine888 can confirm on macOS.

## Summary by CodeRabbit

* **Améliorations**
  * Les menus contextuels s’affichent désormais correctement au-dessus des autres éléments, même dans des interfaces complexes.
  * La barre supérieure reste visible au-dessus du contenu et de la barre de lecture, avec un meilleur affichage des menus déroulants associés.

* **Documentation**
  * Clarification des règles de superposition des fenêtres, menus et popovers pour garantir un rendu cohérent.


InstaZDLL:
@coderabbitai review

ukine888:
Confirmed on macOS! I tested v1.6.4-beta.1, and both issues are completely fixed.
Thanks for the incredibly fast fix, and for the detailed explanation of the root cause! The explanation about stacking contexts was also really interesting.

PR #394, opened 2026-07-18
fix(ui): persist window bounds only in the normal state, dedupe motion artwork lookups

InstaZDLL:
Two fixes found while testing the motion-artwork pipeline.

## 1. Window bounds were persisted while maximized

`useMainWindowBounds` saved on every `onMoved` / `onResized` with no check on window state. Maximizing therefore stored the full work-area geometry, and the Rust restore path applied that **size** without the maximized **state** — so the next launch opened a window that *looked* maximized but wasn't, and could no longer be restored to its previous size.

Minimized was worse: Windows reports an off-screen position (typically `-32000`), which would have parked the window outside every monitor.

The write is now skipped when the window is maximized, minimized or fullscreen, so what persists is always the last real user-sized geometry.

## 2. Motion artwork resolved 3–4× per album

`useAlbumMotionArtwork` is mounted by several surfaces at once (`ImmersiveNowPlaying` + `NowPlayingPanel`), so a single track change fired three to four identical `fetch_album_motion_artwork` calls for the same album. Each is a full fan-out to every enabled metadata plugin, taking that plugin's lock and — when cold — hitting Apple's API. Straight from a dev-session log:

```
18:20:03.450 … album=Better
18:20:04.093 … album=Better
18:20:08.005 … album=Better
18:20:08.631 … album=Better      ← 4× for one album
18:21:46.496 … album=Short n' Sweet
18:21:46.534 … album=Short n' Sweet
18:21:47.260 … album=Short n' Sweet
```

Added a process-wide dedupe: concurrent callers share one in-flight promise, and resolved answers are kept so remounting a panel or flipping back to a recent album costs nothing. The resolved map is capped at 64 entries and evicted oldest-first, so a long session can't retain one entry per album played. Failures are deliberately **not** remembered — they're usually transient (offline, rate limit) and caching them would suppress retries for the rest of the session.

## Checks
- `bun run typecheck` ✅
- `bun run lint` ✅

> The window-state fix is behavioural and worth a quick manual pass: maximize, close, relaunch — the window should come back at its last *normal* size, not maximized-but-not-maximized.

## Summary by CodeRabbit

* **Améliorations**
  * Le chargement des illustrations animées d’albums est désormais plus rapide et plus efficace, notamment lors de montages répétés ou de demandes simultanées.
  * Les résultats disponibles sont réutilisés temporairement afin d’éviter les requêtes inutiles.

* **Corrections**
  * La position et la taille de la fenêtre ne sont plus enregistrées lorsqu’elle est maximisée, réduite ou en plein écran.
  * La restauration des dimensions normales de la fenêtre est ainsi plus fiable.


InstaZDLL:
@coderabbitai review

Issue #378, opened 2026-07-14
feat: Reading the SYNCEDLYRICS tag as primary


NielsHoenders:
### What problem does this solve?

When downloading tracks from Antra, it creates a "SYNCEDLYRICS" and "lyrics" tag visible in kid3. Waveflow currently only reads the "lyrics" tag. A nice feature would be to have it read the synced lyrics tag as a priority.

<img width="994" height="1173" alt="Image" src="https://github.com/user-attachments/assets/aa600fc9-3c15-4677-bac8-7853f4d1b36d" />

### Proposed solution

Have WaveFlow read "SYNCEDLYRICS" as the primary lyrics tag.

### Alternatives considered

Have the option to use LRCLIB even when lyrics are embedded in the file. This seems like a nice idea anyways if lyrics are perhaps wrong or outdated. 

<img width="1503" height="901" alt="Image" src="https://github.com/user-attachments/assets/7385b323-2bd8-4bb5-a48e-093490a2991a" />

Basically having the already integrated option menu as an option even when lyrics are embedded through meta tags. 

<img width="1501" height="896" alt="Image" src="https://github.com/user-attachments/assets/7cfe0353-f59b-4a3f-8d61-e14446b47300" />

### Rough scope

Small (one card / one command / one menu entry)

### Anything else?

When editing lyrics in WaveFlow it creates an "UNSYNCEDLYRICS" tag and removes both "SYNCEDLYRICS" and "Lyrics"? Even when creating synced lyrics with timestamps the tag is called "UNSYNCEDLYRICS" so possibly both a bug and a feature in this one :) 

<img width="994" height="1171" alt="Image" src="https://github.com/user-attachments/assets/18de7de5-231a-4675-b907-7dc19361923a" />

PR #395, opened 2026-07-18
fix(lyrics): stop caching a provider outage as a permanent miss

### LINKED ISSUE #391: Lyrics fallback chain: a provider outage is cached as a permanent "no lyrics" miss
Follow-up from the review of #389 ([thread](https://github.com/InstaZDLL/WaveFlow/pull/389#discussion_r3606130433)).

## Problem

In the post-LRCLIB fallback tier, a **provider outage is indistinguishable from "this track has no lyrics"**, and gets cached as a permanent miss.

[`try_external_fallback`](../blob/main/src-tauri/crates/app/src/commands/lyrics.rs) collapses a transient `Err` into `Ok(None)`:

```rust
Err(err) => {
    tracing::debug!(?err, "external fallback chain failed");
    Ok(None)
}
```

`resolve_after_lrclib_miss` then treats that `None` as a confirmed miss and calls `cache_lyrics_miss`, writing an empty `lyrics` row. Because the waterfall is cache-first and never re-fetches once a row exists, a one-off network blip permanently marks the track as "no lyrics" until the user manually hits Refetch.

This is **pre-existing behaviour**, not a regression: the former `try_external_fallback_or_miss` did the same thing, and its doc comment codified it ("both must treat a provider Err as non-fatal — LRCLIB has already given a verdict, the fallback is best-effort enrichment"). #389 preserved the semantics deliberately and split the function; this issue is about revisiting the policy itself.

## Proposal

Model the three outcomes distinctly at that tier, e.g.:

```rust
enum FallbackOutcome { Found(LyricsPayload), Miss, Unavailable }
```

- `Found` → cache + return (unchanged).
- `Miss` → cache the empty row (unchanged — a confirmed "nobody has it").
- `Unavailable` (transient provider/network error) → **do not cache**; return no lyrics for this attempt so the next panel open retries.

Care needed so the "don't re-hit the network on every panel open" property isn't lost for the genuine-miss case — only the error case should stay retryable.

## Notes

- The real LRCLIB tier already surfaces its own network failures as `Err` (see the `LRCLIB fetch failed` branch), so this only concerns the best-effort fallback chain.
- Worth checking the same policy in the bulk `run_prefetch` loop, which has its own inline copy of the fallback handling.
- With `lyrics.prefer_lrclib` on (#378 / #389), an error at this tier already falls through to the local embedded/sidecar tiers, so the toggle-on path partly mitigates this today.

InstaZDLL:
Closes #391.

## The bug

The post-LRCLIB fallback chain (NetEase / Megalobiz / Genius) collapsed a **provider error** into the same `Ok(None)` it used for a **real negative**, and the caller then wrote an empty `lyrics` row.

Because the waterfall is cache-first and never re-fetches once a row exists, a single network blip marked the track as **permanently lyric-less** — only a manual "Refetch" could clear it.

## The fix

```rust
enum FallbackOutcome { Found(LyricsPayload), Miss, Unavailable }
```

- **Found** → cached, returned (unchanged).
- **Miss** → every provider answered and none had lyrics. Still caches the empty row, so the "don't re-hit the network on every panel open" property is preserved.
- **Unavailable** → the chain couldn't be consulted. **Persists nothing** and returns no lyrics for this attempt, so the next open retries.

Offline folds into `Unavailable` as well: `external_lyrics_search` also short-circuits to `Ok(None)` when offline mode is on. `fetch_lyrics` checks offline before reaching this tier, but the flag can flip mid-waterfall — and *"couldn't ask"* must never be recorded as *"nobody has it"*.

Under `lyrics.prefer_lrclib` (#378) the deferred local tiers still get their turn whether the chain missed **or** was unreachable, so the toggle can't lose lyrics the file already carries.

## `run_prefetch` deliberately unchanged

The issue flagged the bulk loop as carrying its own inline copy of the fallback handling and suspected it needed the same treatment. Checked — it already gets this right:

```rust
Ok(Some(result)) => { … hits += 1; }
Ok(None)         => { upsert empty row; misses += 1; }
Err(err)         => { tracing::warn!(…); failed += 1; }   // ← no caching
```

Both of its fallback sites count an error as `failed` **without** persisting anything, and `prefetch_library_lyrics` refuses to run at all in offline mode. So the bug was specific to the on-demand path.

## Checks
- `cargo check -p waveflow --all-targets` ✅
- `cargo clippy -p waveflow --all-targets` ✅ (no warnings)

> No unit test: exercising this needs a live profile pool plus a way to make a provider fail, and the app-crate test binary doesn't run locally on Windows (`STATUS_ENTRYPOINT_NOT_FOUND`). The change is a pure control-flow split with no new I/O, and both call sites go through the one shared helper.

### DIFF
--- a/src-tauri/crates/app/src/commands/lyrics.rs
+++ b/src-tauri/crates/app/src/commands/lyrics.rs
@@ -367,18 +367,35 @@ fn shared_external_client() -> AppResult<&'static SyncedLyricsClient> {
     }
 }
 
+/// Outcome of one external provider query (#391).
+///
+/// `Miss` and `Unavailable` are kept apart **here**, at the source, rather
+/// than reconstructed by callers: only this function knows whether any
+/// request was actually issued. A caller that tried to infer it afterwards
+/// — say by re-reading the offline flag — would be guessing against state
+/// that can change between the call and the check, and a wrong guess
+/// caches a permanent "no lyrics".
+enum SearchOutcome {
+    Found(ExternalLyricsResult),
+    /// Providers were queried and none had lyrics. A real negative.
+    Miss,
+    /// Nothing was queried at all — offline mode, or the provider list was
+    /// empty once the Musixmatch opt-in filter ran. Never a negative.
+    Unavailable,
+}
+
 async fn external_lyrics_search(
     meta: &TrackMeta,
     providers: Vec<Provider>,
     mode: SearchMode,
     enhanced: bool,
     lang: Option<&str>,
-) -> AppResult<Option<ExternalLyricsResult>> {
+) -> AppResult<SearchOutcome> {
     // Defense in depth: every outbound HTTP path must honour offline mode
-    // regardless of caller. Returning Ok(None) short-circuits before any
-    // request is issued (see the process-wide offline contract).
+    // regardless of caller. Short-circuit before any request is issued
+    // (see the process-wide offline contract).
     if crate::offline::is_offline() {
-        return Ok(None);
+        return Ok(SearchOutcome::Unavailable);
     }
     // Apply the Musixmatch opt-in here — at a single chokepoint — so
     // call sites don't have to know whether Musixmatch is in their
@@ -386,7 +403,7 @@ async fn external_lyrics_search(
     // to query": short-circuit instead of building a no-op client.
     let providers = filter_providers(providers);
     if providers.is_empty() {
-        return Ok(None);
+        return Ok(SearchOutcome::Unavailable);
     }
     // The syncedlyrics client treats `lang = Some(_)` as a hard
     // filter for Musixmatch-only — passing it to a fallback chain
@@ -401,9 +418,9 @@ async fn external_lyrics_search(
     };
     let query = external_query(&meta.title, meta.artist_name.as_deref());
     let client = shared_external_client()?;
-    // A transient provider error surfaces as Err here (not Ok(None)) so
-    // callers don't cache an empty "miss" on a network blip.
-    client
+    // A transient provider error surfaces as Err here (never as a miss) so
+    // callers don't cache an empty row on a network blip.
+    let found = client
         .search(SearchOptions {
             query,
             mode,
@@ -414,7 +431,12 @@ async fn external_lyrics_search(
             netease_cookie: std::env::var("SYNCEDLYRICS_NETEASE_COOKIE").ok(),
         })
         .await
-        .map_err(|err| AppError::Other(format!("external lyrics search failed: {err}")))
+        .map_err(|err| AppError::Other(format!("external lyrics search failed: {err}")))?;
+    // We got here only by actually querying, so `None` is a real negative.
+    Ok(match found {
+        Some(result) => SearchOutcome::Found(result),
+        None => SearchOutcome::Miss,
+    })
 }
 
 /// `profile_setting` key for the user-picked Musixmatch translation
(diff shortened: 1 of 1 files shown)


InstaZDLL:
@coderabbitai Applied — you were right on both counts, and the second one was a real hole.

## The re-check was a guess, not a fact

The previous revision reconstructed Miss-vs-Unavailable in the caller by re-reading `is_offline()` *after* `external_lyrics_search` returned. That's state that can change between the call and the check: if offline was on **during** the call (so `Ok(None)` meant *"couldn't ask"*) and flipped off before the re-check, the result was recorded as a confirmed `Miss` and cached — reintroducing the exact bug #391 is about, just in a narrower window. My own comment justified the re-check by saying the flag "can flip mid-waterfall", which is precisely why the re-check couldn't be trusted either.

## The empty-provider case was unhandled

`filter_providers` can empty the list when the Musixmatch opt-in is off, and that `Ok(None)` meant *"nothing to query"* as well. Unreachable from the fallback chain **today** — NetEase/Megalobiz/Genius all survive the filter — but it would have started caching misses the moment Musixmatch was added to `external_fallback_providers()`. A latent trap rather than a live bug, but worth closing while the code is open.

## What changed

`external_lyrics_search` now returns `SearchOutcome { Found, Miss, Unavailable }` and makes the call at the source, where it alone knows whether a request was actually issued. Callers propagate it; nothing re-derives it.

Auditing the other six call sites turned up the same conflation elsewhere, so they got the distinction too:

- **`refetch_lyrics`** with a pinned provider no longer caches an empty miss *attributed to a provider that was never queried* — e.g. the user pins Musixmatch without the opt-in, or is offline. That badge was actively misleading.
- **`run_prefetch`** counts `Unavailable` as `failed` instead of writing a negative for a track it never asked about.
- **`fetch_radio_lyrics`** leaves the `radio_lyrics` cache untouched, so the next spin of the song on the station's rotation retries.

Checks: `cargo check -p waveflow --all-targets` ✅, `cargo clippy -p waveflow --all-targets` ✅ (no warnings).

InstaZDLL:
@coderabbitai Skipping this one. The single-writer rule you're invoking is real and worth guarding, but batching into 200-row transactions is the wrong shape for `run_prefetch` specifically — it would make the contention worse, not better.

**The write cadence is network-paced, not CPU-paced.** `LRCLIB_THROTTLE` is 500 ms, and there's a `sleep(LRCLIB_THROTTLE)` at the end of every loop iteration plus another between the Musixmatch and LRCLIB calls. That's roughly **two single-row writes per second**.

So a 200-track batch means holding an open write transaction for **100+ seconds**. In WAL there's one writer at a time, so that would block the scanner for that entire window — precisely the failure mode the rule exists to prevent. Note the analyzer that CLAUDE.md cites as the model batches **16**, not 200, because its work is CPU-paced and its transactions are milliseconds long. Copying the batch size without the cadence would invert the intent.

**Batching would also cost crash-resilience.** `run_prefetch` is explicitly idempotent — its own doc comment says so, and the query filters `WHERE l.file_hash IS NULL`. Cancel it or crash mid-run and nothing is lost; the next run resumes on exactly the tracks still missing. With 200-row transactions, a cancel would discard up to 199 tracks' worth of already-completed network work.

On the secondary points: a write that loses to `SQLITE_BUSY` here just leaves that track uncached, and the next run picks it up — benign by construction, given the idempotence above. Parking behind scans is defensible in principle, but prefetch is user-triggered rather than a background daemon, holds no long transaction, and writes one small row every 500 ms; the contention it can generate is negligible.

Scope-wise this is also pre-existing: the write pattern predates this PR, which only changed **outcome classification**. The `Unavailable` arm it adds writes *nothing*, so the change strictly reduces the number of writes on this path.

If you think the park-behind-scan guard is worth having on its own merits, I'm happy to open a separate issue for it — but I don't want to land a batching change here that trades a negligible problem for a hundred-second write lock.

InstaZDLL:
@coderabbitai review

PR #398, opened 2026-07-18
fix(profile): lease the profile pool so a switch cannot close it mid-command

### LINKED ISSUE #332: Suivi: ajouter un mécanisme de lease/refcount pour éviter `PoolClosed` lors des bascules de profil
## Contexte

Lors d’un changement de profil, `src-tauri/crates/app/src/state.rs` ferme immédiatement l’ancienne `SqlitePool` dans `activate_profile`. Toute commande déjà en cours qui a cloné cette pool via `require_profile_pool()` ou `require_profile_snapshot()` peut alors échouer avec `PoolClosed`.

La PR #329 a corrigé le bug de mélange de données en capturant atomiquement `(pool, profile_id)`, mais elle n’a pas vocation à résoudre ce comportement global de cycle de vie des profils.

## Problème

- Le risque `PoolClosed` n’est pas limité à l’export des stats : il peut toucher n’importe quelle commande multi-étapes utilisant une pool clonée.
- L’échec est gracieux (erreur de commande), mais il reste possible pendant une bascule de profil concurrente.

## Changement attendu

Introduire un mécanisme de **lease / refcount de profil** :

- l’acquisition d’un snapshot/pool actif crée un garde (lease) ;
- `activate_profile` / `deactivate_profile` ne ferment pas la pool tant qu’il reste des leases actives ;
- la fermeture intervient seulement après libération de toutes les leases.

## Zones concernées

- `src-tauri/crates/app/src/state.rs`
- Appels à `require_profile_pool()`
- Appels à `require_profile_snapshot()`
- Commandes multi-étapes, notamment `src-tauri/crates/app/src/commands/stats.rs`

## Critères d’acceptation

- Un changement de profil concurrent ne provoque plus `PoolClosed` pour une commande ayant déjà acquis un snapshot/lease valide.
- `activate_profile` et `deactivate_profile` attendent correctement la libération des leases avant fermeture effective de la pool précédente.
- Le comportement reste sûr en cas d’absence de profil actif.
- Des tests couvrent au minimum le scénario de bascule concurrente pendant une commande multi-étapes.

## Backlinks

- PR : https://github.com/InstaZDLL/WaveFlow/pull/329
- Commentaire de demande : https://github.com/InstaZDLL/WaveFlow/pull/329#issuecomment-4827157307
- Demandé par : @InstaZDLL

> Note: l’assignation automatique à @InstaZDLL a échoué côté intégration GitHub, donc l’issue est créée sans assignee.

InstaZDLL:
Closes #332.

## Problem

`activate_profile` closed the previous `SqlitePool` as soon as the new one was swapped in. Any command that had already cloned that pool via `require_profile_pool()` / `require_profile_snapshot()` could then fail with `PoolClosed` on its next query.

PR #329 fixed the *data-mixing* half of this (atomic `(pool, profile_id)` capture) but explicitly left the lifecycle question open — the window is reachable by any multi-step command, not just the stats export that surfaced it.

## Approach

The pool is now handed out **leased**. `require_profile_pool` / `require_profile_snapshot` return a `ProfilePool` holding a refcount on the epoch it came from, and `ActiveProfile::close_when_idle` drains outstanding leases before calling `pool.close()`.

What makes the drain terminate: the swap still happens **first**, under the write lock. Once the outgoing `ActiveProfile` is out of `AppState::profile` no further lease can be issued against it, so the count is monotonically decreasing.

The wait is **bounded** by `LEASE_DRAIN_TIMEOUT` (5 s). A library scan legitimately holds its pool for minutes, and a leaked lease would otherwise wedge profile switching outright — so hitting the bound degrades to the pre-#332 behaviour (close anyway, race whatever remains) and logs at WARN rather than blocking forever. This is a deliberate trade, not an oversight.

`wait_idle` uses `Notified::enable()` before reading the count: tokio registers a waiter on first *poll*, not on future creation, so without it a lease dropped between the load and the await would notify nobody and the drain would sit out its full timeout.

## Call-site fallout

Mechanical, and the compiler proved completeness — 137 sites, all flagged.

- **126 query sites take `&*pool`.** `ProfilePool` derefs to `SqlitePool` and passes anywhere a concrete `&SqlitePool` is expected, but sqlx's query methods are generic over `E: Executor` and deref coercion does not fire against a type variable. The rewrite was driven by the compiler's exact JSON spans rather than a blind sed.
- **Two repository helpers** (`library_repo` / `playlist_repo`) hand an owned pool to `waveflow-core` types that know nothing about leases. They now split with `into_parts()` and park the lease next to the value via the new `state::Leased<T>` wrapper, which derefs to the repository so every call site is unchanged.
- **The DLNA worker opts out** via `into_unleashed()`. It holds its pool across requests for the whole life of the server and is rebuilt on profile switch by design; a lease there would stall *every* switch until the drain timeout. That handle is back to the pre-#332 exposure and the comment says so.

## Tests

`mod lease_tests` in `state.rs`, 7 cases:

- the acceptance scenario — a leased in-memory pool queried on **both sides** of a concurrent `close_when_idle`, asserting the second query does not hit `PoolClosed` and that the switch has not completed while the lease is held;
- the pool really *is* closed once the lease drops (no leaked epoch);
- drain semantics: immediate when idle, blocked while held, **all** leases required, parked drain woken by the release (asserts it resolves promptly rather than timing out), per-epoch isolation.

## Verification

- `cargo check --workspace --all-targets` ✅
- `cargo clippy --workspace --all-targets` ✅ (clean)
- `bun run typecheck` ✅

⚠️ `cargo test -p waveflow --lib` does **not** run on my Windows box — it aborts with `STATUS_ENTRYPOINT_NOT_FOUND` before reaching any test, a pre-existing Tauri DLL issue unrelated to this change. I extracted the mechanism into a standalone crate with the same tokio + sqlx versions and ran the identical 7 tests there: **7 passed**. The regression test fails as expected against the old immediate-close. CI runs them for real.

## Docs

`CLAUDE.md` cross-cutting rule + a new "Pool lifecycle across a profile switch" section in `docs/architecture/storage.md`. The `&*pool` requirement is the kind of thing the next contributor will trip over otherwise.

## Summary by CodeRabbit

* **Améliorations**
  * Renforcement de la fiabilité lors du changement de profil : les opérations en cours peuvent désormais se terminer avant la fermeture de la base de données.
  * Réduction des erreurs temporaires susceptibles d’interrompre la lecture, la lecture de la bibliothèque ou les réglages audio pendant ces transitions.
  * Les services de longue durée, comme la diffusion DLNA, continuent de fonctionner conformément à leur cycle de vie.

* **Documentation**
  * Documentation enrichie sur le cycle de vie de la base de données, les changements de profil et les comportements en cas de délai d’attente.


InstaZDLL:
@coderabbitai review

InstaZDLL:
@coderabbitai Thanks — all three verified against the code, two fixed here and one split out. Pushed in f71b7ca.

**1. Lease doc overclaimed — fixed.** You're right, and it was my wording. `close_when_idle` gives up after `LEASE_DRAIN_TIMEOUT` and closes anyway, so a command holding a lease past 5 s across a switch can still hit `PoolClosed`. The docs now say the guarantee is time-bounded, not absolute, and that long commands (a full library scan) must keep tolerating the error — what the lease buys is that *ordinary* multi-step commands no longer race the close at all. Qualified in `require_profile_pool`, `CLAUDE.md` and `docs/architecture/storage.md`. No runtime change.

**2. Deezer batch — fixed.** Confirmed: `fetch_all_artists_deezer` read its artist list from one pool, then looped over the `enrich_artist_deezer` command wrapper, which re-resolves the active profile on every iteration. A `switch_profile` mid-batch scattered the remaining writes into the other profile. Extracted `enrich_artist_deezer_with_pool` and pointed the batch at the pool it read the list from. Notably the function one level down already had a comment applying exactly this reasoning ("Pass the SAME pool into the inner…") — the batch layer just never got the same treatment.

Worth noting this one is pre-existing rather than introduced here, but it's the same defect class as #332 and the fix is contained, so it belongs in this PR.

**3. DLNA — you're right about the bug; I've split the fix into #399.**

The important part: **my comment was factually wrong.** I wrote that `build_resources` "is re-called on switch, which is the intended lifecycle" and used that to justify the `into_unleashed()` exemption. It isn't — `build_resources` is only called from the boot path and `dlna_set_config`, and `switch_profile` never touches DLNA. I asserted a lifecycle I hadn't verified, and the docs I wrote in `CLAUDE.md`/`storage.md` repeated it. All three now describe the real behaviour: a running server keeps serving the previous profile and then fails against its closed pool.

I did **not** implement the rebuild here, deliberately:

- it predates this PR (the old code closed the pool immediately too, so the failure mode is unchanged — this PR neither causes nor worsens it);
- it's a behavioural change to a worker thread I can't exercise locally, and verifying it needs a real UPnP client to confirm the SSDP announcer and axum bind are torn down cleanly and that clients don't see a new device;
- bundling an untested worker-lifecycle change into a pool-lifecycle PR makes both harder to review and to revert.

#399 has the full analysis and the expected behaviour, including the `deactivate_profile` case you'd otherwise leave serving a closed pool.

`cargo clippy --workspace --all-targets` clean.

Issue #366, opened 2026-07-12
Library: external metadata edits not picked up on rescan, and re-adding a deleted album isn't detected


InstaZDLL:
Reported by a beta tester (email). Two related library-scan problems that together force a full library wipe as the only workaround (which then triggers the stats-loss bug — see linked issue).

## Symptom A — edited metadata/artwork not detected on rescan

> I add "Aladdin Sane" (Bowie). Later I change the picture (I don't like it) or the genre. I'd expect a library sync to update it. The problem is the library doesn't recognize the change as a change, so it keeps the old picture.

The scanner's fast path keys on `(mtime, size)` (+ `file_hash`), so a change that doesn't move those is skipped. Likely cases:
- A **folder cover** (`cover.jpg` / `folder.jpg`) is changed but the audio files' mtime/size are unchanged → the album art isn't refreshed.
- A tag edit done by a tool that **preserves mtime**, or that only touches an embedded picture, may not flip the fast-path key.

## Symptom B — re-adding a deleted album isn't seen

> If I delete "Aladdin Sane" from the library, sync, then add the files again and sync, the library doesn't see "Aladdin Sane" anymore despite it being in the folder. The only fix is to delete the whole library from the app and rescan from scratch.

This is the more serious one — a re-added file with **identical content (same blake3)** isn't resurfaced. Suspect the `file_hash` dedup / the `is_available` handling leaving a stale row that blocks re-insertion, or `removed` rows not being reconsidered on the next scan.

## Log evidence

Normal rescans DO detect some deltas (`scan complete scanned=632 added=38 skipped=594 removed=29`), so change detection partially works — the gap is specifically (A) metadata/artwork-only edits and (B) re-add after delete.

## Investigation starting points

- [`commands/scan.rs`](src-tauri/crates/app/src/commands/scan.rs) — the `(mtime, size)` skip-fast-path + `file_hash` handling + the `removed` / `is_available` reconciliation.
- Decide the intended contract: should an artwork/genre-only edit force a re-read? Should a re-added identical file resurrect a previously-removed row?

## Environment

Windows, `1.6.x`. Reproducible per the steps above.

Related: #<STATS_ISSUE> (the wipe workaround loses stats).

InstaZDLL:
Related: #367 — deleting the library to work around this bug is what loses the tester's stats.

InstaZDLL:
**Symptom B (re-add after delete) → fixed in #385.**

[...]

PR #400, opened 2026-07-18
fix(scan): pick up sidecar cover art replaced between scans

InstaZDLL:
Refs #366 — symptom A. (Symptom B, the byte-identical re-add, was fixed in #385.)

## The bug in one line

The scanner watches the **audio** files' `(mtime, size)`. The file that changed is `cover.jpg`. It never looks at it.

Concretely: you replace the picture next to *Aladdin Sane*, rescan, and nothing happens — the `.flac` files are untouched, so the fast path skips them, `extract_folder_cover` never runs, and the album keeps its old artwork indefinitely. The only way out today is deleting the whole library and rescanning, which loses play stats.

There is a second lock behind the first: even on the slow path, the skip branch only backfills artwork when the album has **none** (`if let Some((Some(aid), None))`), so replacing an existing cover was ignored there too.

## Fix

`refresh_folder_covers`, a post-scan pass that works per **directory** rather than per file — one `read_dir` + one hash for a whole album instead of one per track. It re-resolves the sidecar and updates any album whose stored artwork no longer matches.

**No migration.** `artwork.hash` is already the blake3 of the picture bytes, so the stored row is its own baseline for comparison — there is no new state to persist and nothing to keep in sync.

**No new UI.** This was the explicit design constraint: the fix is automatic and invisible. (A "deep rescan" button was the alternative; it was dropped because it only earns its keep for a failure mode we have no evidence of — see below.)

Three deliberate restrictions:

- **Embedded artwork is never overridden.** Extraction treats the sidecar as a *fallback* for tracks whose tag carries no picture. Replacing an `embedded` row here would silently invert that precedence, so only `folder`-sourced (or absent) artwork is touched.
- **A deleted sidecar does not blank the album.** A vanished cover is far more likely to be a transient state — files being reorganised — than a request for a blank album.
- **A multi-directory album resolves against its first directory** in sorted order. Otherwise per-disc subfolder covers would make the album's picture depend on directory iteration order, i.e. flip on every scan.

Placed in `waveflow-core` next to `merge_implicit_compilations`, which is the same shape (a post-scan pass over the pool), per the crate split rule in `CLAUDE.md`.

## On the "genre" half of the report

The issue also mentions genre edits not being detected, but that part was a **hypothesis in the issue text, never confirmed**. A genre change rewrites the audio file, which moves its mtime, which the fast path does detect. It would only fail with a tagger that restores the original mtime — uncommon; Mp3Tag and Picard don't. Worth confirming with the reporter before building anything for it, rather than adding machinery for a failure mode we haven't observed.

## Tests

Six cases in `waveflow-core`, using a real in-memory SQLite plus temp directories: the regression itself (sidecar swapped, audio untouched), unchanged sidecar is a no-op (otherwise every scan would churn `album.artwork_id` library-wide), embedded artwork wins, an album with no artwork adopts the sidecar, a deleted sidecar leaves the album alone, and rows under a different `folder_id` aren't dragged in.

These run locally: **6 passed.** Unlike the app crate, `waveflow-core` tests execute fine on Windows.

## CI gap found while writing them

`sqlite` is not a default feature of `waveflow-core`, so the module tree behind it — and its tests — only compiles when something in the build graph turns it on. The Linux `cargo test --workspace` gets that via feature unification with the app crate, but the **Windows job excludes exactly that crate**, so these tests would have silently not existed there. I hit this myself: my first run reported "89 filtered out" and none of the new tests, because the feature was off.

Added an explicit `cargo test -p waveflow-core --features sqlite` step that runs on both runners, rather than leaving correctness dependent on feature unification.

## Verification

- `cargo clippy -p waveflow-core --all-targets --features sqlite` ✅ (the config the workspace run doesn't cover)
- `cargo test -p waveflow-core --features sqlite` ✅ 6/6

Docs: `CLAUDE.md` library catalogue + a new "Folder-cover reconciliation" section in `docs/features/library.md`.

## Summary by CodeRabbit

  * Lors d’un scan de dossier, les pochettes “sidecar” (ex. `cover.jpg`) sont maintenant réconciliées et mises à jour même si les fichiers audio ne sont pas rescannés.
  * Les pochettes intégrées, personnalisées et issues de sources externes sont protégées contre le remplacement automatique.
  * Le scan affiche désormais le nombre de pochettes effectivement actualisées.
* **Documentation**
  * Mise à jour de la documentation avec les règles de réconciliation et de priorités.
* **Tests**
  * Les tests liés à SQLite sont exécutés automatiquement sur Windows dans le pipeline CI.


InstaZDLL:
@coderabbitai All four verified against the code and **all four fixed** — none were stale. Pushed in c35e5d2.

**1. Source guard was a denylist — the worst of the four, and my bug.** I tested only for `"embedded"`, but `upsert_artwork`'s own docstring lists `embedded` / `folder` / `deezer` / `user`. So a cover the user **uploaded by hand**, or one fetched from Deezer, was replaced by whatever sidecar happened to sit next to the files. That destroys a deliberate choice, silently, on the next scan.

Inverted to an allowlist — `source = 'folder'`, or no artwork at all. Beyond fixing the two known sources, this makes the failure mode safe by default: a source added later is preserved rather than clobbered by a pass that predates it. Two regression tests (`user`, `deezer`).

**2. Directory selection scoped to the scanned folder — valid, and it broke my own stated invariant.** The query filtered `t.folder_id = ?`, so a two-disc album split across two library folders saw only the directories under whichever folder triggered the scan: disc 1 wins when folder 1 is scanned, disc 2 when folder 2 is, flipping on alternating passes. That is exactly the non-determinism the "first directory in sorted order" rule exists to prevent — the rule was right, the query didn't implement it.

Now a subquery scopes to *albums touched by* the folder, and the outer query takes every available track of those albums. `a_multi_folder_album_resolves_consistently` asserts the same directory wins from both directions. Also renamed `only_the_scanned_folder_is_reconciled` → `an_album_outside_the_scanned_folder_is_untouched`, since its old name now claims something deliberately untrue.

**3. Blocking FS on the async runtime — valid, fixed.** Worth noting this is the module's *existing* pattern (`maybe_link_artist_images` and `link_va_artist_image` both do `read_dir` inline), so I did consider skipping it for consistency. I fixed it because the profile here is different: those are memoised per `(artist, folder)` and mostly no-op, whereas this pass does a `read_dir` + full read + blake3 for **every** distinct directory, once, at the tail of the scan — order of seconds on a large library, blocking IPC replies and the progress emitter throughout.

Resolution now happens in a single `spawn_blocking` batch (not one per directory). `tokio` joins `waveflow-core` as an **optional** dependency gated on the `sqlite` feature, so the postgres-only server build — which skips these modules entirely — doesn't pick it up. Verified with `cargo check -p waveflow-core --no-default-features --features postgres`.

**4. Per-album autocommit — valid, fixed.** Was `pool.acquire()` + two autocommitted statements per album. Now one transaction for the batch: the scanner is the single writer, so N round-trips through WAL bought nothing, and the artwork insert now lands together with the album row that points at it.

## Verification

- `cargo test -p waveflow-core --features sqlite` ✅ **9/9** (6 existing + 3 new)
- `cargo clippy -p waveflow-core --all-targets --features sqlite` ✅
- `cargo check -p waveflow-core --no-default-features --features postgres` ✅ (confirms the tokio gate)

Docs updated in `CLAUDE.md` and `docs/features/library.md` to describe the allowlist and the album-wide directory resolution, rather than the narrower behaviour I originally wrote.

InstaZDLL:
@coderabbitai Both verified. **One fixed here, one split out** — with reasoning, since the second is a bigger change than it looks.

## 1. Re-check the allowlist at write time — valid, fixed

Confirmed the window is real: `candidates` is built from a read taken *before* the `spawn_blocking` directory walk, which on a large library runs for seconds. The user can upload a cover for one of those albums in that gap, and the pass would then overwrite it — the exact failure the up-front filter exists to prevent, just moved later in time.

The UPDATE now carries the allowlist as part of the write, so the check and the update are one statement and nothing can slip between them. Success comes from `rows_affected()` rather than from intent, so an album that left the allowlist neither gets clobbered nor inflates the summary.

Extracted as `link_folder_cover_if_eligible` rather than inlined, specifically so the guard is testable: this race can't be staged deterministically through the full pass. Three cases — replaces a sidecar cover, adopts one when the album has none, refuses `user` / `deezer` / `embedded`.

## 2. `upsert_artwork` infers provenance from a hash-keyed row — valid, but not fixable here → #401

You're right, and it's worse than the immediate symptom: `artwork.hash` is `UNIQUE` and `source` is only written on **insert**, so the column records *who inserted the bytes first*, not where a given album got its cover. It's stored as if it were the latter.

Worked example: a cover both embedded in album A's tags and shipped as `cover.jpg` beside album B, same bytes → same hash. A is scanned first, the row gets `source = 'embedded'`. B is pointed at that same row. The user later replaces B's `cover.jpg` — my guard reads `embedded`, treats B as off-limits, and never refreshes it again. B's cover is silently frozen.

**Why not in this PR.** Both of your suggested shapes — provenance on the association, or `source` in the uniqueness key — require a **new dated migration** plus a backfill decision for existing rows, and they change `upsert_artwork`, which the whole scanner routes through (embedded covers and artist images, not just this pass). That is a schema change to shared infrastructure, and folding it into a scan-behaviour PR would make both harder to review and to revert. It is also not a regression: before this PR no cover was refreshed at all, so the worst case is that the new feature quietly doesn't apply to those albums — nothing is lost or corrupted.

Filed as **#401** with the worked example, both options, and acceptance criteria including your "same hash, two provenances" test. I've also added the caveat to the doc comment right at the guard, so the next reader doesn't trust `source` more than it deserves.

## Verification

- `cargo test -p waveflow-core --features sqlite` ✅ **12/12** (9 + 3 new)
- `cargo clippy -p waveflow-core --all-targets --features sqlite` ✅ — this caught a real slip: my first edit landed the new helper *inside* `refresh_folder_covers`'s doc block, splitting it in two. Fixed before pushing.
- `cargo check -p waveflow-core --no-default-features --features postgres` ✅

InstaZDLL:
@coderabbitai Applied in 9d029b8 — though with a correction to the premise, so the reasoning is on record.

**This was not a live defect.** The suite passes, 12/12 in ~0.03 s, which rules out the failure mode the finding implies. If the pool read had landed on a *different* in-memory database, `album` wouldn't exist and `album_artwork_hash`'s `unwrap` would have panicked; if the pool were capped at one connection, the second acquire would have blocked until the 30 s acquire timeout. Neither happens, because sqlx shares a single in-memory database across a pool's connections.

**But the suggestion is still worth taking**, for a different reason than stated: my assertions were quietly depending on that sqlx behaviour without saying so. If it ever changes, the tests break in a way that looks like a bug in the guard rather than in the harness. `drop(conn)` removes the dependency for one line.

Applied to **three** tests, not two — `the_write_guard_refuses_a_cover_that_became_user_owned` acquires a connection inside its loop and had the same shape. Assertions unchanged, as asked. Added a line to the doc comment explaining why the drop is there, so nobody later removes it as redundant.

`cargo test -p waveflow-core --features sqlite` ✅ 12/12 · `cargo clippy -p waveflow-core --all-targets --features sqlite` ✅

InstaZDLL:
@coderabbitai Valid, fixed in 57a732a.

You're right that the guard was half a check. It asserted the album's artwork was still *sidecar-sourced*, but not that it was still the *same* artwork — and two folder-sourced covers both satisfy that, so the source allowlist can't distinguish them. A fresher sidecar resolved by a concurrent scan of the same album could be rolled back to whatever this pass computed from a staler read, with nothing in the summary to say so.

`observed_artwork_id` now rides from the candidate query into the update as a compare-and-swap, **alongside** the source allowlist rather than replacing it — the two cover different hazards (a concurrent scan vs. a user upload landing mid-walk), and I'd rather each be stated explicitly than argue one subsumes the other.

One detail worth flagging: the CAS binds with `IS`, not `=`. The observed value is legitimately `NULL` for an album that had no artwork, and SQLite's `=` never matches `NULL`, so `=` would have silently broken the "album adopts its first cover" path. That test now passes `None` explicitly and asserts the NULL case.

The candidate tuple reached four positional fields on the way, which was becoming unreadable, so it's a named `CoverCandidate` struct.

## On severity — stated plainly

This is **narrow**: it needs two scans touching the same album *and* the sidecar changing between their directory walks. The outcome is also self-correcting, since the pass compares hashes and the next scan fixes it. I fixed it anyway because a read-compute-write cycle should be a compare-and-swap on principle and here it costs one bind — not because I think it was biting users.

## Test

`the_write_guard_refuses_to_roll_back_a_newer_folder_cover` — three folder-sourced artworks: the one we observed, the one we resolved, and the one a concurrent scan landed. Asserts the stale write neither succeeds nor reports success. I checked it's discriminating: without the CAS the update matches (artwork 3 is `folder`-sourced), `changed` is true, and the assertion fails.

- `cargo test -p waveflow-core --features sqlite` ✅ **13/13**
- `cargo clippy -p waveflow-core --all-targets --features sqlite` ✅
- `cargo check -p waveflow-core --no-default-features --features postgres` ✅

`docs/features/library.md` now notes that the candidate list is a snapshot re-validated at write time.

InstaZDLL:
@coderabbitai review

InstaZDLL:
@coderabbitai **Skipping this one** — I checked it against the codebase and the suggestion inverts the project's actual design. No code change.

## The trait layer isn't the convention here

`repository/` is a per-**entity CRUD** abstraction — `library`, `playlist`, `profile`, `track` — with parallel `sqlite/` and `postgres/` implementations under one trait surface. Its own module doc says so: *"Every trait here describes the persistence surface a single domain entity exposes… without committing to a backend."*

`scanner/upserts` is deliberately not that. The module header states the rule:

> Every helper takes `&mut sqlx::SqliteConnection` so it can participate in the open transaction `scan_folder_inner` runs across a batch — never reach for the pool from inside these to avoid breaking the single-writer discipline.

The concrete connection type **is the contract**. It's what lets the scanner batch a whole commit window across `upsert_artwork` / `upsert_album` / `upsert_artist` / `merge_implicit_compilations`. All ten existing functions in the file take `&mut SqliteConnection` or `&SqlitePool`; `refresh_folder_covers` and `link_folder_cover_if_eligible` follow that exactly. Abstracting only my two would make them the odd ones out in the module they live in.

## The postgres path is already planned — and it isn't a shared trait

`scanner/mod.rs` gates the module and says what the second backend will look like:

```rust
// The future `waveflow-server` ingest job will provide a parallel
// `upserts_pg` module behind `feature = "postgres"` once the schema
// settles — for now the postgres build skips this entirely.
#[cfg(feature = "sqlite")]
pub mod upserts;
```

A **parallel module**, not a shared trait. That's the project's stated decision, and the whole `upserts` module is `#[cfg(feature = "sqlite")]` — the postgres build never compiles a line of it. So the abstraction would buy nothing today: there is no second implementation to satisfy, and the server doesn't scan local filesystems for sidecar covers in the first place.

## What it would cost

Introducing a trait for these two functions inside a bug-fix PR means touching shared scanner infrastructure, diverging from the file's documented transaction discipline, and contradicting a design note in `scanner/mod.rs` — while making the CAS harder to read, since the compare-and-swap and the `rows_affected` semantics are precisely the sqlx-specific behaviour the trait would have to leak anyway.

If the `upserts_pg` module ever lands and the two backends turn out to share a real surface, unifying them then is the right moment — with both implementations in hand rather than guessing the shape from one. Happy to open a tracking issue if you think that's worth queuing, but I don't think it belongs in #366.

PR #402, opened 2026-07-19
fix(stats): keep listening history when its tracks are deleted

### LINKED ISSUE #367: Stats/listening history are lost when the library is deleted, and a backup does not restore them
Reported by a beta tester (email) — recurring data loss, hit **5 times**.

## Problem

> If you delete the library, the app forgets your stats. Even if you did a backup, it doesn't read them back. I've lost my stats 5 times. I'd like a way to save them that doesn't depend on the folder — maybe something keyed only on the track name.

Two distinct failures:

1. **Deleting/resetting the library wipes `play_event`** (the listening history that drives Stats + Wrapped + On Repeat). The tester is forced into this by the library-scan bug (see linked issue), so the two compound.
2. **Backup/restore doesn't bring the stats back.** Either the archive doesn't include the play history, or restore doesn't re-import it — needs verifying.

## Log evidence

```
10:41:47  WARN stop_and_wait failed during reset; proceeding with wipe anyway ...
10:41:49  INFO created default profile profile_id=1
```
A "reset" recreates the default profile from scratch — wiping the per-profile DB where `play_event` lives.

## Design consideration (from the reporter)

Play history is keyed on `track_id`, which is per-library and doesn't survive a wipe/re-scan (a re-added file gets a new id). The tester suggests persisting stats **independent of the folder/library** — e.g. keyed on a stable natural key (artist+title, or file_hash) so a rescan can re-attach history to the same songs.

Options to weigh:
- Make the backup archive **actually include** `play_event` and have restore re-import it (verify current behavior first).
- Add an **export/import listening-history** action (JSON/CSV) independent of the library, re-attached by a stable key (blake3 `file_hash` preferred over artist+title to avoid collisions) on the next scan.
- Guard the destructive "reset library" / "delete library" paths so they don't silently drop history without an explicit, warned confirmation.

## Impact

High — irreversible loss of user data (their entire listening history) with no recovery path today.

## Environment

Windows, `1.6.x`.

Related: #366 (the library bug that pushes users into the wipe).

InstaZDLL:
Closes #367.

## What was actually wrong

The issue listed three hypotheses and asked for them to be verified. One is wrong, and the real cause is elsewhere.

**The backup does contain the listening history.** The archive bundles the entire `data.db`, `play_event` included, and `import_profile` restores it. So "either the archive doesn't include the play history, or restore doesn't re-import it" can be struck out.

The trap is subtler, and worse: restoring gives you back the **old library**, not your history *plus* a fresh scan. A user whose library is broken has to choose between their files and their stats. The reporter's log line (`created default profile profile_id=1`) also suggests they used the factory reset rather than an import.

**The real loss is structural:**

```sql
track_id INTEGER NOT NULL REFERENCES track(id) ON DELETE CASCADE
```

Removing a folder runs `DELETE FROM track WHERE folder_id = ?`, the cascade fires, and the history is gone. Silently, with no confirmation of consequence — the delete button is armed by one click and committed by a second, saying nothing about what it takes with it.

## Approach

`track_id` becomes nullable with `ON DELETE SET NULL`: deleting a track now **orphans** its history instead of destroying it. Each event carries a snapshot of how to find its track again — `snapshot_hash`, `snapshot_path`, `snapshot_artist`, `snapshot_title` — written at **insert** time, which is the only moment that information is guaranteed to exist. By the time a folder is deleted, the row we'd have read is already gone.

`reattach_orphaned_play_events` runs after every scan, strongest key first:

1. **`file_hash`** — same bytes, moved or re-added. Exact, but a tag edit rewrites the file through lofty, so the blake3 changes even though the music didn't.
2. **`file_path`** — catches precisely that case: same file, different hash. Fails if folders were reorganised.
3. **artist + title** — a re-rip or different encoding. Loosest, deliberately last: it can't tell a live version from the studio one.

Each step only claims what the previous left, so a strong match is never overwritten by a weak one, and only `is_available = 1` tracks match — attaching to a vanished file would just re-orphan on the next pass.

## A consequence worth stating

Stats now split coherently, and it's worth knowing before writing a new query:

- **Aggregate totals** (play count, listening time, monthly histogram) read `play_event` directly, so they **survive a library delete**. That's the whole point of the issue.
- **Per-item breakdowns** (top tracks, top artists) join `track` and therefore exclude orphans — a row you can't name can't be rendered. They come back when the files are re-scanned.

No query changes were needed for this; it falls out of the join structure. Documented rather than "fixed", because it's the behaviour you want.

## Delete confirmation

The folder delete had a two-step arm/commit but gave no indication of what was at stake. It now shows how much history the folder carries. Since that history survives, the copy **informs rather than threatens** — it says the plays will be kept and re-attached on the next scan.

`count_folder_play_events` is fetched on arm, not up-front, so listing folders stays one query. The result is keyed by folder id, so a late response can't be shown against a different folder.

## Tests

**8 cases** in `waveflow-core`, on a real in-memory SQLite with `PRAGMA foreign_keys = ON` (SQLite defaults them **off** per connection — without it the orphaning under test never fires):

- the regression itself: deleting a track orphans its history instead of erasing it
- re-attach by hash, by path (hash changed), by artist+title
- **priority** — the hash match wins even when a decoy with a lower row id matches a weaker key
- an unmatchable orphan is left alone, not dropped and not attached to something arbitrary
- an unavailable track is not a match
- an already-attached event is never re-pointed

**The migration was run for real**, against a pre-migration schema with data: all four snapshot columns backfill from the track, deleting the track orphans instead of cascading, and the three indexes rebuild. A broken migration panics every existing install at boot, so this wasn't something to eyeball.

## Verification

- `cargo test -p waveflow-core --features sqlite` ✅ **120/120** (112 + 8 new)
- i18n propagated to all 17 locales, with the CLDR categories each one actually needs — `ru` takes `_few`/`_many`, `ar` takes `_two`/`_few`/`_many`. Omitting them leaks English at runtime.

Docs: `CLAUDE.md` library catalogue + a "Listening history outlives its tracks" section in `docs/architecture/storage.md`.


InstaZDLL:
@coderabbitai Both valid. Fixed in 243ea27 — the second one in the opposite direction to what you suggested, with evidence.

## 1. `record_play_event` → `insert_play_event` — my error

The function is `insert_play_event` ([analytics.rs:339](../blob/main/src-tauri/crates/app/src/audio/analytics.rs#L339)). I invented the name while writing the prose and never checked it against the code. Corrected; the surrounding explanation is untouched.

## 2. Turkish register — harmonised the other way

You're right that there's a mismatch, and it matters: the two strings alternate on the *same button*, so the register visibly flips as it arms.

But I checked which one is the outlier before changing anything, and it's **mine**. `tr.json` is written in the informal register throughout:

```
"Müzik klasörünü seçmek için tıkla"
"Klasörü değiştirmek için tıkla"
"Aşağıdan kendi dilini seç"
"Profil kullanılamıyor — biraz sonra tekrar dene."
"...istediğin zaman Ayarlar → Görünüm bölümünden değiştirebilirsin."
```

Informal imperatives and informal possessives, consistently — including `removeConfirm` itself, the sibling key this one alternates with. My new string was the one that broke the pattern with `tıklayın`.

So I changed `removeConfirmWithPlays_*` to `tıkla` rather than rewriting `removeConfirm`. Updating the existing string would have moved the file *further* from its own convention, and it would mean editing a translation I didn't author, in a language I don't speak, on the strength of nothing but "make these two agree" — when the file itself already says which way agreement should go.

Worth noting the locale isn't perfectly consistent (there's at least one stray `tıklayın` elsewhere), but the dominant register and the immediate sibling both point the same way.

`bun run typecheck` ✅ · `bun run lint` ✅ · `tr.json` re-parsed clean. Diff is 3 lines across 2 files.

InstaZDLL:
@coderabbitai review

PR #404, opened 2026-07-19
fix(artwork): record cover provenance on the album, not the artwork row

### LINKED ISSUE #401: artwork.source is unreliable: rows are deduped on hash alone, so provenance is whoever inserted first
## Problem

[`upsert_artwork`](../blob/main/src-tauri/crates/core/src/scanner/upserts.rs) dedups on the content hash alone:

```sql
SELECT id FROM artwork WHERE hash = ?
```

The `source` column is only written on **insert**, so whichever caller happens to see the image first decides its recorded provenance forever. The column therefore describes "how this picture first entered the library", not "where this album's cover came from" — but it is stored as if it were the latter, and `artwork.hash` is `UNIQUE`, so there is no second row to disambiguate.

## Why it matters now

PR #400 added `refresh_folder_covers`, which decides whether an album's cover may be replaced by reading `artwork.source` and only accepting `folder` (or no artwork). That guard is correct in intent but reads a field that can be wrong.

Concretely, with a rip whose cover is **both** embedded in the tags of one album and shipped as `cover.jpg` next to a different one — the same bytes, so the same hash:

1. The embedded album is scanned first → `artwork` row created with `source = 'embedded'`.
2. The sidecar album is scanned → `upsert_artwork` returns that same row, and the album is pointed at an `embedded`-labelled row even though its cover came from a sidecar.
3. The user later replaces that `cover.jpg`. `refresh_folder_covers` sees `source = 'embedded'`, treats the album as off-limits, and **never refreshes it again.**

The album's cover is silently frozen. Nothing is lost or corrupted, and it is not a regression — before #400 no cover was refreshed at all — but the new feature quietly doesn't work for those albums.

The same reasoning applies to `link_local_artist_image`, which also routes through `upsert_artwork`.

## Options

1. **Move provenance to the association.** Add `album.artwork_source` (and `artist.artwork_source`), leaving `artwork` as a pure content-addressed blob table. Most correct: provenance genuinely is a property of the link, not of the bytes.
2. **Include `source` in the uniqueness key** — `UNIQUE (hash, source)`. Smaller change, and cheap on disk since the file is hash-addressed and still stored once; costs a duplicated row per (image, source) pair.

Either way this needs a **new dated migration** plus a backfill decision for existing rows, which is why it was not folded into #400.

## Acceptance

- An image used as both an embedded cover and a sidecar keeps the two provenances distinct.
- An album whose cover came from a sidecar stays refreshable by `refresh_folder_covers` regardless of what other albums embed the same bytes.
- A test covering one hash used with two sources.

Found during review of #400 (thanks @coderabbitai).

InstaZDLL:
Closes #401.

## The bug

`artwork` rows dedup on the content hash alone, and `source` is only written on INSERT. So the column records **whoever put those bytes in the library first**, not where a given album got its cover.

The image embedded in album A's tags and shipped as `cover.jpg` beside album B is one row, labelled `embedded`. B's cover then looks untouchable to `refresh_folder_covers`, whose allowlist reads that column — so replacing B's `cover.jpg` does nothing, permanently.

## Fix

Provenance is a property of the **link**, not of the bytes, so it moves to `album.artwork_source`. Every site that writes `album.artwork_id` writes it too: the scanner's three paths, the tag editor, Deezer enrichment, manual upload, and the sidecar pass. `artwork.source` stays a rough origin label for the bytes and is no longer consulted to decide what may be overwritten.

`artist` gets no equivalent column — its guard is `artwork_id IS NULL` and never reads a source.

## I reversed my own recommendation, and why

The issue offered two options and I called the additive column "most correct" but the uniqueness widening "smaller". **That was wrong, and dangerously so.**

Widening `artwork` to `UNIQUE (hash, source)` needs the create-copy-drop-rename rebuild. But `artwork` is a **parent** of `album.artwork_id` and `artist.artwork_id`, both `ON DELETE SET NULL`. The profile pool opens connections with `foreign_keys = ON`, and SQLite's `DROP TABLE` performs an implicit `DELETE` that **fires foreign-key actions**.

I verified it against a real database rather than reasoning about it:

```
hors transaction     | foreign_keys=1 | album.artwork_id après rebuild = None  !!! PERDU !!!
dans une transaction | foreign_keys=1 | album.artwork_id après rebuild = None  !!! PERDU !!!
```

The rebuild blanks **every album and artist cover link in the library**. And `PRAGMA foreign_keys` can't rescue it: it's a no-op inside a transaction, which is exactly where sqlx runs migrations.

`ALTER TABLE … ADD COLUMN` cannot do that. Seven mechanical call-site edits are a far better trade than a migration that silently wipes a column across every install.

## ⚠️ A separate, pre-existing problem this uncovered

The same mechanism means **`20260428000000_drop_metadata_fks.sql`, which is already shipped, destroyed data when it ran.** It rebuilds `artist` and `album` via `DROP TABLE`, and their children are `track_artist.artist_id ON DELETE CASCADE`, `track.primary_artist ON DELETE SET NULL`, `track.album_id ON DELETE SET NULL`.

Reproduced faithfully:

```
artist rows conservés : 1 (attendu 1)
track_artist restants : 0  -> !!! CASCADE A EFFACÉ LA LIAISON !!!
track.primary_artist  : None -> !!! SET NULL A FRAPPÉ !!!
```

Every user who upgraded through that migration lost their multi-artist links and primary-artist links until their next scan. `PRAGMA legacy_alter_table = ON` at the top of that file does not help — it governs how `ALTER TABLE RENAME` rewrites references, not foreign-key enforcement.

Not fixable in place (migrations are immutable, and the rows are gone — the scanner rebuilds them on rescan). **I have not opened an issue for it yet — say the word and I will**, with the reproduction. It may also explain past reports of artists or albums looking wrong after an update.

Both `CLAUDE.md` and `docs/architecture/storage.md` now carry the rule: never `DROP TABLE` a parent in a migration.

## A test-fidelity bug I found in my own tests

The fixture had no CHECK constraint, and I had been seeding `source = 'user'`. Production cannot hold that value — the schema allows only `embedded` / `folder` / `deezer` / `manual`. The tests were asserting against a state that cannot occur. Fixture now carries the real constraint; tests use `manual` and are renamed accordingly.

## Tests

Two regression cases for the exact #401 setup:

- a sidecar album **stays refreshable** when another album embeds the same image (the freeze)
- an embedded album **stays protected** even when the shared row happens to say `folder` (the mirror case, so the fix isn't just inverted)

## Verification

- `cargo test -p waveflow-core --features sqlite` ✅ **122/122**
- `bun run typecheck` ✅
- Migration run against a real database: links preserved, provenance backfilled, an album without a cover left without one.

## Summary by CodeRabbit

- **Améliorations**
  - La provenance des pochettes d’album est désormais conservée séparément pour chaque album, même lorsque plusieurs albums partagent la même image.
  - Les pochettes provenant de Deezer, de fichiers manuels, des tags, de dossiers ou de sidecars sont mieux identifiées.
  - Les mises à jour automatiques respectent davantage les pochettes déjà ajoutées manuellement.

- **Documentation**
  - Documentation enrichie sur la gestion des pochettes et les règles de migration des données.


InstaZDLL:
@coderabbitai Valid, fixed in 04d7474 — and the root cause turned out to be worth removing rather than working around.

## Why these sites weren't transactional

`deezer.rs` carried its own `upsert_artwork_row`: a byte-for-byte copy of core's `upsert_artwork`, differing only in taking `&SqlitePool` instead of `&mut SqliteConnection`. Taking the pool is *exactly* what makes a helper unable to join a caller's transaction — and `CLAUDE.md` states the opposite as the rule for this family:

> Every helper takes `&mut sqlx::SqliteConnection` so it can participate in the open transaction … never reach for the pool from inside these.

So instead of wrapping the duplicate in `pool.begin()` as suggested, I deleted it and pointed all callers at the shared helper. Same outcome, one fewer copy of the same function, and the next site that needs it can't accidentally reintroduce the problem.

## Scope: four sites, not two

You named the two album paths. The two **artist** paths (`deezer.rs:786`, `:824`) have the identical shape and route through the same duplicate helper — fixing half would have left it alive to serve the other half. All four are transactional now.

On the artist paths there's a detail worth stating: the `rows_affected() == 0` early return now leaves the transaction **un-committed**, so a missing artist rolls the artwork insert back rather than orphaning it. That's a small behavioural improvement the transaction gives for free.

`edit.rs` also stops doing `pool.acquire()` → `drop(conn)` → update on a *second* pooled connection; it's one transaction now.

## On severity — stated plainly

I don't think this was a live defect. `artwork` is content-addressed, so an orphan row is inert and the next upsert of the same bytes reuses it; the album simply kept its old cover and the user retried. Nothing tears, nothing corrupts. And since #401 moved provenance onto the album, `artwork_id` and `artwork_source` are set by a single statement and can't diverge.

I made the change for the reason your comment implies rather than the one it states: these two statements belong together, and someone adding a third to one of these blocks would reasonably assume they already were atomic.

- `cargo test -p waveflow-core --features sqlite` ✅ 122/122
- Net effect on `deezer.rs`: −1 line, having removed a 24-line duplicate.

Issue #174, opened 2026-05-28
bug(audio): WASAPI Exclusive should retry S24/S16 before falling back to shared


InstaZDLL:
## Symptom

On Windows with the WASAPI Exclusive Mode toggle enabled, init fails on every stream open with:

```
WARN waveflow_lib::audio::wasapi_exclusive: wasapi exclusive init failed
    err=Audio("device does not support 32-bit float in exclusive mode: Windows(Error { code: HRESULT(0x88890008), message: \"\" })")
WARN waveflow_lib::audio::output: WASAPI Exclusive init failed, falling back to shared mode
INFO waveflow_lib::audio::output: cpal output stream opened sample_rate=44100 channels=8 sample_format=F32
```

The fallback to cpal shared mode works fine, but the user gets two `WARN` log lines on every track load and never actually obtains bit-perfect playback — defeating the point of enabling the toggle.

## Root cause

[`audio/wasapi_exclusive.rs`](https://github.com/InstaZDLL/WaveFlow/blob/main/src-tauri/crates/app/src/audio/wasapi_exclusive.rs) requests the IAudioClient with `WAVE_FORMAT_IEEE_FLOAT` (32-bit float, what cpal uses internally). Many consumer DACs and integrated codecs (Realtek ALC, Conexant CX, MOTU's UltraLite in some modes…) **only accept** `WAVE_FORMAT_PCM` with `wBitsPerSample = 16` or `24` in exclusive mode — they reject `IEEE_FLOAT` outright with `AUDCLNT_E_UNSUPPORTED_FORMAT` (0x88890008).

WASAPI Exclusive is supposed to negotiate format down — the canonical pattern is `IAudioClient::IsFormatSupported` with progressively narrower types until one is accepted, then we convert F32 samples to that bit depth in the decoder before pushing to the ring.

## Fix sketch

In `wasapi_exclusive::open` (Windows):

1. Try `IEEE_FLOAT` 32-bit at the target sample rate (current behaviour)
2. On `AUDCLNT_E_UNSUPPORTED_FORMAT`, try `PCM` 24-bit packed (`wBitsPerSample = 24`, `nBlockAlign = nChannels * 3`)
3. On the same error, try `PCM` 24-bit padded (`wBitsPerSample = 24`, container 32-bit, `nBlockAlign = nChannels * 4`)
4. On the same error, try `PCM` 16-bit
5. If all four fail, fall back to shared mode as today

[...]

PR #410, opened 2026-07-19
fix(audio): negotiate WASAPI exclusive against the endpoint format

### LINKED ISSUE #409: WASAPI Exclusive: negotiation uses the shared mix format, so surround-configured devices are always rejected
## Problem

`try_open_exclusive` takes its sample rate and channel count from the **shared-mode mix format**:

```rust
// wasapi_exclusive.rs
let sample_rate = mix_format.get_samplespersec() as usize;
let channels    = mix_format.get_nchannels() as usize;
```

and then varies only the sample *format* across `FORMAT_FALLBACK_CHAIN` (F32 → S24_3LE → S24_4LE → S16_LE).

The mix format describes how the Windows audio engine **mixes for shared mode**. It is not what the device accepts in exclusive mode, and on a surround-configured endpoint the two are wildly different.

Real failure, on a headset configured as 7.1:

```
WARN wasapi exclusive init failed
     err=is_supported rejected S16_LE in exclusive mode: HRESULT(0x88890008)
INFO cpal output stream opened sample_rate=44100 channels=8 sample_format=F32
```

`0x88890008` is `AUDCLNT_E_UNSUPPORTED_FORMAT`, and all four candidates were rejected — because every one of them was offered at **44100 Hz / 8 channels**. That 7.1 is a *virtual* rendering produced by the Windows mixer. Exclusive mode bypasses the mixer, so there is nothing left to synthesise eight channels: no hardware endpoint is going to accept it.

The ladder cannot recover, because it only ever moves along one axis.

## Why this matters beyond one user

Anyone running a surround-configured output — gaming headsets with virtual 7.1 are the common case, and they are exactly the audience likely to try an "exclusive/bitperfect" toggle — gets a hard rejection with no path to success. The feature simply cannot be enabled, and the log gives no hint that the channel count is the culprit.

It also makes the failure look like a hardware limitation when it is our negotiation choosing an impossible target.

## Expected

Negotiate against what the device declares for exclusive mode, not against the mixer.

Windows exposes it: the endpoint property store's `PKEY_AudioEngine_DeviceFormat` is the "Default Format" shown under **Sound → device → Properties → Advanced**, and it is precisely the format exclusive mode is built around. Trying that first would resolve this case outright.

Failing that, or in addition, the ladder should be able to move on more than one axis — falling back to stereo, and to the common rates (48000, 44100), rather than treating the mix format's rate and channel count as fixed.

Worth logging the attempted `(rate, channels, format)` triple on rejection too. The current message names only the sample format, which is what sent the initial diagnosis in the wrong direction here.

## Acceptance

- a device configured for virtual surround can still open in exclusive mode, at whatever channel count it genuinely supports
- rejection logs say what was attempted, all three parameters
- devices that already work (`S24_4LE` and `S16_LE` at 48000/2 in the wild) keep working — this must not regress the negotiation that currently succeeds

## Workaround meanwhile

Set the output device to stereo in Windows Sound (or disable virtual surround in the vendor's software) before enabling WASAPI Exclusive.

## Notes

Distinct from #405 (position counter, fixed in #407) and from the device-reset failures in #355/#365 — this one never gets as far as opening a stream.

InstaZDLL:
Fixes #409.

## The bug

Exclusive-mode negotiation asked the device for the layout reported by `IAudioClient::GetMixFormat`. That is the **shared-mode mix format** — the shape of the pipe into the Windows audio engine, *after* whatever the engine puts in front of the endpoint. With Audio Enhancements, Spatial Sound or a virtual-surround driver active it comes back multi-channel for a physically stereo jack.

On the reporting machine: Windows shows the endpoint as **stereo 44.1 kHz**, cpal reports **8 channels**. Every entry in the format chain was then rejected with `AUDCLNT_E_UNSUPPORTED_FORMAT` (`0x88890008`) — not because the bit depth was wrong, but because no consumer codec opens 8 channels in exclusive mode. The toggle silently fell back to shared mode, so bit-perfect output was simply unreachable.

Disabling the audio enhancements does **not** work around it: the log still showed 8 channels afterwards.

## The fix

Negotiation now walks **layout × bit depth** instead of one layout × bit depth.

Layouts, in order:

1. `endpoint` — `PKEY_AudioEngine_DeviceFormat`, the "Default Format" picked in the Windows Sound control panel. This describes the endpoint itself rather than the engine in front of it, and is what exclusive mode actually wants.
2. `mix` — the previous behaviour, kept as a fallback.
3. `stereo` — 2 channels at each of the rates above, for drivers whose reported default no application can open.

`wasapi` 0.23 already exposes the property read as `Device::get_device_format()`, so no raw Win32 was needed.

Identical pairs are deduped. On the majority of machines the endpoint and mix formats agree, so those users probe **exactly as many times as before** — this is additive, not a behaviour change for the working case.

Rejections now log the full `(rate, channels, format, origin)` tuple. The old line named only the bit depth, which is precisely what made a channel-count mismatch read as a format problem for weeks.

## Notes

- The negotiated layout is stored into `SharedPlayback.{sample_rate,channels}` before init is reported to the caller, and the decoder thread is spawned only afterwards — so it resamples and downmixes to what the device accepted, not to what the mix format claimed. That ordering already held; I documented it rather than changed it.
- Docs: `docs/architecture/audio.md` also still claimed exclusive mode was "32-bit float stereo, anchored on `KSDATAFORMAT_SUBTYPE_IEEE_FLOAT`" — stale since the format chain landed in #174. Corrected in the same pass.

## Testing

The ordering/dedupe logic is split into a pure `build_layout_candidates` so it is testable without a COM apartment or real hardware. 7 unit tests cover the reporter's exact case, the agreeing-formats common case, both-inflated, disagreeing rates, a driver that doesn't publish the key, no usable format, and malformed formats.

**I could not execute them locally**: `cargo test -p waveflow --lib` aborts with `STATUS_ENTRYPOINT_NOT_FOUND` on this Windows box (a pre-existing Tauri DLL issue, not related to this change). To avoid claiming untested code, I extracted the function plus its tests into a standalone binary and ran them there — 7/7 pass. CI runs them for real on the Windows runner.

`cargo clippy -p waveflow --all-targets` is clean.

The end-to-end path needs hardware confirmation from someone whose device exhibits the mismatch.

## Summary by CodeRabbit

  * Amélioration du mode exclusif WASAPI sous Windows grâce à une négociation plus fiable du format audio.
  * Prise en charge prioritaire du format recommandé par le périphérique, avec plusieurs formats de secours.
  * Conservation de la lecture audio via le mode partagé en cas d’échec du mode exclusif.

* **Documentation**
  * Mise à jour de la documentation sur la sélection des formats, les compromis et le comportement de bascule.

### DIFF
--- a/docs/architecture/audio.md
+++ b/docs/architecture/audio.md
@@ -44,15 +44,34 @@
 [`audio/wasapi_exclusive.rs`](../../src-tauri/crates/app/src/audio/wasapi_exclusive.rs) is a parallel output backend to the cpal shared-mode default. Engaged via the `audio.wasapi_exclusive` profile setting (toggle in Settings → Audio). When on:
 
 1. `output::spawn_output_with_mode` tries the exclusive backend first via the [`wasapi` crate](https://crates.io/crates/wasapi).
-2. The backend opens the device in **event-driven exclusive mode** at the device's mix-format sample rate (whatever the user picked in the Windows Sound control panel). 32-bit float stereo, anchored on `KSDATAFORMAT_SUBTYPE_IEEE_FLOAT`.
-3. If init fails (device busy with another exclusive app, no float-32 support, COM apartment conflict), the engine logs a warning and falls back transparently to the cpal shared backend so the user keeps hearing audio.
+2. The backend opens the device in **event-driven exclusive mode**, negotiating over two axes (see below).
+3. If every candidate fails (device busy with another exclusive app, no supported format, COM apartment conflict), the engine logs a warning and falls back transparently to the cpal shared backend so the user keeps hearing audio.
+
+### Format negotiation (two axes)
+
+`open_exclusive_session` walks **layout × bit depth** and takes the first pair the driver accepts.
+
+**Layout** — `(sample rate, channel count)`, from `build_layout_candidates`, in order:
+
+| # | Origin     | Source                                                                                           |
+|---|------------|--------------------------------------------------------------------------------------------------|
+| 1 | `endpoint` | `PKEY_AudioEngine_DeviceFormat` — the "Default Format" picked in the Windows Sound control panel |
+| 2 | `mix`      | `IAudioClient::GetMixFormat` — the shared-mode mix format                                        |
+| 3 | `stereo`   | 2 channels at each of the rates above                                                            |
+
+The endpoint format leads on purpose (#409). The mix format describes the pipe into the Windows audio engine **after** its own processing, so with Audio Enhancements, Spatial Sound or a virtual-surround driver active it reports e.g. 8 channels for a plain stereo jack. Negotiating exclusive mode against that asks the hardware for a layout it has never supported: every format is rejected with `AUDCLNT_E_UNSUPPORTED_FORMAT` (`0x88890008`) and the user silently lands back in shared mode. `PKEY_AudioEngine_DeviceFormat` describes the endpoint itself, which is what exclusive mode wants. The two agree on most machines, so candidate 2 is deduped away.
+
+**Bit depth** — `FORMAT_FALLBACK_CHAIN`, high to low: `F32` → `S24_3LE` → `S24_4LE` → `S16_LE`. Most audiophile DACs take Float32 natively; Realtek ALC and many integrated codecs reject it outright but accept PCM.
+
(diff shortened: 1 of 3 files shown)


InstaZDLL:
@coderabbitai review

PR #413, opened 2026-07-19
fix(dlna): rebuild the worker's resources on a profile switch

### LINKED ISSUE #399: DLNA worker keeps serving the previous profile after a profile switch
## Problem

The DLNA worker snapshots the active profile's pool + artwork dirs into `DlnaResources` when the server starts ([`commands/dlna.rs::build_resources`](../blob/main/src-tauri/crates/app/src/commands/dlna.rs)), and holds that snapshot for the life of the server.

`build_resources` is only ever called from two places: the boot path in `lib.rs` and the `dlna_set_config` command. **`switch_profile` never touches it.**

So after a profile switch with DLNA running:

1. the worker keeps serving the *previous* profile's library to the LAN — a privacy-adjacent surprise if the user switched to a different profile precisely to change what is exposed;
2. that profile's pool is then closed by `activate_profile`, so every subsequent `/stream/<id>` and `/art/<hash>` request fails against a closed pool until the user stops and restarts the server.

The stale `DlnaResources` also keeps `profile_artwork_dir` pointing at the old profile's artwork tree.

## Not a regression

This predates the lease work in #332 — the previous code closed the old pool immediately too, so the failure mode is unchanged. #332 only made the situation explicit, since `build_resources` is the one caller of the new `into_unleashed()` escape hatch and its comment had to say *why* it opts out. The docstring there and in `docs/architecture/storage.md` now describe the real behaviour and point here.

## Expected

`switch_profile` (and `deactivate_profile`) should rebuild or stop the DLNA worker so it never retains resources from a profile that is no longer active:

- if the server is running (`state.dlna.status().await.running`), rebuild via `build_resources` for the new profile and restart it with the existing `DlnaConfig`;
- stop the old worker before replacing it, so the SSDP announcer and the axum bind are not duplicated;
- on `deactivate_profile` (no active profile), stop the server rather than leaving it serving a closed pool.

## Notes

Needs verification with an actual UPnP client (VLC / a TV) — the failure is invisible from the app's own UI, which is likely why it went unnoticed. Worth checking whether the announced server name / UUID should stay stable across the restart so clients do not treat it as a new device.

Found during review of #398 (thanks @coderabbitai for flagging the lifecycle claim).

InstaZDLL:
## Summary

Fixes #399.

`DlnaResources` (the worker's pool + artwork dirs) was only ever rebuilt from the boot path and `dlna_set_config`. `switch_profile` never touched it, so a running DLNA server:

1. kept exposing the **previous** profile's library to the LAN after a switch — a privacy-adjacent surprise, since switching profiles is often exactly how a user changes what's exposed;
2. then had its pool closed out from under it once `activate_profile` dropped the old profile, so every subsequent `/stream/<id>` and `/art/<hash>` request failed against a closed pool.

`switch_profile` now rebuilds `DlnaResources` from the newly-activated profile and pushes them through `DlnaServer::start` — but only when `state.dlna.status().await.running` is true, so a switch can't accidentally start a server the user has turned off. There's no lighter "just swap the pool" path on the worker (`WorkerState::start` already tears down and rebinds on every `Start`), so this restarts the listener + SSDP announce, same as a manual reconfigure.

## Test plan

- [x] `cargo check -p waveflow --all-targets` — clean, no warnings
- [ ] Manual: enable DLNA, switch profiles, confirm a controller (or `curl` against `/ContentDirectory/control`) sees the new profile's library and `/stream/<id>` no longer 500s

## Summary by CodeRabbit

* **Améliorations**
  * Le service DLNA actualise désormais automatiquement ses ressources lors d’un changement de profil, lorsqu’il est déjà actif.
  * Le changement de profil continue de fonctionner même si la génération des ressources DLNA échoue.

### DIFF
--- a/src-tauri/crates/app/src/commands/dlna.rs
+++ b/src-tauri/crates/app/src/commands/dlna.rs
@@ -41,10 +41,10 @@ pub async fn dlna_get_status(state: tauri::State<'_, AppState>) -> AppResult<Dln
 /// Snapshot the per-profile pool + artwork dirs into a `DlnaResources`
 /// the worker thread can hold across requests.
 ///
-/// Only called from the boot path and from `dlna_set_config` — notably
-/// **not** from `switch_profile`, so a running server keeps serving the
-/// profile it was started with until it is stopped and restarted. That
-/// gap predates the lease work and is tracked in issue #399.
+/// Called from the boot path, from `dlna_set_config`, and from
+/// `switch_profile` when the worker is already running (issue #399) —
+/// a running server otherwise keeps serving the profile it was started
+/// with, and its pool goes stale the moment the old profile is closed.
 pub async fn build_resources(state: &AppState) -> AppResult<DlnaResources> {
     // Deliberately unleashed: the worker holds this pool for the life of
     // the server, not for the span of a command, so a lease would stall
--- a/src-tauri/crates/app/src/commands/profile.rs
+++ b/src-tauri/crates/app/src/commands/profile.rs
@@ -163,6 +163,30 @@ pub async fn switch_profile(
 
     state.activate_profile(profile_id).await?;
 
+    // Refresh the DLNA worker for the new profile if it's currently
+    // serving — otherwise it keeps exposing the previous profile's
+    // library over the LAN, and its pool becomes closed out from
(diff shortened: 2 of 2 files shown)


InstaZDLL:
@coderabbitai review

PR #418, opened 2026-07-20
feat(artwork): let a user set an album's motion cover manually

### LINKED ISSUE #408: Allow setting an album's animated cover manually from a local file
## What

Let the user pick their own animated cover for an album from a local file, the same way they can already upload a static one.

Today motion artwork only arrives through a plugin: [`fetch_album_motion_artwork`](../blob/main/src-tauri/crates/app/src/commands/motion_artwork.rs) fans `album-info` out to enabled metadata plugins and takes the `motion-cover-url` they return. That covers the catalogue the plugin can match and nothing else — a self-released album, a bootleg, a rip the plugin doesn't recognise, or simply a loop the user made themselves has no way in.

Static covers already have this escape hatch (`set_album_artwork_from_file`, stored with `source = 'manual'`). Motion artwork should have the same.

### 1. It must not live in the LRU cache

The obvious place is wrong. `motion_cache/` is a **1 GB LRU with mtime-based eviction** ([`motion_cache::DEFAULT_MAX_CACHE_BYTES`](../blob/main/src-tauri/crates/core/src/artwork/motion_cache.rs)) — it exists precisely because plugin-fetched videos are large and disposable. A file the user deliberately chose is neither, and dropping it there means their cover silently disappears once someone else's downloads push the cache over the cap.

A user-supplied file belongs in a per-profile directory that is never evicted, alongside `artwork/`. That also needs its own **asset-protocol scope entry** in `tauri.conf.json` — `motion_cache/**` is declared, a new directory would not be covered by it.

### 2. GIF does not play in a `<video>` tag

[`MotionCoverOverlay`](../blob/main/src/components/player/MotionCoverOverlay.tsx) renders `<video src=…>`. A GIF is an image, not a video stream — it will not render there at all. So accepting GIF means one of:

- **branch the overlay** on file type and render an `` for GIFs (simplest; loses `muted`/`playsInline` semantics, and a long GIF is far heavier than the equivalent mp4)
- **transcode to mp4 on import** (best result, but pulls in an encoder dependency for a niche path — hard to justify)
- **accept mp4 only** and say so in the picker

My inclination is the third for a first cut, with GIF added later if people ask. Worth deciding explicitly rather than discovering it at implementation time.

### 3. Container and codec

The desktop webview plays **h264 in mp4** reliably; there is no HLS.js, which is why plugins are already required to return a directly-playable mp4. WebM/VP9 support varies between WebView2 and WebKitGTK. Validating by magic bytes (as `set_album_artwork_from_file` does for images) and rejecting what won't play beats letting the user pick a file that renders as a black rectangle.

### 4. Precedence

A manual pick must **win over the plugin** and survive re-fetches — the same rule static artwork follows, where `source = 'manual'` is never overwritten by the scanner or by enrichment (see #401). Without that, the next plugin refresh silently replaces the user's choice.

Removing a manual motion cover should fall back to the plugin result rather than to nothing.

### 5. Scope

Albums only, to start. That is the level the current pipeline works at, and it is where the visual payoff is. Per-track motion artwork would be a much larger change for a much rarer want.

## Acceptance

- an album can be given a local mp4 as its animated cover, and cleared again
- the file survives cache eviction, app restarts, and plugin re-fetches
- a manual cover takes precedence over a plugin-supplied one; clearing it restores the plugin's
- unplayable files are rejected at pick time with a clear reason, not accepted and rendered blank
- the picker states which formats are accepted

## Notes

Sizes here are an order of magnitude above static artwork, so a per-file cap and some visibility of total usage (the motion cache already surfaces its size under Settings → Plugins) are probably worth carrying over.

InstaZDLL:
## Summary

Closes #408.

Motion covers previously only arrived through a metadata plugin (`fetch_album_motion_artwork` fanning `album-info` out to enabled plugins) — that covers whatever the plugin's catalogue can match and nothing else. Static covers already have a manual escape hatch (`set_album_artwork_from_file`); this gives motion covers the same one.

**Design points from the issue, implemented as specified:**
- Stored in a never-evicted per-profile directory (`profile_motion_dir`), **not** the 1 GB LRU `motion_cache_dir` — that cache exists precisely because plugin-fetched videos are disposable, and a user's deliberate choice isn't.
- **mp4 only** for this first cut (validated by magic bytes — the `ftyp` box — rejected at pick time, not rendered as a black rectangle). No GIF, no transcoding.
- A manual pick **wins over the plugin outright** and survives re-fetches — `fetch_album_motion_artwork` checks the new `album_motion_artwork` table before any plugin runs. Clearing it falls back to the plugin result.
- Albums only, per the issue's stated scope.

## Changes

**Backend**
- New migration: `album_motion_artwork` table (fresh `CREATE TABLE`, no touch to `album`/`artwork`).
- `paths.rs`: `profile_motion_dir`, added to `ensure_profile_dirs`.
- `commands/motion_artwork.rs`: `set_album_motion_artwork_from_file` (magic-byte validation + 64 MiB cap, hash-addressed), `clear_album_motion_artwork`, and `fetch_album_motion_artwork` gained an optional `album_id` to check the manual override first.

**Frontend**
- New `MotionCoverPickerModal` (file-only, mirrors `ArtistImagePickerModal`'s clear-button pattern), wired into `AlbumDetailView` next to the existing static cover picker.
- `useAlbumMotionArtwork` / `MotionCoverOverlay` / `plugins.ts` thread `albumId` through; `NowPlayingPanel` + `ImmersiveNowPlaying` pass `currentTrack.album_id`.
- i18n keys propagated to all 17 locales.

## Test plan

- [x] `cargo check --workspace --all-targets` — clean, no warnings
- [x] `cargo test -p waveflow --lib` — 120/120 passing, no regressions
- [x] `bun run lint` — clean
- [x] All 17 locale JSON files validated for parity + valid JSON
- [ ] Manual: pick an mp4 for an album, confirm it overrides a plugin-resolved cover in Now Playing / Immersive view, confirm clearing it falls back — not done in this session (no motion-artwork plugin installed to compare against)

## Summary by CodeRabbit

  * Définissez une pochette animée personnalisée pour un album à partir d’un fichier MP4 local.
  * Supprimez la pochette personnalisée pour revenir à celle fournie automatiquement.
  * Les pochettes personnalisées sont conservées et prioritaires lors de la lecture.
  * Les nouvelles options sont disponibles dans la vue détaillée de l’album et traduites dans plusieurs langues.

### DIFF
--- a/src-tauri/crates/app/src/commands/motion_artwork.rs
+++ b/src-tauri/crates/app/src/commands/motion_artwork.rs
@@ -282,9 +282,14 @@ pub async fn set_album_motion_artwork_from_file(
     let pool = state.require_profile_pool().await?;
     let profile_id = state.require_profile_id().await?;
     let motion_dir = state.paths.profile_motion_dir(profile_id);
-    std::fs::create_dir_all(&motion_dir)?;
+    tokio::fs::create_dir_all(&motion_dir).await?;
 
-    let bytes = std::fs::read(&file_path)?;
+    // A motion cover can be up to MAX_MANUAL_MP4_BYTES (64 MiB) — an order
+    // of magnitude past the static-cover uploads this command's shape is
+    // otherwise modelled on, so the blocking read/write is routed through
+    // `tokio::fs` (spawn_blocking under the hood) instead of `std::fs`
+    // directly, unlike those smaller precedents.
+    let bytes = tokio::fs::read(&file_path).await?;
     if bytes.len() as u64 > MAX_MANUAL_MP4_BYTES {
         return Err(AppError::Other(format!(
             "file too large: {} bytes (max {MAX_MANUAL_MP4_BYTES})",
@@ -297,8 +302,8 @@ pub async fn set_album_motion_artwork_from_file(
 
     let hash = blake3::hash(&bytes).to_hex().to_string();
     let target = motion_dir.join(format!("{hash}.mp4"));
-    if !target.exists() {
-        std::fs::write(&target, &bytes)?;
+    if !tokio::fs::try_exists(&target).await? {
+        tokio::fs::write(&target, &bytes).await?;
     }
 
     sqlx::query(
--- a/src-tauri/crates/app/src/commands/motion_artwork.rs
+++ b/src-tauri/crates/app/src/commands/motion_artwork.rs
@@ -51,6 +51,13 @@ pub struct MotionArtwork {
 /// `motion-cover-url`. Returns `None` when offline, when no metadata plugin
 /// is installed, or when none has motion artwork for the album.
 ///
+/// When `album_id` resolves a row in `album_motion_artwork` (issue #408),
+/// that user-supplied file wins outright and the plugin fan-out below never
+/// runs — same precedence a manual static cover already gets. `album_id` is
+/// `Option` because a couple of call sites (radio, any future text-only
+/// lookup) don't always have one; they simply lose the manual-override path
+/// and behave exactly as before this field existed.
+///
 /// The lookups fan out (each on its own blocking task) and the first hit
 /// wins; the rest are dropped. Every lookup is bounded by [`PLUGIN_TIMEOUT`],
 /// so one slow or hung plugin can't stall the others or the now-playing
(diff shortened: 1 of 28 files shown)


InstaZDLL:
@coderabbitai review

PR #419, opened 2026-07-20
feat(scan): add a per-folder deep rescan action

InstaZDLL:
## Summary

Addresses the remaining half of issue #366's Symptom A (not auto-closing — this needs runtime confirmation, see Test plan).

The scanner's Phase 1 fast path in `scan_folder_inner` skips a file whose `(mtime, size)` matches the stored row — cheap, and what makes re-scans of a large library fast. But it means a tag or embedded-picture edit made by an external tool that preserves the file's mtime (Mp3tag has a "preserve modification time" option, among others) is invisible: the file never reaches the hash/re-extract step that would actually notice the change.

`refresh_folder_covers` already closed the *sidecar* `cover.jpg` half of this symptom (a prior PR). This closes the *embedded* tag/picture half — the part still open.

## Changes

- `scan_folder_inner` gains a `deep: bool` parameter. When true, Phase 1's `(mtime, size)` check is skipped entirely, so every file in the folder is re-hashed and re-read (`hash_file`'s doc comment already notes a tag rewrite shifts bytes in the hashed window, so the hash-compare step downstream correctly picks up the edit — Phase 1 was the only thing standing in the way).
- Every existing caller (fs-watcher, scan-on-start, `rescan_library`, `import_paths`) passes `false`, so normal scans are unaffected.
- New optional `deep` argument on the `scan_folder` command, wired to a per-folder action (spinning refresh icon) in the library's "Dossiers" tab — the only place a user targets an individual folder today.
- **Deliberately scoped to per-folder**, not a `rescan_library`/global Settings option: the reported problem is "I edited one album's tags," not "make every scan slow by default."
- i18n key (`library.folderList.deepRescan`) propagated to all 17 locales.

## Test plan

- [x] `cargo check --workspace --all-targets` — clean
- [x] `cargo test -p waveflow --lib` — 120/120 passing
- [x] `bun run lint` — clean
- [ ] Manual: edit a track's genre with an external tool that preserves mtime, confirm the normal rescan still misses it, confirm the new deep-rescan action picks it up — not done in this session (no such external-tool edit scenario reproduced locally)

## Summary by CodeRabbit

  - Ajout d’un bouton de rescan approfondi pour chaque dossier de bibliothèque.
  - Le rescan approfondi relit et réanalyse tous les fichiers afin de détecter les changements de métadonnées ou de pochettes, même sans modification de date.
  - Un indicateur de progression s’affiche pendant l’opération, et les rescans simultanés sont empêchés.

- **Améliorations**
  - Les libellés de cette fonctionnalité sont désormais disponibles dans les langues prises en charge.
  - Les scans habituels conservent leur comportement rapide.


InstaZDLL:
@coderabbitai review

Issue #405, opened 2026-07-19
WASAPI: audio recovers after a device flap but player state does not (lyrics, seek bar, toggle)


InstaZDLL:
Split out of #403 (reported by @jo-el414 on `1.6.4-beta.1`), which bundled two unrelated problems.

## The audio recovery itself now works

Worth stating first, because it changes what this issue is about. The #365 fix shipped in `1.6.4-beta.1` and his log confirms all three acceptance criteria:

- **one** `rebuilding cpal output after DeviceNotAvailable`, not two or more — the self-sustaining loop is gone
- no `0x8889000A` / `AUDCLNT_E_DEVICE_IN_USE`
- exclusive renegotiated and re-engaged in **42 ms**, decoder resumed the same track

```
08:08:52.951614Z INFO engine: rebuilding cpal output after DeviceNotAvailable device="<os-default>" exclusive=true
08:08:52.993532Z INFO wasapi_exclusive: negotiation succeeded sample_rate=48000 channels=2 format="S16_LE"
08:08:52.993584Z INFO output: audio output: WASAPI Exclusive Mode engaged
08:08:52.994721Z INFO decoder: decoding start dst_sample_rate=48000 track_id=123
```

He confirms it in prose too: *"the music keep playing"*. Before, playback died until a restart.

## The remaining problem: sound recovers, app state does not

After a device flap, with audio still playing:

- **lyrics stop following** — side panel and fullscreen both
- **the seek bar disappears**
- **the WASAPI Exclusive toggle in Settings is stuck on** and can't be turned off
- only a full restart clears it

## Candidate mechanism (unconfirmed)

Position events are emitted **only while the player state is `Playing`** — [`decoder.rs:1173`](../blob/main/src-tauri/crates/app/src/audio/decoder.rs#L1173):

```rust
if last_position_emit.elapsed() >= POSITION_EMIT_INTERVAL
    && shared.state() == PlayerState::Playing
{
```

If the rebuild leaves `shared.state()` as anything other than `Playing` while the ring buffer keeps draining, you get exactly this shape: sound continues, and everything downstream of position freezes. Lyrics sync is position-driven, and the seek bar has nothing to render.

The rebuild path does try to restore state — [`force_rebuild_output`](../blob/main/src-tauri/crates/app/src/audio/engine.rs) captures `was_playing` / `track_id` / `position_ms` and re-dispatches `LoadAndPlay` — so if the theory holds, the bug is in that handoff rather than in a missing resume.

[...]

PR #421, opened 2026-07-20
fix(audio): resync the WASAPI exclusive toggle after an automatic rebuild

InstaZDLL:
## Summary

Addresses part of #405 — not auto-closing, see Test plan. Thanks @jo-el414 for the 1.6.5-beta.1 log + screenshot that pinned this down.

His log confirmed the position/lyrics/seek-bar fix (#407) holds and the `RebuildGate` (#393) works as designed — both the immediate and the 300ms-gated rebuild succeeded in exclusive mode, no `AUDCLNT_E_DEVICE_IN_USE`. But the Settings toggle still showed stuck ON afterward, alongside a raw `build_output_stream` error banner.

Root cause, traced through the code: `ExclusiveModeCard.tsx` only ever called `playerGetWasapiExclusive()` on mount and immediately after a manual click. Nothing re-checked it after the engine rebuilt the output stream on its own — a device flap, or a device-picker switch. In the reporter's log the toggle happened to still read correctly because the rebuild landed back in exclusive mode; but nothing would have caught it silently falling back to shared, which is the scarier version of the same bug.

## Fix

`engine.rs` now emits `player:audio-mode-changed` from all three sites that write `wasapi_exclusive_active`: the automatic device-error rebuild, the output-device switch, and the manual toggle path itself (redundant there, kept for consistency). `ExclusiveModeCard` listens and re-fetches — same shape as `PlayerContext`'s existing `sample_rate`/`channels` refresh on `player:track-changed`.

## Test plan

- [x] `cargo check --workspace --all-targets` — clean
- [x] `cargo test -p waveflow --lib` — 120/120 passing
- [x] `bun run lint` — clean
- [ ] Manual: reproduce a device flap on the reporter's hardware, confirm the toggle now updates on its own instead of needing a restart to trust — not done in this session (Windows/WASAPI-only, no matching hardware here)

## Summary by CodeRabbit

* **Améliorations**
  * Le réglage du mode audio exclusif WASAPI se met à jour automatiquement sur Windows lorsque le moteur audio change de mode.
  * L’interface reflète plus fidèlement l’état WASAPI réel après une reconfiguration du flux, notamment lors d’un basculement de périphérique ou d’un rebuild forcé.
  * La synchronisation est assurée y compris pour les changements effectués en dehors de l’interaction avec le bouton.

### DIFF
--- a/src-tauri/crates/app/src/audio/engine.rs
+++ b/src-tauri/crates/app/src/audio/engine.rs
@@ -12,7 +12,7 @@ use std::time::{Duration, Instant};
 
 use crossbeam_channel::{unbounded, Sender};
 use rtrb::Producer;
-use tauri::AppHandle;
+use tauri::{AppHandle, Emitter};
 use tokio::sync::mpsc::unbounded_channel;
 
 use crate::error::{AppError, AppResult};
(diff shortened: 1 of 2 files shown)


InstaZDLL:
@coderabbitai review

InstaZDLL:
@copilot le fix ne contient pas de régression

Issue #365, opened 2026-07-12
WASAPI Exclusive: concurrent rebuild paths collide on DeviceNotAvailable (AUDCLNT_E_DEVICE_IN_USE 0x8889000A)


InstaZDLL:
Sub-issue of #355 — concrete root-cause from the reporter's full log (`1.6.x`, exclusive mode ON, a DAC that resets its session periodically).

## What the log shows

Every device flap (seen at 09:10:01, 16:17:17, 06:55:11, 06:57:47) produces the **same** pattern — one rebuild succeeds in exclusive, a **second** concurrent rebuild collides:

```
09:10:01.341  cpal stream error err=DeviceNotAvailable
09:10:01.377  wasapi exclusive negotiation succeeded ... format="S24_4LE"   ← path A: SUCCESS (+36 ms)
09:10:01.377  wasapi exclusive stream opened
09:10:01.377  audio output: WASAPI Exclusive Mode engaged
09:10:01.378  decoder picked up new ring producer                            ← A is a full engine rebuild
09:10:01.379  decoding start track_id=432
09:10:01.653  rebuilding cpal output after DeviceNotAvailable exclusive=true ← path B starts (+312 ms)
09:10:01.674  wasapi exclusive init failed HRESULT(0x8889000A)               ← AUDCLNT_E_DEVICE_IN_USE
09:10:01.674  WASAPI Exclusive init failed, falling back to shared mode
09:10:01.680  cpal output stream opened F32
09:10:01.739  auto-rebuild after DeviceNotAvailable failed: device no longer available
```

## Analysis

- **Two rebuilds fire for one `DeviceNotAvailable`.** Path B at +312 ms is [`output.rs`'s 300 ms-delayed auto-rebuild](src-tauri/crates/app/src/audio/output.rs#L515-L528) → `engine.try_rebuild_after_device_error()`. Path A fires at +36 ms — *before* that delay — so a **second trigger** re-opens exclusive almost immediately.
- Path A's exclusive stream (S24_4LE) is still held when path B runs, so path B's `force_rebuild_output` exclusive open hits **`AUDCLNT_E_DEVICE_IN_USE (0x8889000A)`**, then its shared fallback ALSO fails (`device no longer available`).
[...]

PR #429, opened 2026-07-25
fix(audio): release the exclusive stream before re-opening the same device

### LINKED ISSUE #405: WASAPI: audio recovers after a device flap but player state does not (lyrics, seek bar, toggle)

InstaZDLL:
Closes #405

Joël's 2026-07-25 log shows the Settings toggle **still stuck on** after `v1.6.6-beta.1`. The audio/lyrics/seek-bar recovery fixed in the earlier betas holds — this is a different failure, and reading his log line by line turned up three separate defects.

## 1. Leaving exclusive mode could never succeed

`set_wasapi_exclusive` spawned the new output stream **before** releasing the old one:

```rust
let (producer, handle) =
    spawn_output_with_mode(self.shared.clone(), self.app.clone(), active, enabled)?;
// ...only then: guard.take() / old.stop()
```

That order is correct for a device *switch* — two different endpoints, so a failed spawn can roll back to the working stream. But a **mode** toggle re-opens the **same** endpoint, and a WASAPI exclusive client owns its endpoint outright: no other client, shared *or* exclusive, can open it until that client is released. So the shared open failed every time.

The consequences chain neatly into the report:

- the `?` returned `Err` **before** `player_set_wasapi_exclusive` persisted the preference,
- `wasapi_exclusive_active` was never updated and no `player:audio-mode-changed` was emitted,
- the frontend's `catch` set an error string but left `enabled` untouched,
- → the toggle stayed latched on the mode the user was trying to leave, restart being the only way out. Exactly what Joël describes, twice.

It's visible in his log at `05:26:32.931`: a bare `cpal output stream opened` with no WASAPI line and no rebuild line before it — a manual toggle-off attempt. (That message is logged in `build_stream_inner` *before* `build_output_stream` runs, so it appears even when the open then fails.)

`force_rebuild_output` already learned this in #322. Same release-first rule applied here.

## 2. The exclusive backend died silently

The WASAPI exclusive backend runs its own event loop in `wasapi_exclusive::run_event_loop`, completely separate from cpal's `err_fn`. When the device went away there — `wait_for_event` or `write_to_device` failing — the thread just `break`'d out and returned. No `player:error`, no `player:state`, no recovery scheduled. Its `AppHandle` parameter was literally named `_app`.

The engine then kept an `OutputHandle { wasapi_exclusive: true }` for a thread that no longer existed, and `wasapi_exclusive_active` stayed latched on forever.

`run_event_loop` now returns an `ExitReason`, and `DeviceLost` goes down the same two paths the cpal callback uses. Those were extracted out of `err_fn` so both backends report a device loss identically:

- `output::notify_device_lost` — park the player, emit `player:state` + `player:error`, sync the OS media controls;
- `output::schedule_device_rebuild` — 300 ms backoff, `#365` gate, same-device rebuild on the blocking pool.

Both failure arms re-check the shutdown channel first, so a deliberate teardown racing the failure is reported as `Shutdown` and doesn't schedule a rebuild against a stream the engine is replacing on purpose.

## 3. The mode switch triggered a rebuild that fought it

Seizing the endpoint exclusively kicks the outgoing shared client off it, so the toggle itself made the old stream fire `DeviceNotAvailable` — and `err_fn` dutifully scheduled a recovery rebuild for it. In Joël's log that's the second exclusive open ~300 ms after the first:

```
05:26:08.532  WASAPI Exclusive Mode engaged
05:26:08.814  rebuilding cpal output after DeviceNotAvailable  exclusive=true
05:26:08.855  WASAPI Exclusive Mode engaged      ← again
```

`begin_deliberate_output_change()` opens the existing `REBUILD_SETTLE_WINDOW` around the toggle so those self-inflicted echoes are ignored. No new mechanism — it just marks the gate finished, which is what a real rebuild already does on its way out.

## Failure paths now publish honest state

Every path that ends with **no output thread at all** — a spawn that fails after the pre-release, a `SwapProducer` hand-off to a dead decoder — now stores `wasapi_exclusive_active = false` and emits `player:audio-mode-changed` before returning the error. A toggle describing a stream that no longer exists is the exact shape of #405. The `SwapProducer` rollback in `set_wasapi_exclusive` also gained the `handle.stop()` that `force_rebuild_output` already had, so a failed hand-off no longer leaks the output thread until process exit.

`ExclusiveModeCard` re-reads the engine state on a **failed** toggle too, not just a successful one, so the UI converges on the truth either way.

## Docs

New **Output-stream lifecycle & recovery** section in [`docs/features/playback.md`](docs/features/playback.md) — the three replacement paths and which of them must release first, the two independent device-loss surfaces, and what `RebuildGate` / `FlapWindow` each protect against.

## Validation

- `cargo check --workspace --all-targets` ✅
- `cargo test -p waveflow --lib` ✅ 120 passed

⚠️ `wasapi_exclusive.rs` is `#[cfg(target_os = "windows")]`, so **it does not compile locally** — this box is Linux with no rustup and no Windows target available. CI's `windows-latest` matrix slot runs `cargo check --workspace --all-targets` and is what actually compiles that file; please let it go green before merging.

Runtime confirmation needs a Windows machine — the toggle should now switch off cleanly while exclusive mode is engaged.

## Summary by CodeRabbit

- **Améliorations**
  - Reprise plus robuste après perte ou changement de périphérique : passage en pause, notifications d’état/erreur et synchronisation des contrôles médias OS.
  - Reconstruction mieux orchestrée en cas de “device lost”, avec cibles de reprise, backoff, garde-fous anti-boucles et traitement distinct des bascules volontaires (dont WASAPI exclusif).
  - Mode WASAPI exclusif : resynchronisation automatique de l’interface après échec d’activation pour éviter un toggle bloqué.
- **Documentation**
  - Nouvelle section sur le cycle de vie de l’output stream et les scénarios de récupération.
- **Tests**
  - Test unitaire pour valider l’armement d’une reconstruction même durant la fenêtre de stabilisation.
- **Chores**
  - Mise à jour des installations Linux dans les workflows CI et CodeQL.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,18 +48,29 @@ jobs:
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v6
 
-      - name: Cache APT packages
+      - name: Install Linux build dependencies
         if: runner.os == 'Linux'
         # Tauri 2 on Linux still uses GTK 3 + WebKitGTK; cpal needs ALSA
         # development headers; the rest mirrors what `cargo check` will
         # link against transitively from the Tauri runtime.
-        uses: awalsh128/cache-apt-pkgs-action@553a35bb8ebd9fcabcb1c9451aa4c98e1b4ca8a9 # v1.6.3
-        with:
-          packages: >-
-            libgtk-3-dev libwebkit2gtk-4.1-dev libsoup-3.0-dev
-            libayatana-appindicator3-dev librsvg2-dev libasound2-dev
+        #
+        # Installed directly rather than via cache-apt-pkgs-action on
+        # purpose: the runner image's pre-baked apt index pins package
+        # versions (e.g. gstreamer1.0-plugins-good, a libwebkit2gtk dep)
+        # that the Ubuntu mirror rotates out, so a cached install with no
+        # `apt-get update` 404s. The action also keys its cache purely on
+        # the package list, so a failed run saves an empty cache that
+        # poisons every later run under the same immutable key. `apt-get
+        # update` first refreshes the index to whatever the mirror
+        # currently serves; skipping the apt cache removes the poison
+        # trap. The expensive cache — compiled Rust artifacts — stays on
+        # Swatinem/rust-cache below.
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y --no-install-recommends \
+            libgtk-3-dev libwebkit2gtk-4.1-dev libsoup-3.0-dev \
+            libayatana-appindicator3-dev librsvg2-dev libasound2-dev \
             libssl-dev pkg-config
-          version: v1
 
       - uses: dtolnay/rust-toolchain@2c7215f132e9ebf062739d9130488b56d53c060c # stable
         with:
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -88,14 +88,19 @@ jobs:
       # proc macros (#[tauri::command], sqlx::query!, etc.) and external
       # crate symbols. Without it, analysis quality drops below the 50 %
       # call-target threshold and GitHub flags the run as low-quality.
-      - name: Cache APT packages
-        uses: awalsh128/cache-apt-pkgs-action@553a35bb8ebd9fcabcb1c9451aa4c98e1b4ca8a9 # v1.6.3
-        with:
-          packages: >-
-            libgtk-3-dev libwebkit2gtk-4.1-dev libsoup-3.0-dev
-            libayatana-appindicator3-dev librsvg2-dev libasound2-dev
+      # Installed directly rather than via cache-apt-pkgs-action: the
+      # runner image's pre-baked apt index pins package versions the
+      # Ubuntu mirror rotates out, so a cached install with no `apt-get
+      # update` 404s, and the action's package-list-keyed cache saves an
+      # empty entry on failure that poisons later runs. Same rationale as
+      # ci.yml.
+      - name: Install Linux build dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y --no-install-recommends \
+            libgtk-3-dev libwebkit2gtk-4.1-dev libsoup-3.0-dev \
+            libayatana-appindicator3-dev librsvg2-dev libasound2-dev \
             libssl-dev pkg-config
-          version: v1
 
       - uses: dtolnay/rust-toolchain@2c7215f132e9ebf062739d9130488b56d53c060c # stable
         with:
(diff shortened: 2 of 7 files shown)


InstaZDLL:
@coderabbitai review

InstaZDLL:
@coderabbitai review

PR #462, opened 2026-07-30
feat(library): grid tab for user playlists

### LINKED ISSUE #461: feat: Playlists (the ones we make) grid view page .
Hi @InstaZDLL ,
can you consider to add a new page with grid view (similer to albums view ) for the playlists .

i never like the Spotify style of display the playlist .. for us the users that like to make our on playlists , they are kind of albums , it is much more user freindly to use grid view with side scroollbar , to view all of them together ... then just have a link/button long list on left side panel .... view one by one .

i see in a few others music players App , that use same style of spotify , slowly start to add this feature , as users do ask for it ..  

thank you.

InstaZDLL:
Closes #461 — @jo-el414 asked for a grid view of the playlists *we make*, arguing that someone who curates their own playlists reads them as albums, and that a one-line-at-a-time sidebar list means viewing them "one by one".

## Approach

A sixth **My music** tab, next to Tracks / Albums / Artists / Genres / Folders.

**No backend work was needed.** `PlaylistContext` already carries everything a grid wants — `cover_path`, `track_count`, `total_duration_ms`, `position`, `updated_at` — because the sidebar renders the same rows. So the tab issues no query, has no loading state and no skeleton, and the navigation reuses the existing `navigateToPlaylist` / `activePlaylistId` path.

**User playlists only** (`is_smart === 0`), which is what the issue asked for and what the codebase already does: `HomeView` filters `is_smart === 1` into its "Made for you" carousel, with a comment noting user playlists keep their dedicated list.

## Details worth knowing

- **Row-virtualized, not a plain CSS grid.** It mirrors the albums grid: `usePageScroll` + `@tanstack/react-virtual`, so the app stays on its single Spotify-style scrollbar and a profile with hundreds of playlists doesn't mount that many covers on a tab switch.
- **Sort persists.** Goes through the shared `SortDropdown` + `useSortMemory`, so it behaves like every other tab. Default is `custom` = the sidebar's manual `position` order, so the grid opens in the arrangement the user already curated rather than reshuffling it.
- **No cover → the sidebar's own icon + colour tile**, via the existing `resolvePlaylistColor` / `PlaylistIcon`, so one playlist looks identical in both places.
- **Each tile is a real `<button>`** — tabbable and Enter/Space-activatable, rather than a `div` with an `onClick` needing a keydown handler bolted on.
- Sorting works on a copy: `playlists` comes from the context and is shared with the sidebar, which renders it in `position` order.

## Scope

Click-to-open only, no per-tile context menu. Rename / delete / change cover already exist in the sidebar and in `PlaylistView`, so nothing is lost, and it keeps this first pass small — worth revisiting once there's real usage.

## Verification

`bun run typecheck`, `bun run lint`, prettier — clean (exit codes read directly). I also cross-checked every `t()` key this adds against all 17 locale files: none missing, with the CLDR plural categories Russian and Arabic need for the two counted strings (a missing `_other` on ru is what silently falls back to English).

**Not visually verified** — the app doesn't run outside the Tauri shell here, so tile metrics and the virtualized scroll are reasoned from the albums grid they copy. Worth a look at a few window widths, and with a playlist that has no cover.

## Summary by CodeRabbit

  * Ajout d’un onglet **Playlists** dans la bibliothèque.
  * Affichage des playlists personnelles sous forme de grille, avec couvertures, métadonnées et état vide.
  * Navigation vers les détails d’une playlist depuis la grille.
  * Ajout du tri par nom, nombre de titres, durée, date de modification ou ordre personnalisé.

* **Documentation**
  * Documentation mise à jour pour préciser l’accès aux playlists depuis la bibliothèque et la barre latérale.

* **Localisation**
  * Nouvelles traductions pour l’onglet, les compteurs, la grille et les options de tri dans les langues disponibles.


InstaZDLL:
@coderabbitai Bien vu, corrigé en 83ea6b2 — et le défaut avait un second volet que la remarque anticipe justement.

Non seulement `PlaylistGrid` ne se montait jamais à vide (son `EmptyState` était du code mort), mais la branche générique est fausse deux fois pour cet onglet : elle lit `library.empty.<tab>.*`, absent ici, donc i18next aurait peint la clé brute à l'écran — j'ai vérifié, `library.empty.playlists` n'existe dans aucune des 17 locales — et elle propose « Importer un dossier », ce qui ne crée pas de playlist.

J'ai pris la première option : l'onglet est rendu même vide, la grille garde son propre message, **aucune clé générique ajoutée**. La raison est écrite à côté de la condition pour que quelqu'un ne la resserre pas par symétrie plus tard.

J'ai aussi masqué le dropdown de tri quand la grille est vide : proposer un tri sur rien était le petit cousin de la même erreur.

Issue #443, opened 2026-07-26
feat: Release Radar plugin + waveflow:ui/v1 plugin UI surface


InstaZDLL:
## Goal

**Release Radar** — surface new & upcoming releases from the artists already in your library, Spotify-style. To be built **from scratch as a clean plugin**; the WIP attempt (#312) is closed in favour of this.

## Why a rebuild

#312 landed the feature and the plugin-UI plumbing together as a WIP. Rather than untangle it, we redo it cleanly: a small, well-scoped `ui`-world plugin surface in the core, and the Release Radar itself as a separate plugin in the plugins repo — consistent with how Web Radio and Apple Motion Artwork are shipped.

## Part 1 — `waveflow:ui/v1` plugin world (this repo)

A UI-world lets a plugin render its own view inside WaveFlow without shipping React:

- **JSON view descriptor** — the plugin returns a declarative view (sections, cards, lists, images, buttons); a **generic React renderer** in the app draws it. No arbitrary DOM/JS from the guest.
- **User actions** — the renderer sends typed action events (button click, item open) back to the plugin; the plugin can **update its content live** (re-emit a descriptor).
- **Sidebar entry** — a UI plugin can register a navigable entry (icon + label) in the sidebar.
- **New permission-gated host import `library.read_artists`** — a **redacted** read (artist names / ids only, not a full library dump), permission-gated + surfaced in the plugin's "Can reach / uses" chips like the existing capabilities. Honours `offline::is_offline()` where it fetches.

Design constraints: reuse the existing manifest permission-gate model ([plugins guide](docs/features/plugins.md)); the renderer must be safe by construction (declarative descriptor, no raw HTML); keep the guest read-only against the library.

## Part 2 — Release Radar plugin (separate repo, waveflow-plugins)

- Reads the library's artists via the gated import, queries **MusicBrainz** + **Cover Art Archive** for recent / upcoming releases, and renders them through the UI descriptor (grid of release cards with cover, title, date, artist).
- **Explicitly out of scope** (as in #312): no YouTube downloading, no audio extraction, no hidden playback. Release cards open **external search / listen links** only.
- Permission-gated network (MusicBrainz + Cover Art Archive hosts allowlisted in its manifest), offline-aware.

## Acceptance sketch

- A UI plugin can declare the `ui` world + `library.read_artists`, register a sidebar entry, render a descriptor, and handle actions + live updates.
[...]

PR #312, opened 2026-06-24
WIP: add release radar plugin

InstaZDLL:
## Summary

- adds the `waveflow:ui/v1` plugin invocation surface and a redacted `library.read_artists` host permission
- bundles a new Release Radar UI plugin powered by MusicBrainz and Cover Art Archive
- adds a generic React renderer for plugin-provided JSON view descriptors and a sidebar entry for Release Radar

## Notes

This intentionally does not add YouTube downloading, audio extraction, or hidden YouTube playback. Release Radar opens external search/listening links only.

## Validation

- `cargo component build --release`
- `cargo check --manifest-path src-tauri/Cargo.toml -p waveflow --all-targets`
- `bun run typecheck`
- `bun run lint`
- `cargo test --manifest-path src-tauri/Cargo.toml -p waveflow-core --test plugin_release_radar --features plugins,sqlite`
- `cargo test --manifest-path src-tauri/Cargo.toml -p waveflow-core plugin::runtime::tests::build_linker_succeeds --features plugins,sqlite`

## Summary by CodeRabbit

* **New Features**
  * Ajout d’une nouvelle vue **Release Radar** dans l’application.
  * Les plugins peuvent désormais afficher une interface dédiée, recevoir des actions utilisateur et mettre à jour leur contenu en temps réel.
  * Le plugin **Release Radar** est désormais disponible avec un affichage enrichi et des actions rapides.

* **Bug Fixes**
  * Amélioration de la prise en charge des permissions affichées dans les paramètres des plugins.
  * Meilleure gestion de l’accès à la bibliothèque pour certains plugins.

PR #474, opened 2026-08-01
feat(plugins): waveflow:ui/v1 world host surface + redacted library read (#443 Part 1)

InstaZDLL:
Part 1 of #443 (Release Radar + UI plugin surface), **backend-first**. A clean rebuild of the closed WIP #312, fixing its two biggest smells: the sidebar is driven off `manifest()` (not hardcoded to one plugin), and the `ui` world reuses `source`'s host types via bindgen `with:` (not a duplicated impl set).

The frontend generic renderer + dynamic sidebar land in a **follow-up PR** — this one is self-contained and fully covered by an integration test.

## What a `ui`-world plugin is

It renders a custom view **without shipping any React**. The guest returns a **JSON view descriptor** (sections / cards / images / action buttons) the host draws with native components — never HTML/CSS/JS/React, so a hostile descriptor can't run code in the app's origin. Exported `extension` interface:

- `manifest() -> mount-point` — sidebar registration (label + optional lucide icon + initial path).
- `render(path) -> result<string, string>` — the current view as a descriptor string.
- `on-event(event, payload) -> result<string, string>` — a user action round-trips here; the plugin returns the **next full descriptor** (no diff protocol). `open-url` actions are handled host-side and never reach the guest.

## The redacted `library.read_artists` capability

A new host import, `library.list-artists(limit) -> [{id, name, track-count}]` — **names + aggregate counts + an opaque id only**, no file paths or per-track rows. Permission-gated (`library.read_artists`), doubly clamped (`MAX_LIBRARY_ARTISTS`), and **snapshot-injected**: the host queries the active profile async-side and hands the guest a ready list, so the guest never touches SQLite (local-only ⇒ works offline). The redaction is host-enforced — a plugin without the permission gets `Err("permission denied: library.read_artists")` even when the snapshot is present.

## Changes

- **WIT/SDK**: promote the `ui/v1` skeleton to the manifest/render/on-event triad; align the `ui` host.wit with `source` (adds `config`) + the new `library` interface; add the `library.read_artists` SDK permission.
- **Core**: `ui` bindgen (`with:` reuse); `library::Host` on `HostCtx` (gated + clamped + snapshot); manifest gate; `runtime::{ui_manifest, ui_render, ui_event}`.
- **App**: `list_ui_plugins` / `plugin_ui_render` / `plugin_ui_event` + an async redacted-snapshot loader; registered in `lib.rs`.
- **Test**: a minimal, never-bundled `plugins/ui-fixture/` component + `tests/plugin_ui.rs` proving the round-trip and **both** sides of the permission gate against the same wasm.
- **Docs**: `plugins.md` UI-world section + `CLAUDE.md` clause.

## Validation

- `cargo clippy -p waveflow-core --features plugins,sqlite --all-targets` — clean
- `cargo clippy -p waveflow --lib` — clean
- `cargo test -p waveflow-core --features plugins,sqlite` — 193 passed (188 unit + 4 new ui + 1 web-radio)

Relates to #443. Follow-up: frontend renderer + dynamic sidebar.

## Summary by CodeRabbit

  * Prise en charge des plugins d’interface utilisateur avec vues JSON déclaratives, navigation latérale dynamique et gestion des événements.
  * Ajout d’un accès contrôlé à une liste limitée et anonymisée d’artistes.
  * Les actions d’ouverture d’URL sont gérées par l’application hôte, tandis que les autres actions peuvent mettre à jour la vue.
  * Les erreurs de configuration des plugins sont ignorées afin de préserver l’affichage des autres extensions.

* **Documentation**
  * Documentation mise à jour sur le développement et les permissions des plugins UI.

* **Tests**
  * Tests ajoutés pour le rendu, les événements, les limites et les autorisations.


InstaZDLL:
@coderabbitai addressed in 12bba19a:

- **`load_library_artist_snapshot` — gate on permission before the query** ✅ Added `plugin_grants_library_read` (manifest read, id-pinned, default-closed); the snapshot loader now short-circuits to an empty list before acquiring the profile pool or running the query when the plugin didn't declare `library.read_artists`. Authorized plugins keep the capped query, so both callers inherit it.

- **Validate `plugin_ui_render` / `plugin_ui_event` output as JSON** ✅ Added `validate_ui_descriptor`; both commands now parse the guest's output with `serde_json` before returning, surfacing a malformed descriptor as a clean plugin-tagged `AppError` instead of a raw frontend `JSON.parse` throw. The `spawn_blocking` + success flow is unchanged; the frontend still validates `schemaVersion` itself.

- **Fixture exercises host-side clamping** ✅ `ui-fixture` now requests `u32::MAX` artists, and a new `ui_fixture_render_clamps_snapshot_to_host_max` test feeds a snapshot larger than the cap and asserts the guest sees exactly `MAX_LIBRARY_ARTISTS`. Fixture wasm rebuilt + refreshed.

- **`wit-bindgen-rt` 0.44 vs bindings generated by 0.41** — skipped: this mirrors the shipped `web-radio` plugin exactly. `cargo-component` bundles its own 0.41 generator while the runtime crate is `wit-bindgen-rt = "0.44"`; the two are compatible (build + tests green), and changing it would diverge the fixture from the one bundled plugin we already ship this way.

Validation: `cargo test -p waveflow-core --test plugin_ui` (5 passed), `cargo clippy -p waveflow --lib` clean.

InstaZDLL:
@coderabbitai addressed in 241e46d6: `load_library_artist_snapshot` now clamps `track_count` with `.clamp(0, u32::MAX as i64) as u32` instead of a bare `.max(0) as u32`, so a value above `u32::MAX` saturates instead of wrapping. Unreachable in practice (it's a track `COUNT`), but the narrowing is now lossless; negatives still floor to 0 via the clamp's lower bound. clippy clean.

PR #475, opened 2026-08-01
feat(plugins): render ui-world plugin views + dynamic sidebar (#443 Part 2)

InstaZDLL:
Frontend half of the `waveflow:ui/v1` surface. **Stacked on #474** (backend) — base is the backend branch so this diff is frontend-only; GitHub retargets it to `main` once #474 merges.

Consumes the `list_ui_plugins` / `plugin_ui_render` / `plugin_ui_event` commands to draw a ui plugin's JSON view descriptor with native React — the guest ships no UI code.

## What's here

- **`PluginUIView`** — generic descriptor renderer (header + sections + item cards + badges + action buttons). `open-url` actions open in the OS browser via `openUrl`; `event` actions round-trip through the plugin and swap in the next descriptor. Respects section titles, bakes in no per-plugin assumptions. A non-destructive error banner keeps the last good view.
- **Dynamic sidebar** — `useUiPlugins` → `list_ui_plugins` renders one nav entry per enabled ui plugin; label + icon + landing path all come from each plugin's `manifest()` (no hardcoded per-plugin wiring). Icon resolved from a curated lucide allowlist (`pluginIcons`); unknown/absent → generic Puzzle glyph.
- **Routing** — a `plugin-ui` `HistoryEntry` carries the plugin id; `AppLayout` keys the view remount on it so switching plugins reloads cleanly.
- **`plugins.ts`** — descriptor types (single source) + `parsePluginUiDescriptor` (validates `schemaVersion`) + wrappers; `libraryReadArtists` on `PluginPermissionsInfo`.
- **`PluginsCard`** — a "reads your artists" permission chip for `library.read_artists`.
- **i18n** — `pluginView.*` + `settings.plugins.permissions.libraryReadArtists` ×17.

## Validation

- `bun run typecheck` — clean
- `bun run lint` — clean

Relates to #443. Backend: #474.

## Summary by CodeRabbit

  - Ajout d’interfaces dédiées pour les plugins, accessibles depuis la barre latérale.
  - Prise en charge du contenu dynamique : sections, images, badges, actions et liens externes.
  - Actualisation, gestion des erreurs, nouvelle tentative et affichage des vues vides.
  - Ajout d’une permission permettant aux plugins de lire les artistes de la bibliothèque.
  - Icônes de plugins prises en charge avec une icône de remplacement.

- **Traductions**
  - Ajout des libellés liés aux interfaces de plugins et aux permissions dans les langues disponibles.


InstaZDLL:
@coderabbitai addressed in 5cf93bed:

- **Recursive descriptor validation** ✅ `parsePluginUiDescriptor` now walks the whole tree beyond the `schemaVersion` gate — root object + string `title`, header `actions`, each `section` (object + array `items`), each `item` (string `id`/`title`, array `badges`/`actions`), and each `action` (known `kind` discriminator + the field that kind requires: `label`, plus `url` for open-url / `event` for event). It throws on the first malformed node, so `PluginUIView` never receives a structure that could throw mid-render (a non-array `sections`/`items`/`badges` would break `.map`; an unknown `kind` would misroute). The throw surfaces as the existing error banner.

- **Refresh + retry loading states** ✅ Both now take `disabled={isLoading}` and swap to a spinning `Loader2` while loading, matching the `busyAction` behaviour in `renderAction`. Labels/icons/`load` calls unchanged when idle.

- **Hindi grammar** ✅ `libraryReadArtists` → `आपके कलाकारों को पढ़ता है`.

Validation: `bun run typecheck` + `bun run lint` clean.

InstaZDLL:
@coderabbitai addressed in 5f4e3ec5:

- **Separate flow tokens** ✅ `load` and `runAction` now use distinct request refs (`loadReqRef` / `eventReqRef`), so a load and an event can't invalidate each other's completion (which could previously leave `isLoading` / `busyAction` stuck true). Action buttons + refresh/retry mutually disable while any op is in flight (`isLoading || busyAction !== null`) so competing flows can't race.
- **Clear error on new action + surface open-url failure** ✅ `runAction` clears the prior error at the start of every action; the `open-url` catch now calls `setError(String(e))` in addition to `console.error`, so a failed `openUrl` shows in the banner.
- **Error banner a11y** ✅ `role="alert"` added.
- **Empty-state condition** ✅ The empty view now renders only when a `descriptor` exists with no items; a null `descriptor` after a failed initial load leaves the error banner as the visible state instead of a misleading "nothing to show".
- **Text-field validation** ✅ `parsePluginUiDescriptor` now type-checks the directly-rendered optional text fields too — `subtitle` / `emptyTitle` / `emptyHint` (root), `section.title`, item `subtitle` / `detail`, and each `badges` element — so a non-string can't make React throw mid-render.

Validation: `bun run typecheck` + `bun run lint` clean.

InstaZDLL:
@coderabbitai addressed in 0e63f31b:

- **Busy-gate open-url** ✅ The open-url branch now sets `busyAction` before starting `openUrl` (blocking other actions + refresh while it's pending, same as the event path) and clears it in both the success and failure handlers. Error logging + `setError` on failure preserved.
- **Reject null badge elements** ✅ `validateUiItem` now requires each `badges` element to be a `string` (throws otherwise, with the indexed context) — badge elements aren't optional the way `subtitle` / `detail` are, and a null would render an empty pill. `assertOptionalString` stays for the genuinely optional fields.

Validation: `bun run typecheck` + `bun run lint` clean.

PR #480, opened 2026-08-02
feat(plugins): support direct download_url in registry entries

InstaZDLL:
Adds an optional `download_url` to a registry entry. When present, `install_plugin_from_registry` downloads the release asset from **that URL** instead of the GitHub `releases/download` URL built from `repo`.

This decouples the store's install path from GitHub, so the app-controlled registry endpoint (`waveflow.app/api/plugins/registry`) can host a plugin's binary itself — enabling a **closed-source, binary-only** plugin with no public source repo (the first consumer being the Spotify Canvas plugin, whose source stays private).

## What changes

- **`RegistryEntry.download_url: Option<String>`** (serde `default`, so every existing entry decodes unchanged; older/GitHub-listed plugins simply omit it).
- **`install_plugin_from_registry`**: if `download_url` is set, use it (required to be `https` — no downgrade / loopback target); otherwise the existing `github.com/{repo}/releases/download/v{version}/{asset}` URL. The downloaded bytes are still **blake3-verified against the registry pin** either way, so the trust model is unchanged — the registry remains the single source of truth, `download_url` just moves *where* the (still-verified) bytes come from.

No schema change to the public `waveflow-plugins` registry is needed: entries carrying `download_url` are injected by the app-controlled endpoint, not committed to the GitHub registry (they never pass its CI). Older app builds ignore the field and fall back to the GitHub URL.

## Validation

- `cargo check -p waveflow` + `cargo clippy -p waveflow --all-targets` — clean

## Summary by CodeRabbit

  - L’installation des plugins prend désormais en charge une URL de téléchargement directe lorsqu’elle est fournie.
  - En l’absence d’URL directe sûre, le téléchargement continue d’utiliser automatiquement la source GitHub configurée.

- **Sécurité**
  - Les téléchargements directs sont limités aux adresses HTTPS autorisées.
  - Les redirections sont contrôlées et validées à chaque étape afin de bloquer les destinations non sûres.
  - La vérification d’intégrité BLAKE3 et les contrôles d’installation restent appliqués à tous les plugins.

### DIFF
--- a/src-tauri/crates/app/src/commands/plugin_store.rs
+++ b/src-tauri/crates/app/src/commands/plugin_store.rs
@@ -107,6 +107,15 @@ struct RegistryEntry {
     blake3: String,
     #[serde(default)]
     asset: Option<String>,
+    /// Optional direct download URL for the release asset. When present it
+    /// OVERRIDES the GitHub `releases/download` URL built from `repo` — this
+    /// lets the app-controlled registry endpoint host a plugin's binary itself
+    /// (e.g. on `waveflow.app`) instead of a public GitHub release, so a
+    /// closed-source plugin can be distributed binary-only with no public repo.
+    /// Trusted like the rest of the entry (the registry is the trust anchor)
+    /// and still blake3-verified after download; required to be `https`.
+    #[serde(default)]
+    download_url: Option<String>,
     permissions: RegistryPermissions,
     #[serde(default)]
     tags: Vec<String>,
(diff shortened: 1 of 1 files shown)


InstaZDLL:
@coderabbitai hardened — pushed.

The `download_url` was only prefix-checked for `https://`. Now the install download reuses the shared **`is_safe_motion_url`** guard (from `waveflow_core::artwork::motion_cache`) for both the URL and every redirect hop:

- **`download_url` validation**: replaced the `starts_with("https://")` check with `is_safe_motion_url(direct)` — requires https AND rejects `localhost` / loopback / private / link-local / unspecified hosts, including bracketed IPv6 literals, IPv4-mapped/compatible IPv6, and userinfo (`user@host`) forms pointing at them.
- **Redirect validation**: the download `reqwest::Client` now uses `redirect::Policy::custom` that re-runs `is_safe_motion_url` on **each hop** (capped at 10, matching the default) — a GitHub release URL still follows its 302 to `objects.githubusercontent.com` (public → allowed), but a redirect to an internal target is refused. This covers HTTP→internal redirects for both the `download_url` and the existing GitHub path.

Even though the URL is registry-sourced (the trust anchor) and the bytes are blake3-verified, this stops a compromised entry from using the app as an SSRF probe.

**Skipped** the "resolve DNS and reject internal resolutions" part: same as the motion-cache guard, DNS-rebinding is a documented accepted residual ("a hostname that resolves to an internal IP at connect time … is a residual we accept — plugins are curated and user-installed"). Adding per-hop DNS resolution is out of scope and a deliberate existing trade-off.

On tests: the validation logic IS `is_safe_motion_url`, which is exhaustively unit-tested in `motion_cache` (localhost, loopback, private/link-local v4+v6, IPv4-mapped, userinfo, non-https). The install download path itself is network-bound with no unit-test harness, so I didn't add a live-network test there.

Validation: `cargo clippy -p waveflow --all-targets` clean.

InstaZDLL:
@coderabbitai skipping the custom-`Resolve` DNS-rebinding hardening, with reasoning.

The download URL comes from the **registry — the trust anchor**. An attacker who can set `download_url` has already compromised the catalogue, at which point DNS rebinding buys them nothing extra: they control which plugins are offered. The only marginal capability rebinding adds is a **blind SSRF GET to an internal address**, and the response is useless — it must match the registry-pinned **blake3** to install anything, which an internal service won't produce. So the residual is "an authenticated-by-registry-compromise blind GET", not a code-exec or data-exfil path.

This matches the existing, documented stance of the shared guard: `is_safe_motion_url`'s doc explicitly accepts DNS rebinding as a residual ("a hostname that resolves to an internal IP at connect time … is a residual we accept — plugins are curated and user-installed"). Adding a custom `reqwest::dns::Resolve` that resolves every hop + rejects internal `SocketAddr`s (plus a mock-resolver test) is a disproportionate addition for that threat model, and would diverge this one call site from the guard's consistent policy.

What IS in place after this PR: https-only + literal loopback/private/link-local/unspecified rejection (incl. IPv6-mapped + userinfo forms) on the initial URL **and** re-validated on every redirect hop. That covers the practical SSRF surface for a curated, blake3-verified install path. If we later decide to add resolve-time IP filtering, it should land once in `is_safe_motion_url`'s layer for all callers, not bolted onto this one client.

PR #483, opened 2026-08-03
feat(artist): spotify-style hero backdrop on the artist page (#482)

### LINKED ISSUE #482: feat: Spotify-style hero background on artist pages (TheAudioDB fanart + blurred-photo fallback)
## Goal

Give the artist detail page a **full-bleed hero background** behind the header, Spotify-style (cf. the wide artist banner on Spotify's artist page) — instead of today's flat surface with just a circular avatar + name.

**Backdrop precedence** (highest wins), decided: **1. real wide fanart (TheAudioDB) → 2. blurred existing square photo → 3. colour gradient (nothing at all)**. Ship both tiers in one PR (fanart pipeline + blurred fallback).

## Why two tiers

- A crisp **wide fanart** gives the real Spotify look, but not every artist has one.
- The **blurred square photo** (Deezer / local `artist.jpg`) is always available and works offline — the universal fallback, using the existing ambient-backdrop technique.

## Current state (grounded)

The data pipeline carries **only the square artist photo** (`picture_*`) end to end — there is **no** fanart/background/banner field anywhere yet. Key sites:

- **Header** — [`src/components/views/ArtistDetailView.tsx`](src/components/views/ArtistDetailView.tsx): root `div` at L279 (`space-y-8 animate-fade-in pb-20`), header row at L281 (`flex items-center space-x-8`), circular photo `w-48 h-48 rounded-full` L283–319, name/stats/actions column L321–375. **No internal max-width wrapper** → a hero can break full-bleed here.
- **TheAudioDB client** — [`src-tauri/crates/core/src/metadata/theaudiodb.rs`](src-tauri/crates/core/src/metadata/theaudiodb.rs): `ArtistPayload` (L35–59) parses **bio fields only**. `search.php` already returns `strArtistFanart` / `strArtistFanart2/3/4` / `strArtistWideThumb` / `strArtistBanner` — add them here.
- **Enrichment DTO** — [`src-tauri/crates/app/src/commands/deezer.rs`](src-tauri/crates/app/src/commands/deezer.rs) `DeezerArtistEnrichment` (L251–266, plain snake_case Serialize) + TS mirror in [`src/lib/tauri/detail.ts`](src/lib/tauri/detail.ts) (L122–133). Also the richer `ArtistDetail` DTO (`getArtistDetail`, detail.ts L71–87) feeds the header.
- **Cache table** — `metadata_artist` in the shared `app.db` (keyed by Deezer id). Follows a `picture_url` + `picture_hash` pair; add `background_url` + `background_hash` the same way. Migrations under [`src-tauri/migrations/app/`](src-tauri/migrations/app/) (started as `deezer_artist`, renamed `metadata_artist`).
- **Artwork cache pattern** — [`src-tauri/crates/core/src/artwork/metadata.rs`](src-tauri/crates/core/src/artwork/metadata.rs): `metadata_artwork::download_and_cache(url, dir)` → blake3 hash → `existing_path(dir, h)`, thumbnail tiers via `thumbnails::thumbnail_paths_for`. Reuse verbatim for the background image.
- **Blurred-backdrop pattern to reuse** — [`src/components/layout/SkinAmbientBackdrop.tsx`](src/components/layout/SkinAmbientBackdrop.tsx): `bg-center bg-cover` + `filter: blur(...) saturate(...)` + `transform: scale(1.6)` + vignette gradient + `key={resolved}` remount cross-fade + `prefers-reduced-motion` gate + `convertFileSrc` for local asset paths. Cross-fade keyframe in [`src/styles/skins/lounge.css`](src/styles/skins/lounge.css) L504–512.

## Scope — backend

- **Parse fanart in `theaudiodb.rs`**: add `strArtistFanart` (+ optionally `strArtistWideThumb` / `strArtistFanart2/3/4` as ordered fallbacks) to `ArtistPayload`; surface a `fanart_url: Option<String>` on the public return type. Honour [`offline::is_offline()`](src-tauri/crates/app/src/offline.rs).
- **New `app` migration**: `metadata_artist` gains `background_url TEXT` + `background_hash TEXT` (mirror `picture_url`/`picture_hash`). New dated file — migrations are immutable once merged; do **not** edit an existing one.
- **`enrich_artist_deezer`**: resolve the fanart URL (TheAudioDB), `download_and_cache` it into `metadata_artwork/<blake3>.jpg`, persist `background_hash`, return `background_path`/`background_url` (and optional `_1x`/`_2x` tiers via `thumbnail_paths_for` — a hero is large, full-res is probably fine, tiers optional).
- **DTOs**: add the background field(s) to `DeezerArtistEnrichment` (Rust) **and** the TS interface, plus `ArtistDetail` if the header should get it directly from `getArtistDetail`.
- **Open question**: the fanart fetch should be **independent of the `metadata.bio_source` setting** (fetch the fanart from TheAudioDB even when the bio comes from Last.fm), or gate it — decide during impl. TheAudioDB free key rate-limits, so cache aggressively (the 30-day `metadata_artist` TTL already applies).

## Scope — frontend

- **Hero layer** in `ArtistDetailView`: a full-bleed absolutely-positioned background behind the header (break out at root L279 / header L281). Reuse the `SkinAmbientBackdrop` approach with a **lighter blur (or none) for real fanart** and a **stronger blur for the square-photo fallback**. `convertFileSrc` for local cache paths; `key`-remount cross-fade; `prefers-reduced-motion` gate.
- **Legibility**: force a **dark bottom gradient overlay + white header text** (eyebrow/name/stats) so it reads in **both light and dark themes** — Spotify's header is always dark-on-image. Reconcile with the current `text-zinc-900 dark:text-white` name.
- **Precedence**: real fanart (`background_path`) → blurred square photo (`picture_*`/local `artwork_path`) → existing flat surface (no image).
- **Per-profile toggle** Settings → Appearance, `ui.artist_hero` (**default ON** — it's a baseline visual, not extra motion), following the [`useCoverSlideshow`](src/hooks/useCoverSlideshow.ts) / [`usePlayerBarLayout`](src/hooks/usePlayerBarLayout.ts) pattern (serialized writes, profile guard, refresh event). Consider honouring `prefers-reduced-motion` only for the cross-fade, not the static image.
- **i18n**: propagate any new strings (toggle label/subtitle) to all **17 locales** (fr source of truth).

## Acceptance

- Artist page shows a crisp wide fanart hero when TheAudioDB has one; a blurred photo hero otherwise; the flat surface when the artist has no image at all.
- Header text stays legible over the hero in every theme × skin.
- Toggle in Settings → Appearance turns it off (falls back to today's flat header). Offline: no network fetch, blurred-photo tier still works.
- `bun run typecheck` + `bun run lint` + `cargo check --workspace --all-targets` + `cargo clippy` clean; new migration validated against a real DB.
- Docs: `docs/features/library.md` (artist detail) + CLAUDE.md feature catalogue updated.

## Reference

Idea + reference look: Spotify artist page hero banner (user-provided screenshots). Related existing backdrops: `SkinAmbientBackdrop` (Lounge), cover slideshow (#466/#470), track Canvas (#442/#473) — all share the blur/gradient/cross-fade/reduced-motion vocabulary.

InstaZDLL:
Closes #482.

Gives the artist detail page a full-bleed hero behind the header, Spotify-style, instead of today's flat surface with a circular avatar. Both tiers ship here: the real wide fanart pipeline **and** the blurred-photo fallback.

## Backdrop precedence

**real wide fanart (TheAudioDB) → blurred square photo (Deezer / local `artist.jpg`) → flat header**

A crisp wide fanart gives the real look but not every artist has one; the blurred square photo is always available and works offline, so it's the universal fallback.

## Backend

- **`theaudiodb.rs`** — `ArtistPayload` now parses `strArtistFanart` (+ `2/3/4`), `strArtistWideThumb`, `strArtistBanner`. `artist_bio` becomes **`artist_info`** and returns bio *and* `fanart_url` from a single lookup, picking widest-and-cleanest first (logo banner last, its baked-in text can clash with the header copy). It returns `Some` for any name match even with neither bio nor fanart, so the caller can cache the "looked, nothing there" outcome.
- **New app migration** `20260802120000_metadata_artist_background.sql` — `background_url` + `background_hash` + **`background_fetched_at`**. That last column is the "we already looked" marker: without it a NULL hash is indistinguishable from "never queried", and every fanart-less artist would re-hit a rate-limited API on each page visit. It's stamped only when the API was *reached* (match or not) and left NULL on a transport error, so a network blip retries instead of caching "no fanart" for the row's whole 30-day TTL. Pre-existing rows have NULL there and backfill once on their next refresh.
- **`enrich_artist_deezer`** — resolves the fanart **independently of `metadata.bio_source`** (the issue's open question): Last.fm has no equivalent image, so gating on the bio source would leave every Last.fm user with no hero at all. One TheAudioDB request serves both consumers — the bio half is used only when TheAudioDB *is* the selected source, which matters on their shared rate-limited free key. Downloaded through the usual `metadata_artwork::download_and_cache` and kept **full-res** (no `_1x`/`_2x` — downscaling a full-bleed banner only softens it). Offline mode short-circuits before all of it.
- **`get_artist_detail`** returns `background_url` / `background_path` too, so the hero paints on the first frame for an already-enriched artist instead of waiting for the enrichment round-trip.

## Frontend

- **`ArtistHeroBackdrop`** — fanart shown nearly crisp (`blur(2px)`, just enough to keep JPEG artefacts from crawling under the white copy); the square photo gets the heavy colour-field treatment `SkinAmbientBackdrop` uses (`blur(56px) saturate(190%)`, `scale(1.35)`) because a 1:1 image stretched across a banner is unreadable otherwise.
- **Legibility** is not left to the theme: dark scrim + header copy (eyebrow / name / stats) forced **white in every theme**, matching Spotify, whose artist header is dark-on-image in light mode too. Secondary buttons swap to a translucent white treatment. The bottom edge fades through a CSS **mask** rather than a hard-coded colour stop, so the hero dissolves into whatever the current theme × skin paints behind it.
- **`ArtistDetailView`** breaks out of `<main>`'s `p-8` with `-mx-8 -mt-8` so the image reaches the column edges (and shrinks with the column when a right panel opens). The enrichment response only ever *sets* the fanart — an empty refresh (offline, TheAudioDB down) must not blank a hero the cache already produced.
- **Toggle** `ui.artist_hero`, per-profile, **default ON** (a baseline visual, not extra motion), Settings → Appearance; write machinery mirrors `useCoverSlideshow`. `prefers-reduced-motion` skips the cross-fade only — the image itself is static.
- **i18n** propagated to all 17 locales.

## Docs

`docs/features/ui.md` (new *Artist hero* section), `docs/features/library.md` (*Wide artist fanart (hero)* — the backend pipeline), CLAUDE.md (UI + Integrations catalogues).

## Checks

- `cargo check --workspace --all-targets` ✅ · `cargo clippy` — no new warning (the 7 reported are pre-existing in `scanner/extract.rs` + `browse.rs:462`)
- `cargo test -p waveflow-core` — 102/102 ✅, including 3 new `theaudiodb` tests covering the fanart fallback order
- Migration applied against a real SQLite DB (full `migrations/app/` chain + the exact INSERT the code issues) ✅

**Not verified by me:** the visual result in the running app (crisp fanart vs blurred fallback, legibility across the 5 skins × 14 themes). Worth a look before merge.

https://claude.ai/code/session_01N9WXrurarkoiT2jgimvyj1

## Summary by CodeRabbit

  * Ajout d’un arrière-plan panoramique sur les pages de détail des artistes.
  * Priorité à l’image TheAudioDB, avec repli sur une photo locale ou Deezer floutée.
  * Activation ou désactivation depuis les paramètres d’apparence, avec préférences par profil.
  * Respect de la réduction des animations.

* **Améliorations**
  * Mise en cache en pleine résolution et nouvelles tentatives après les erreurs réseau.
  * Traductions ajoutées dans les langues prises en charge.

* **Documentation**
  * Documentation ajoutée sur le bandeau artiste et ses options.


InstaZDLL:
@coderabbitai Two findings from this review are real but deliberately out of scope for this PR — tracked in #485:

1. **`set_profile_setting` isn't profile-scoped (TOCTOU).** The JS guard runs before the `await`, so a `switch_profile` landing before `require_profile_pool()` acquires the lease writes into the wrong profile. Only the backend can close this, by validating an expected id under the same lock that resolves the pool. The command is shared by 22 call sites across 18 files — fixing it for this one hook would leave the other 21 exposed while looking covered.

2. **The preference-hook pattern is copy-pasted 7×.** This review hardened `useArtistHero` alone (stale-read invalidation, per-profile `resolved` marker). The other six carry the same races; they're invisible there only because those hooks default to OFF, so an unresolved frame paints nothing. The right fix is one shared `useProfileBooleanSetting`, not six more rounds of the same patch.

Everything else from the review is applied on this branch. The remaining suggestions I did not take, with reasons:

- **Read/write barrier in `refresh`:** not reachable. `ARTIST_HERO_EVENT` has a single emitter, gated on a successful write *and* on `seq !== writeSeqRef.current` — so a newer in-flight write suppresses the older one's broadcast, and when an event does fire, the persisted value is already the one being read.
- **Gating the profile stamp on the read sequence:** would regress. The sequence is invalidated *by* a write, so if that write then fails there's no broadcast and no further read — `resolved` would stay false forever and the hero would never appear. The value exposed in that window isn't stale; it's the optimistic click, newer than the read.
- **Broadcasting on write rollback:** nothing to resynchronise. The optimistic update is local to the instance that clicked; other instances only move on the broadcast, which fires on success only. On failure they're already showing the last backend-confirmed value — exactly where `confirmedEnabledRef` rolls the writer back to.

PR #489, opened 2026-08-06
fix(deps): force fast-uri >= 3.1.5 and resync the Flatpak sources

InstaZDLL:
Closes Dependabot alert #17.

## The vulnerability

`fast-uri` < 3.1.5 only recognises a URI authority after a literal `//`, so a reference using `\\`, `/\` or `\/` as the introducer folds the host into the *path*. Node's WHATWG URL (behind `fetch()`, undici, `http`/`https`) treats `\` as interchangeable with `/` for special schemes, so the two parsers extract **different hosts from the same input** — `\\evil.com/path` against `https://allowed.com/` gives `https://allowed.com/%5C%5Cevil.com/path` to one and `https://evil.com/path` to the other. Anything that validates a host with one parser and fetches with the other can be pointed elsewhere.

## The fix

It reaches us transitively — `@commitlint/cli` → `ajv` → `fast-uri ^3.0.1` — so an `overrides` entry is enough. Pinned to `^3.1.5`: patched, and inside the `3.x` line `ajv` asks for (`>=3.1.5` would have allowed 4.x).

The override lives in `package.json`, which is *also* what the Flatpak lockfile is generated from, so **one entry covers both lockfiles**:

- `bun.lock` — carried `fast-uri@3.1.2`, older than the flagged one. Dependabot doesn't parse `bun.lock`, so it never flagged the project's actual lockfile. Now `3.1.5`.
- `packaging/flatpak/generated/package-lock.json` — the one in the alert. Now `3.1.5`.

Editing the generated lockfile by hand would have been undone by the next `generate-sources.sh` run, and would have left `node-sources.json` pointing at the 3.1.4 tarball — so the sources were regenerated with the documented script instead.

## Why the diff is 15k lines

Regenerating surfaced a **separate, pre-existing problem**: `cargo-sources.json` was stale. `chore: update dependencies` (23be643) bumped `Cargo.lock` without re-running `generate-sources.sh`, so **53 crates in the lock had no source entry** — including wasmtime `46.0.1` declared against `47.0.3` in the lock. Flathub builds with the network disabled, so that build could only fail on a missing crate.

Verified programmatically, before and after:

| | crates in `Cargo.lock` with no source | npm packages with no tarball |
|---|---|---|
| before | **53** | 0 |
| after | **0** | 0 |

So the bulk of the diff is the generated sources catching up on dependency drift accumulated since the last regeneration — not fast-uri.

## Checks

- `bunx commitlint` still runs on the overridden `fast-uri` (it's the consumer, via ajv) ✅
- Every crate in `Cargo.lock` has a source entry, every package in the npm lockfile has a tarball ✅

**Not verified by me:** an actual `flatpak-builder` run. The consistency checks above are structural — they prove nothing is missing, not that the build succeeds end to end.

Worth considering separately: the Flatpak sources drifted because nothing enforces regeneration when `Cargo.lock` or `package.json` changes. A CI check that re-runs the generator and fails on a dirty tree would have caught this at 23be643.

https://claude.ai/code/session_01N9WXrurarkoiT2jgimvyj1

## Summary by CodeRabbit

* **Maintenance**
  * Mise à jour de la version utilisée pour une dépendance interne afin d’améliorer la stabilité et la compatibilité de l’application.


InstaZDLL:
@coderabbitai review

PR #496, opened 2026-08-09
feat(audio): native DSD via DoP (DSD over PCM) on WASAPI Exclusive

### LINKED ISSUE #495: feat(audio): DSD natif via DoP (DSD over PCM) en WASAPI Exclusive
## Contexte

WaveFlow décode aujourd'hui les fichiers DSD (`.dsf` / `.dff`) **uniquement en DSD → PCM** : le flux 1-bit est converti en PCM via un FIR Blackman-Harris (taps réglables 256 / 1024 / 2048, `profile_setting['audio.dsd_precision']`), puis envoyé à la sortie comme n'importe quel flux PCM.

- Parseur + convertisseur : [`waveflow_core::audio_format::dsd`](src-tauri/crates/core/src/audio_format/dsd/pcm.rs) (`DsdToPcm::new_with_taps`)
- Ouverture de stream : [`ActiveStream::open`](src-tauri/crates/app/src/audio/crossfade.rs)
- Sortie exclusive : [`audio/wasapi_exclusive.rs`](src-tauri/crates/app/src/audio/wasapi_exclusive.rs) (Windows, opt-in, fallback cpal shared)

Pour le public audiophile (celui qui active déjà WASAPI Exclusive + la précision DSD), il manque le **vrai DSD bit-perfect** : envoyer le flux 1-bit tel quel à un DAC compatible, qui le décode nativement, au lieu de le convertir en PCM côté hôte.

## Objectif — DSD natif via DoP (DSD over PCM)

Ajouter un chemin **DoP** : encapsuler le flux DSD 1-bit dans des trames PCM (marqueurs `0x05` / `0xFA` en poids fort, échantillons 24-bit @ `dsd_rate / 16`) et l'envoyer au DAC **en WASAPI Exclusive uniquement**. Le DAC détecte le motif DoP, reconstruit le flux 1-bit d'origine et affiche « DSD » — sortie réellement bit-perfect. Cibles : DSD64 / DSD128 / DSD256.

### Périmètre v1 (Windows-only)

- [ ] **Probing des capacités réelles du DAC** en mode exclusive (fréquences / formats / canaux acceptés au-delà de la table générique de l'OS) — prérequis : savoir *si* le device accepte le débit DoP avant de basculer.
- [ ] **Encodeur DoP** : DSD 1-bit → trames PCM 24-bit marquées, au rate `dsd_rate / 16` (DSD64 → 176.4 kHz, DSD128 → 352.8 kHz, DSD256 → 705.6 kHz).
- [ ] **Négociation au stream-open** : si (a) sortie = WASAPI Exclusive, (b) le DAC accepte le format DoP requis, (c) l'utilisateur a activé l'option → chemin DoP ; **sinon fallback automatique** vers le DSD → PCM existant (déjà en place, ne rien casser).
- [ ] **Moteur DoP persistant gapless** : ne verrouiller le DAC qu'une seule fois pour enchaîner les pistes DSD sans clic ni re-lock.
- [ ] **UI** : toggle « DSD natif (DoP) » dans Réglages → Playback + popup des sorties, badge dans l'`AudioPipelinePopover` / `AudioQualityFooter` (« DSD natif » vs « DSD → PCM »).
- [ ] **i18n** : nouvelles clés propagées sur les 17 locales (fr source de vérité).
- [ ] **Docs** : [`docs/features/playback.md`](docs/features/playback.md) + note cross-cutting DSD dans `CLAUDE.md`.

### Contraintes techniques

- **WASAPI Exclusive obligatoire** : le DoP ne passe pas par le mixeur partagé Windows (il resample → détruit le motif). Uniquement sur le chemin `wasapi_exclusive.rs`.
- **Hot path audio** : l'encapsulation DoP se fait côté décodeur (avant le ring SPSC), jamais dans le callback. Respecter la règle « le callback n'alloue/lock/log pas ».
- **Volume géré par le DAC/ampli** en DoP (bit-perfect oblige — pas de gain côté hôte). Griser / neutraliser le volume logiciel quand le chemin DoP est actif, comme pour le bit-perfect existant.
- **Fallback silencieux** si le DAC refuse : retomber sur DSD → PCM sans erreur visible, juste le badge qui change.

### Hors périmètre v1

- macOS / Linux (l'équivalent CoreAudio / ALSA DoP est un chantier séparé).
- DSD512+ (à ajouter si un DAC de test le supporte).

## Notes

- Le chemin DSD → PCM existant reste le comportement par défaut et le fallback ; DoP est une couche opt-in par-dessus, sans régression pour les DAC non compatibles.

### LINKED ISSUE #497: feat(audio): native DSD via DoP on macOS (CoreAudio hog mode)
Follow-up to #495 (native DSD via DoP). DoP now ships on **Windows** (WASAPI Exclusive) and **Linux** (raw `hw:` ALSA). This issue tracks the **macOS** backend via **CoreAudio hog mode**, which was deliberately deferred because it's low-level FFI that needs an actual Mac to compile + validate.

## Why it's a separate task

The whole DoP stack is already platform-agnostic — the encoder ([`DsdToDop`](../blob/main/src-tauri/crates/core/src/audio_format/dsd/dop.rs)), the decoder path (`play_dop_track`), the engine (`switch_output_for_track`), and the shared byte/idle packing ([`dop_pack`](../blob/main/src-tauri/crates/app/src/audio/dop_pack.rs), including `fill_dop_period_i32` / `render_dop_silence_i32` for MSB-justified 32-bit samples). **Only a new output backend is needed**, exactly like `alsa_exclusive` was for Linux.

## Implementation plan

Add `src-tauri/crates/app/src/audio/coreaudio_exclusive.rs` (`#[cfg(target_os = "macos")]`, registered in `audio/mod.rs`) exposing:

```rust
pub fn spawn_coreaudio_dop_output_thread(
    shared: Arc<SharedPlayback>,
    app: AppHandle,
    device_name: Option<String>,
    dop: DopFormat,
) -> AppResult<(Producer<f32>, OutputHandle)>
```

mirroring `spawn_alsa_dop_output_thread`'s contract (returns the ring producer + an `OutputHandle` with `dop_rate: Some(dop.sample_rate)`, or an error → the engine falls back to DSD → PCM).

Steps inside:
1. **Resolve the device** — `kAudioHardwarePropertyDefaultOutputDevice` (or match `device_name` against `kAudioObjectPropertyName`).
2. **Acquire hog mode** — set `kAudioDevicePropertyHogMode` to our pid for exclusive access (no system mixing). Release it on teardown.
3. **Set the physical format** — `kAudioStreamPropertyPhysicalFormat` on the output stream: `mSampleRate = dop.sample_rate`, `mFormatID = kAudioFormatLinearPCM`, `mFormatFlags = kAudioFormatFlagIsSignedInteger | kAudioFormatFlagIsPacked` (and `kAudioFormatFlagIsBigEndian` **cleared** for LE), 32-bit. If the device won't take the exact DoP rate → error → fallback.
4. **Feed samples** — `AudioDeviceCreateIOProcID` + `AudioDeviceStart`; the IOProc fills the output `AudioBufferList` from the SPSC ring via [`dop_pack::fill_dop_period_i32`](../blob/main/src-tauri/crates/app/src/audio/dop_pack.rs) (MSB-justified — marker in the top byte, same as ALSA `S32_LE`) and `render_dop_silence_i32` on pause/drain.
5. **Teardown** — `AudioDeviceStop` + `AudioDeviceDestroyIOProcID` + release hog mode; report a device loss through `output::notify_device_lost` + `schedule_device_rebuild` like the other backends.

### Wiring (already stubbed for it)

- `output::spawn_output_with_mode` currently routes macOS DoP into the "not supported yet" arm — swap it for `coreaudio_exclusive::spawn_coreaudio_dop_output_thread`.
- `AudioEngine::switch_output_for_track` sets `exclusive_available = false` on macOS — flip it to `true` so the DoP toggle engages (the hog-mode open IS the exclusive path, same reasoning as Linux `hw:`).
- Settings toggle: `SettingsView.tsx` `dopPlatformSupported` currently shows on Windows + Linux — add macOS.
- Add `coreaudio-sys` (or `coreaudio-rs`) under `[target.'cfg(target_os = "macos")'.dependencies]`.

## Validation

Needs a Mac + a DoP-capable USB DAC. Confirm: the DAC displays "DSD", playback is bit-perfect, pause/seek don't drop DoP lock (the `0x69` idle frames), and a device that refuses the DoP rate falls back cleanly to DSD → PCM.

## Notes

- The MSB-justified 32-bit convention (`dop_word_to_i32`) is shared with the Linux backend and already unit-tested; reuse it rather than re-deriving.
- Playing DoP to a non-DoP DAC is white noise — keep it behind the existing opt-in / default-OFF toggle.

InstaZDLL:
Closes #495 and #497.

Native DSD playback via **DoP (DSD over PCM)** on **all three desktop OSes**: a `.dsf` / `.dff` track is shipped as raw 1-bit DoP frames to a DoP-capable DAC, which decodes the DSD in hardware — truly bit-perfect, no host-side FIR conversion.

## Platform coverage

| OS | Exclusive backend | Status |
|----|-------------------|--------|
| **Windows** | WASAPI Exclusive | done, compiled |
| **Linux** | raw `hw:` ALSA, `S32_LE` marker-MSB-justified | done — compiles + clippy clean + `dop_pack` tests green on Linux (alsa 0.10) |
| **macOS** | CoreAudio hog mode + forced physical format via `AudioUnit` | done — compiled + clippy-clean **on a Mac**, CoreAudio FFI **runtime-probed** (fallback confirmed) |

Written + verified on real Linux and macOS boxes, not blind — only a borrow-scope fix (Linux) and two import fixes (macOS) were needed against the actual crate APIs.

The stack is platform-agnostic: the encoder, decoder path, engine switch, and the shared [`dop_pack`](../blob/feat/dsd-dop-495/src-tauri/crates/app/src/audio/dop_pack.rs) packing are one implementation; each OS only adds a thin output backend.

## ⚠️ Requires on-hardware validation before merge

I don't have a DoP-capable DAC, so two assumptions **must be checked on real hardware**:
- **Bit order** — DoP payload is MSB-first in time; DFF verbatim, DSF bit-reversed. Wrong ⇒ harsh noise.
- **DoP idle/silence** — pause/underrun emit `0x69` payload frames with a live marker so the DAC keeps DoP lock. Wrong ⇒ clicks/dropout on pause.

Opt-in, **default OFF, fully fail-soft**. ⚠️ **Playing DoP to a non-DoP DAC produces white noise** — hence the explicit toggle for users who know their DAC supports it.

## How it works

- **Core encoder** [`DsdToDop`](../blob/feat/dsd-dop-495/src-tauri/crates/core/src/audio_format/dsd/dop.rs) — pure, streamable: 1-bit DSD → 24-bit DoP words (marker `0x05`/`0xFA` per frame, payload MSB-first) at `dsd_rate / 16` (DSD64 → 176.4 kHz …). **8 unit tests**.
- **Shared packing** [`dop_pack`](../blob/feat/dsd-dop-495/src-tauri/crates/app/src/audio/dop_pack.rs) — byte (WASAPI packed/padded) + MSB-justified i32 (ALSA/CoreAudio) word writers, DoP idle-frame generators, and `fill_dop_period*` (pull whole frames, silence-fill underrun). **8 unit tests**.
- **Backends** — WASAPI: forced DoP format + a bit-exact `run_dop_event_loop`. ALSA: raw `hw:` at the exact DoP rate in `S32_LE`, rate re-verified, XRUN-recovered. Both surface a device loss like the PCM path. A refused format errors → fallback.
- **Per-track output switch** — on a cold `LoadAndPlay`, `maybe_switch_dop_output` parses the DSD header and calls `AudioEngine::switch_output_for_track`, which re-opens the exclusive output at the DoP rate and hands the fresh ring producer straight back to the decoder (not via `SwapProducer`). No-op when the format already matches. `play_dop_track` is the bit-perfect twin of `play_track` (no crossfade / gapless / EQ / RG / normalize / mono / speed).
- **Gating** — Windows: DoP rides the WASAPI Exclusive opt-in. Linux: the DoP toggle engages the `hw:` path itself. Elsewhere: DoP stays OFF, no wasteful rebuilds.
- **UI** — a "Native DSD (DoP)" toggle under Settings → Playback (default OFF, shown on Windows + Linux); the pipeline popover shows a "Native DSD" pill from `player_get_state.dop_active` (what actually engaged). i18n across all 17 locales.

## Validation done

- `cargo test -p waveflow-core` — DSD tests green (8 DoP encoder + existing).
- `cargo check` + `cargo clippy` (app + core) clean, and `dop_pack` unit tests compile (app-crate tests run in CI).
- `bun run typecheck` + `bun run lint` clean.
- **Windows** build green. **Linux** `cargo check` + `cargo clippy` clean, `dop_pack` tests green (alsa 0.10). **macOS** `cargo check` + `cargo clippy` clean and the CoreAudio FFI runtime-probed on a MacBook (device/hog/format-negotiation execute; built-in speakers offer no DoP rate → clean fallback to DSD → PCM, as designed).
- **Still needs on-hardware DoP validation** with a real DoP-capable DAC on each OS (bit order + idle-frame lock) — none of the three dev boxes had one connected.

## Summary by CodeRabbit

  * Ajout de la sortie DSD native via DoP vers les DAC compatibles.
  * Ajout d’un réglage pour activer ou désactiver le mode DoP.
  * Affichage de l’état DoP actif et du transport bit-perfect dans le pipeline audio.
  * Prise en charge de Windows, Linux et macOS en sortie exclusive.
  * Repli automatique vers la conversion DSD → PCM si nécessaire.

* **Documentation**
  * Mise à jour des conditions et limites du mode DoP.

* **Localisation**
  * Ajout des traductions du mode DSD natif via DoP.


InstaZDLL:
Cross-checked the DoP protocol layer against an independent GPL-3.0 implementation (`rustmusic`, which ships a "WASAPI DoP" pipeline mode). Licence-compatible with WaveFlow, so reading it is fair game.

**Everything in the protocol core agrees** — markers `0x05`/`0xFA`, marker in bits 23..16 of the 24-bit word, alternation per *frame* shared across channels, carrier rate `dsd_rate / 16`, DSF bit-reversal, and `0x69 0x69` silence carrying live markers so the DAC keeps its lock. Our ALSA / CoreAudio path also matches theirs on MSB justification.

**One real divergence, now fixed here** (`6afa3d18`): `write_dop_word`'s padded branch right-justified the word in the 32-bit container, leaving `0x00` in the top byte — precisely the byte the DAC scans for the cadence. [WAVEFORMATEXTENSIBLE](https://learn.microsoft.com/windows-hardware/drivers/ddi/ksmedia/ns-ksmedia-waveformatextensible#remarks) requires the valid bits left-aligned when `wValidBitsPerSample < wBitsPerSample`, and `rustmusic` declares the same `(32, 24)` format and stamps the marker into the MSB. The in-code comment asserting WASAPI wanted the low position was inverted; corrected too.

Reachable only on DACs that refuse packed 24-bit, since `DOP_FORMAT_CHAIN` tries `Pcm24Packed` first (container == word, no shift needed).

The same inversion exists on the plain PCM path, which predates this branch and ships today — fixed separately in #520.

**What this does and doesn't buy us:** the wire format is now corroborated by the spec and a second implementation, which is the part hardware testing would most likely have caught. It says nothing about device negotiation (exclusive mode, `hw:` ALSA, CoreAudio hog) or an actual DAC locking — that still needs real hardware.

InstaZDLL:
@coderabbitai Thanks — this was a useful pass. 16 of the 18 findings landed in 8 commits (`0e3d4da3`..`f203eb45`); 2 are deferred with a reason. Disposition below, in your order.

### Fixed

**`Cargo.toml`** — the comment was wrong: the backend only ever used the safe `macos_helpers` layer, never `coreaudio::sys`. Rewritten.

**`alsa_exclusive.rs` — `resolve_hw_device`** — agreed, and the failure mode is worse than a wrong device: a DoP stream could open on whatever card sits at index 0. Friendly names are now matched against `alsa::card::Iter` (exact on the short name, containment on the long one), and an explicit name we can't place returns an error instead of guessing — the engine then falls back to DSD → PCM on the device the user actually picked. Only an *absent* selection still takes `hw:0,0`.

**`alsa_exclusive.rs` — partial `writei`** — fixed. The unwritten tail is re-offered until the period lands; recovery re-prepares and retries *only* the remainder (re-sending accepted frames would duplicate them and shift the cadence). `Ok(0)` on a blocking device is treated as a device loss rather than spun on.

**`crossfade.rs` — short DSD reads** — fixed for both the DoP path you flagged and the DSD → PCM path next to it, which had the same single-`read` shape. `read_full_block` loops until the aligned block is full or the file genuinely ends, so a short return now means EOF, which is what both callers already assumed.

**`crossfade.rs` — stride/offset duplication** — extracted to `interleave_stride(&DsdLayout)` and `aligned_byte_offset(target, stride, data_len)`, used by all four sites.

**`decoder.rs` — channels guard** — added; a zero channel count would have the backend draining frames of nothing and the position clock dividing by a placeholder.

**`decoder.rs` — speed on DoP** — real bug, thank you. Speed is a resampler feature the DoP path doesn't run, but `current_position_ms` still scaled by it, so a 1.5× left the progress bar and the scrobble clock running away from music playing at 1×. A `DopSpeedPin` guard holds the shared speed at 1.0 for the track and restores the user's value on every exit path, so the next PCM track still gets it.

**`decoder.rs` — A-B repeat** — implemented. You're right that the doc comment's justification didn't apply: A-B is a seek, and a seek is the one transport action that touches no DoP word. It now mirrors `play_track`'s handling.

**`dop_pack.rs` — silence phase** — this was the most valuable finding, and the fix is broader than the underrun seam. The cadence was being numbered *twice*: the encoder stamped real words, each backend numbered its own idle frames. The encoder's phase also restarts on `reset()` after every seek. So a marker could repeat or skip at exactly the seams where the DAC is watching. The backend is now the single source — `stamp_marker` rewrites bits 23..16 of every outgoing frame, real or idle, from the output thread's own counter. Structurally impossible to desync. Cross-checked against an independent GPL-3.0 implementation that reaches the same encoder/backend split. New tests cover the override and the real → idle → real seam.

**`dop_pack.rs` — byte/i32 duplication** — done, and it's directly related to the above: the two near-identical copies are how the cadence logic could have drifted in the first place. Both now run one generic fill/idle path over a `DopSink`.

**`engine.rs` — DoP output reuse** — fixed. `OutputHandle` carries the whole `DopFormat` instead of just the rate, and the engine compares it as a unit.

**`output.rs` / `state.rs` / `commands/player.rs` / `player.ts` / `SettingsView.tsx` comments** — all corrected; three exclusive backends have shipped on this branch since those were written.

**`wasapi_exclusive.rs` — `sample_bytes`** — using the helper.

**`AudioPipelinePopover.tsx`** — fixed, and `isSpeedShifted` had the same problem, so it's included: all five DSP indicators are forced off under `isDopNative`.

**`SettingsView.tsx` — toggle race** — writes are chained through a ref'd promise so two fast clicks can't land in the backend out of order, and the rollback is conditional on still being the latest intent.

**`SettingsView.tsx` — profile-scoped reload** — fixed; the effect keys on `activeProfile?.id` with a `stale` latch, matching the pattern already used elsewhere in the file.

### Deferred, with reasons

**`coreaudio_exclusive.rs` — device-loss detection** — addressed, but by polling rather than a listener. `macos_helpers` does export `AliveListener`, and that is the right long-term mechanism; its API needs a macOS build to validate and this project has neither a macOS CI job nor a reachable Mac right now. So the thread parks on `recv_timeout` and queries the device object between waits, using only FFI already exercised on this branch, then reports through `notify_device_lost` + `schedule_device_rebuild` like the other two backends. Worst case it never fires, which is today's behaviour — no regression risk. `AliveListener` is a follow-up.

**`coreaudio_exclusive.rs` — restore the physical stream format** — valid, not fixed. `coreaudio-rs` 0.14.2 has `set_device_physical_stream_format` and `get_supported_physical_stream_formats` but **no getter for the current format**, so saving it means raw `AudioObjectGetPropertyData` FFI — new unsafe surface that can't be compiled, let alone run, anywhere in this project's current setup. Writing it blind would be worse than the leak it fixes. Tracking as a follow-up alongside `AliveListener`.

**Tests for the DoP position / A-B behaviour** — not added. `play_dop_track` needs a real `AppHandle` plus a live output thread; the app crate has no harness for that, and its test binary doesn't even launch on Windows. The `dop_pack` layer, which is where the cadence logic now lives, is unit-tested (10 tests, all passing).

### Validation

`cargo clippy -p waveflow --all-targets -- -D warnings` on both the default and `--no-default-features --features updater` configs, `bun run typecheck`, `bun run lint`, `cargo test -p waveflow-core`. The `dop_pack` tests were run in an isolated crate as well, since the app crate's test binary can't start on Windows (Tauri DLL) — CI's Linux job runs them for real.

Still outstanding on the PR as a whole: **no validation on real DoP hardware**. The protocol layer has been cross-checked against an independent implementation, but device negotiation and actual DAC lock are unverified. The feature is opt-in and default OFF, and will be announced as such.

InstaZDLL:
@coderabbitai Round 2 — 2 fixed, 1 partially, in `d83d23c0`..`2fd7f9d9`.

**`decoder.rs` — `resync` swallows an explicit return to 1.0×** — correct, and a good catch: inferring intent from the effective value can't distinguish "the user asked for 1.0×" from "we pinned it ourselves", so a track started at 1.5× and reset to 1× mid-play would come back at 1.5× on the next PCM track. Rather than shadow the preference in a second field, the pin now triggers on the existing `speed_dirty` flag — set by `set_playback_speed` on every user request, and consumed by nobody else while the DoP loop is running (the PCM loop is the normal consumer and it isn't executing). `new()` clears the flag its own pin raises, `resync()` swaps it and takes whatever value it finds — 1.0× included — as the new `restore`.

**`crossfade.rs` — sub-stride remainder at a premature EOF** — valid, fixed on both paths. `read_full_block` does only come up short at real EOF, but a header claiming more payload than the file holds (truncated file, bad `data_len`) makes that EOF land mid-cycle. The short read is now trimmed to a whole stride via a shared `sub_stride_trim`, and a remainder below one stride is treated as EOF — matching what the `aligned == 0` branch above already does.

**`SettingsView.tsx` — profile-scoped write queue** — the cross-profile leak is fixed: each queued entry captures the profile id it was scheduled under, and both the write and the rollback drop out if the active profile changed in the meantime.

The second half — disabling the toggle while the panel loads — I'm skipping here, and deliberately. The stale-value-during-load window isn't specific to this control: `normalize`, `mono`, `gapless`, `replaygain` and `dsd_taps` all hydrate from the same `playerGetAudioSettings()` effect and all predate this PR. Gating one of the six would be inconsistent, and gating all six is a change to an existing panel that has nothing to do with native DSD. Better as its own change.

Two CoreAudio items from round 1 remain deferred for the reasons given there (no macOS CI, no reachable Mac, and no getter for the current physical format in `coreaudio-rs` 0.14.2).

Validation unchanged: clippy `-D warnings` on both feature configs, typecheck, lint, `waveflow-core` tests. CI was fully green on the previous push, including `Rust (ubuntu-latest)`, which is the only job that compiles the ALSA backend and actually runs the `dop_pack` tests.

---

### Round 3 — `184de8f5`..`dd70a23b`

**`decoder.rs` — DoP negotiation only on `LoadAndPlay`** — this one was a real bug and the most serious thing in the whole review, thank you. `LoadRemoteFileAndPlay` and `LoadUrlAndPlay` never touch the output format, so a DoP stream survived into the next remote-file or web-radio track — and that output reads every ring value as a 24-bit DoP word, stamps a marker on it and hands it to a DAC still in DSD lock. Ordinary PCM comes out as full-scale noise. Both branches now call `restore_pcm_output` at the same point `LoadAndPlay` negotiates (after the ring drain, before the open); the engine no-ops when the output is already PCM.

I checked the fourth path you listed: the `store_next` / prefetch swap is already safe — every prefetched stream is opened with `dop = false` and there's a comment saying so, and a DoP track never prefetches in the first place. Going the other way — actually *enabling* DoP for a locally-reconciled remote track — is a feature rather than a fix (it needs the rebuild + `play_dop_track` branch too), so I've noted it in the helper's doc comment instead of growing this PR.

**`crossfade.rs` — truncation in the seek offset** — valid, fixed on both DSD paths. `rate * channels / 8 / 1000` collapsed to an integer bytes-per-ms before it ever met `ms`: DSD64 stereo is 705.6 and became 705, so the error scaled with the seek target — about 0.15 s short three minutes in. One shared `dsd_byte_offset_for_ms` now divides last, in u128.

**`SettingsView.tsx` — profile-scoped writes** — the queue leak was fixed in round 2 and I'm holding there on the rest, deliberately:

- Passing a captured profile id down to `player_set_dsd_dop` so it can use `require_profile_pool_for(Some(id))`: the API does exist, you're right about that. But *every* audio command — `player_set_normalize`, `player_set_mono`, `player_set_gapless`, `player_set_replaygain`, `player_set_dsd_precision` — resolves the active profile pool the same way, and all of them have exactly this exposure. Making DoP the one exception doesn't close the pattern, it just makes it inconsistent. If the window is worth closing it should be closed for the audio command family at once, in its own change.
- Disabling the control during hydration: same argument, and the same five siblings hydrate from that one `playerGetAudioSettings()` effect.
- Ignoring stale reads by generation: already covered — the effect has a `stale` latch that drops a response belonging to the profile we left.
- The test: this project ships no frontend test suite (no vitest/jest harness), so there's nothing to add it to.

What remains after this round is unchanged: the two deferred CoreAudio items, and the fact that **nothing here has been validated against real DoP hardware**.

InstaZDLL:
@coderabbitai Round 4 — both fixed in `0514647f`.

**`decoder.rs` — swallowed `switch_output_for_track` errors** — valid, and worse than it looks from the call site. Reading the engine side: that function only returns `Err` after it has already stopped the previous output *and* failed to open a replacement (it even calls `publish_output_lost_if_gone` on the way out). So returning `false` there meant carrying on with a producer whose consumer is gone — the UI showing a playing track with no device on the other end, silently. All three call sites now emit `player:error` and go Idle, exactly like the `ActiveStream::open` failure right below them.

`Ok((_, false))` keeps its meaning untouched, as you asked: that's the DAC refusing the DoP format with a working PCM output already open behind it, and it still falls back to DSD → PCM.

**`crossfade.rs` — unclamped u128 → u64 cast** — fixed with `.min(u64::MAX as u128)`. Only reachable with an absurd `ms`, but you're right that the failure mode is nasty rather than loud: `aligned_byte_offset` clamps the top end only, so a wrapped offset would land somewhere plausible inside the payload and be seeked to as if it were real.

Validation: clippy `-D warnings` on both feature configs, typecheck, lint. The previous push was green on every job including `Rust (windows-latest)` and `Analyze (rust)`; `Rust (ubuntu-latest)` — the one that compiles the ALSA backend and runs the `dop_pack` tests — was still running when this push landed, so it re-runs against this commit.

PR #521, opened 2026-08-18
fix(audio): restore the CoreAudio physical format and watch IsAlive

InstaZDLL:
Closes the two macOS items deferred during the #496 review. Both were deferred for the same reason — macOS builds nowhere in this project (no CI job) and the Mac was offline — so writing HAL code blind would have been worse than the leaks it fixed. The Mac is back, so this is **compiled and run there**, not reasoned about.

## Physical stream format is restored

Pinning the physical format is a change to the **device**, not to our stream. A DoP track left the DAC clocked at the DoP rate for every other app on the machine, indefinitely — the most user-visible of the two.

The format in force is now read before the change and put back by a `PhysicalFormatGuard`, so it also runs on the `?` paths (AudioUnit build / init / start failures), and before hog mode is released — restoring the format on a device we no longer own would be a race with whoever takes it next. The guard is declared before the `AudioUnit` so drop order stops the unit first.

`coreaudio-rs` 0.14.2 sets that property and lists supported formats, but never returns the one in force (its setter reads it internally and drops it), so the read is one direct `AudioObjectGetPropertyData` call. It uses `objc2-core-audio` / `objc2-core-audio-types` at the same version and feature set `coreaudio-rs` itself pins, so the `AudioStreamBasicDescription` we read is the exact type its setter takes. **No new crates enter `Cargo.lock`** — both were already in the graph — so the Flatpak generated sources stay valid (verified with `check-sources.py`).

## Device loss is pushed, not guessed

`AliveListener` subscribes to the device's `IsAlive` property, so the wait in the output thread only decides how soon a sleeping thread notices the flag. The previous property-query poll stays as the fallback when registration fails. The listener registers a pointer to itself, so it stays a local of the frame that registered it and never moves; its `Drop` unregisters.

## Validation

Run on the Mac (macOS 26.6.1), not inferred:

- `cargo check -p waveflow --all-targets` — clean, and clippy reports nothing on this file.
- The ignored runtime probe now exercises both additions, so neither is validated by compilation alone:

```
default output device id = 72, current hog pid = -1
current physical format: 48000 Hz, 32 bits, 2 ch, format id 0x6c70636d
alive listener registered, device reports alive
DoP 176400 Hz / S32 int / 2ch: not available
```

(`0x6c70636d` is `'lpcm'`; the built-in speakers offering no DoP rate is the expected fallback case.)

- `cargo test -p waveflow --lib audio::` on macOS — 66 passed, which also runs the `dop_pack` marker-cadence tests on a third platform.
- Windows: `cargo clippy -p waveflow --all-targets -- -D warnings` clean.

Still unchanged, and worth repeating: **none of the DoP work has been validated against a real DoP DAC.** This PR removes two side effects; it doesn't test playback.

## Summary by CodeRabbit

  - Restauration automatique du format audio physique initial après une lecture DoP en mode exclusif sur macOS.
  - Détection plus réactive de la déconnexion des périphériques audio.

- **Corrections**
  - Ajout d’une vérification périodique de secours lorsque la surveillance directe du périphérique ne peut pas être activée.
  - Amélioration de la gestion des sorties anticipées afin de garantir la restauration du format audio.

- **Documentation**
  - Documentation macOS mise à jour pour expliquer la restauration du format et la détection de perte du périphérique.


InstaZDLL:
@coderabbitai All three fixed.

**`ioDataSize` passed from a shared borrow** — the important one. It's an in/out parameter: CoreAudio writes the size it actually produced back through that pointer, so deriving it from a `&` is UB regardless of whether it miscompiles today. Now `let mut data_size` + `NonNull::from(&mut data_size)`. Worth noting that upstream `coreaudio-rs` 0.14.2 does it the unsound way in its own setter — which is where I mirrored the call shape from, so thank you for not taking the precedent at face value.

**Read failure downgraded to `None`** — you're right that this defeated the point. Proceeding with nothing to restore means re-clocking the user's device with no way back, which is the exact leak the PR exists to close. It now aborts the DoP open, which isn't a hard failure for the user: it's the ordinary fail-soft path, so the engine plays the track as DSD → PCM on that device. In practice a read failure all but guarantees the set would fail too — the setter reads the same property first.

**Docs** — reworded: the listener is the normal path, `get_hogging_pid` polling is the fallback when `AliveListener::register()` fails.

Re-verified on the Mac after the change, since the FFI call shape moved:

```
current physical format: 48000 Hz, 32 bits, 2 ch, format id 0x6c70636d
alive listener registered, device reports alive
```

Plus clippy clean on macOS and `-D warnings` clean on Windows.

PR #522, opened 2026-08-18
feat(mini-player): volume slider and mute in the cover overlay

### LINKED ISSUE #511: feat: Mini player - Volume controll
Hi @InstaZDLL , i hope you well ...

will be great to have a volume control + mute  , in  mini player too .

thank you.

InstaZDLL:
Closes #511 (@jo-el414).

## What's there

A compact volume slider + mute button on a second row of the cover's hover overlay, under the transport controls. That placement keeps the widget's stated design — idle shows just the artwork — and avoids any layout shift in a window the user can resize down to something small.

Interaction matches the PlayerBar's `VolumeControl` exactly: pointer drag with capture, wheel (5 % steps), arrows / Home / End, `role="slider"` with the full ARIA value set. It reuses the existing `player.volume.*` i18n keys, so no new strings in the 17 locales.

## The part that isn't UI

Volume belongs to the engine, but each window runs its own `PlayerContext`. Hydration happens once at mount and there was no volume event, so a slider in the mini-player would have drifted from the main window's while both drove the same output — the first thing anyone would report after this shipped.

`player_set_volume` now emits `player:volume-changed` (0..1, same scale as the snapshot field) and every window adopts it. One subtlety: our own pushes echo back, and they're debounced 60 ms, so during a drag an echo can carry a value the user has already moved past. The window that just touched its slider therefore keeps authority for 500 ms and ignores what comes back. The sleep-timer fade benefits from the same path — the mini-player's slider now follows it down.

Not fixed here, and now written down in the docs: **repeat and shuffle still drift between the two windows.** `player:options-changed` is only emitted for external surfaces (MPD), so the in-app toggles update optimistically and never reach the other webview. It's a pre-existing gap with a different fix, and folding it in would have made this PR about something else.

## Also

The overlay now reveals on `focus-within`. Its controls were already in the tab order while invisible — tolerable for buttons, not for a slider you can drag without seeing it. Making the overlay `inert` instead would have removed keyboard access to the transport controls entirely, which is worse.

## Validation

`bun run typecheck`, `bun run lint`, `cargo clippy -p waveflow --all-targets -- -D warnings`. Not exercised at runtime — I don't have the app running here, so the mini-player window itself is unverified visually.

## Summary by CodeRabbit

  * Ajout d’une barre de volume compacte au mini-lecteur, avec bouton de sourdine.
  * Réglage du volume à la souris, à la molette et au clavier, avec prise en charge des touches directionnelles, `Home` et `End`.
  * Les commandes s’affichent au survol ou lors de la navigation au clavier.
* **Améliorations**
  * Le volume est synchronisé entre les fenêtres de l’application.
  * Les changements de volume et de sourdine restent cohérents lors d’actions rapides.


InstaZDLL:
@coderabbitai review

PR #524, opened 2026-08-18
fix(player): broadcast repeat, shuffle and like to the other window

### LINKED ISSUE #523: bug: repeat, shuffle and like drift between the mini-player and the main window
Found while implementing #511 (PR #522). Not fixed there — different mechanism, and folding it in would have made that PR about something else.

## What happens

The mini-player and the main window each run their own `PlayerContext` in their own webview. They drive **one** engine, but several pieces of shared state only reach a window if the backend broadcasts an event. Three don't:

| State | Symptom |
| --- | --- |
| Repeat mode | Cycling repeat in the mini-player leaves the main window's button showing the old mode, and vice versa |
| Shuffle | Same |
| Like (♥) | Liking a track in one window doesn't fill the heart in the other until the track changes |

The *engine* is correct throughout — it's only the two UIs disagreeing about what they're showing. Which is arguably worse than a real bug: the user toggles repeat in the mini-player, opens the main window, sees repeat off, and clicks it again — turning it off for real.

## Why

- **Repeat / shuffle** — `player:options-changed` exists but [is emitted from one place only](https://github.com/InstaZDLL/WaveFlow/blob/main/src-tauri/crates/app/src/commands/player.rs), the MPD path. Its own doc comment says so: *"the in-app buttons update their own state optimistically"*. That assumption held while there was exactly one window; the mini-player broke it.
- **Like** — no event at all. `MiniPlayer.tsx` refetches `listLikedTrackIds()` keyed on `currentTrack?.id`, so it only re-syncs when the track changes.

## Repro

1. Open the mini-player (main window hides).
2. Cycle repeat to "all".
3. Restore the main window from the mini-player's Maximize button.
4. The PlayerBar's repeat button is still "off".

## Suggested fix

PR #522 does exactly this for volume and the pattern transfers:

- `player_toggle_shuffle` / the repeat command call the existing `emit_options_changed` instead of leaving it to MPD, and `PlayerContext`'s existing `player:options-changed` listener does the rest — it already adopts both fields.
- Like needs a new event (`track:liked-changed` with `{ trackId, liked }`), plus a listener in the mini-player next to the current fetch.

Repeat/shuffle need no echo guard — they're discrete toggles, not a dragged value, so adopting an echo of our own state is a no-op. That's the one part of the volume plumbing that doesn't carry over.

InstaZDLL:
Closes #523. Found while doing #511 / #522.

## The shape of it

The mini-player is a second webview with its own `PlayerContext`. Both windows drive **one** engine, so anything the user can change from either place has to be broadcast, or the two copies drift. Three didn't travel:

| State | Before |
| --- | --- |
| Repeat / shuffle | `player:options-changed` existed but was emitted from the MPD path only |
| Like (♥) | No event, and two identical copies of the liked-id set — one per surface |

The engine was right throughout; it was the two UIs contradicting each other. That's the part worth fixing: the user sets repeat in the mini-player, opens the main window, sees it off, clicks — and turns it off for real.

## Repeat / shuffle

`player_toggle_shuffle` and `player_cycle_repeat` now call the existing `emit_options_changed`, and `PlayerContext`'s existing listener adopts both fields. Its doc comment said the in-app buttons "update their own state optimistically" and don't need it — true while there was one window, so I rewrote that comment rather than leave the next reader with the assumption that produced the bug.

**No echo guard here**, deliberately, and it's the one thing that doesn't carry over from the volume plumbing in #522: repeat and shuffle are discrete toggles with no debounce, so the emitting window receives an echo of the value it already holds and adopting it is a no-op. Volume needed the guard because its push is debounced and a returning value can be one the user has already dragged past.

## Like

New `track:liked-changed` event carrying `{ trackId, liked }`, so no window that didn't ask for the change pays a round-trip. It's emitted only when the row actually moved — same `did_change` condition as the outbox enqueue immediately above it, which exists for the same reason (a no-op `INSERT OR IGNORE` shouldn't produce a phantom op, or here, repaint every heart in the app).

The set itself moves into `useLikedTracks`, used by the player bar and the mini-player, which each had their own copy of the identical fetch/refetch/optimistic-toggle logic. Two copies of the same state in two webviews is exactly how this drifted, so one hook now owns it and the subscription lives there once.

`LikedView` also refetches on the event, via a `useLikedChanged` sibling hook shaped like the existing `useTrackUpdated`. It was the one surface that could end up showing a list contradicting its own hearts — unlike from the mini-player and the row just sat there.

## Docs

`invariants.md`'s Events section now says *why* these exist ("shared state needs an event, because there is more than one window") rather than listing them, and records the echo-guard asymmetry. `ui.md`'s mini-player section replaces the "repeat/shuffle still drift" note I added in #522.

## Validation

`cargo clippy -p waveflow --all-targets -- -D warnings`, `bun run typecheck`, `bun run lint`.

**Draft because none of it is exercised at runtime.** This is a behaviour that only shows up with two windows open, and it needs the four repro paths walked by hand:

1. Toggle repeat in the mini-player → the main window's button follows (and back).
2. Same for shuffle.
3. Like from the mini-player → the player bar's heart fills, without waiting for a track change (and back).
4. `LikedView` open in the main window, unlike from the mini-player → the row disappears.

## Summary by CodeRabbit

  - Synchronisation entre fenêtres du volume, du mode aléatoire, de la répétition et de l’état « aimé ».
  - Les changements de favoris se reflètent automatiquement dans le mini-lecteur, la barre du lecteur et la vue des titres aimés.
  - Les commandes du lecteur notifient les autres fenêtres après modification.
  - La documentation décrit les événements et comportements de synchronisation.

- **Corrections**
  - Réduction des mises à jour en écho lors des changements de volume.
  - Conservation des changements de favoris reçus pendant le chargement.


InstaZDLL:
@coderabbitai One fixed, two skipped with reasons.

**`useLikedTracks` — event lost against an in-flight reload** — valid, fixed in `f547d0fb`. A `track:liked-changed` arriving while `listLikedTrackIds()` was in flight was applied to the current set and then overwritten by the snapshot the fetch answered with, taken before the change. Worse than a transient glitch: the hook only reloads on track change, so the wrong heart persisted until then. Changes seen during a fetch are now recorded and replayed on top of the snapshot; the map only exists while a request is pending.

I did **not** restructure to subscribe-then-read. Making the subscription part of the fetch effect means tearing it down and re-establishing it on every track change — that trades one millisecond-wide gap at mount for one at every track change, which is strictly worse. What remains uncoverable is an event fired before the very first `listen()` resolves, with no listener in existence to record it; the next track change repairs it, and it's documented in the hook.

The `LikedView` half I'm skipping: it calls `listLikedTracks` (full track metadata, ordered by liked-at) while the hook owns `listLikedTrackIds` (a Set of ids for rendering one heart). They're different queries serving different components; folding the list view into the ids hook would make it fetch metadata nobody there uses. Its stale-response protection is also already in place — the effect has a `cancelled` latch and checks it before every `setState`.

**Serializing like mutations per `track_id`** — skipping, and it's a judgment call rather than a disagreement about the race. It's real: two commits can land in one order and their emits in the other, leaving a window showing the wrong heart.

What makes it not worth a keyed lock across commit-and-emit:

- Reaching it needs two toggles of the *same track* within microseconds, and `toggle_like_track` has exactly one caller — the frontend command. No sync, MPD or scanner path touches it, so it takes a human clicking two hearts in two windows simultaneously.
- The read-modify-write is already atomic; it's inside the transaction. Only the emit ordering is loose.
- The window that clicked is never wrong: it sets state from the command's own return value, which is the committed truth. Only the *other* window can briefly show a stale heart, and the next track change reloads it.

The requested test can't be written either: it needs two concurrent invocations of a `#[tauri::command]` with a live `AppHandle`, and the app crate has no harness for that — its test binary doesn't even start on Windows.

**Serializing repeat/shuffle with a shared lock** — skipping, same weighing, plus one property that changes the calculus. `emit_options_changed` re-reads both values from the database rather than echoing what the caller computed. So even if two concurrent toggles lose an update, every window still converges on the persisted state; the symptom is "a simultaneous double-click counted once", not divergence. Divergence is what this PR exists to fix, and that part holds. Extending the lock to the MPD paths as well would be a meaningful chunk of shared-state machinery for a race that self-corrects.

If either of these ever shows up in a real report, the lock is the right fix and this comment is the record of why it was deferred.

InstaZDLL:
@coderabbitai Applied.

**Readiness barrier — fixed** (`eed32546`). You're right that the two were unordered: `listen()` is a round-trip and Tauri replays nothing emitted before it resolves, so a change landing in that gap was missed by the event *and* could be missed by a snapshot that raced the write — leaving the set wrong until the next track change.

`useLikedChanged` now returns a readiness flag, and both `useLikedTracks` and `LikedView` gate their fetch on it. Everything committed before the snapshot is in it, everything after arrives as an event. A failed `listen()` also flips the flag, so a broken subscription degrades to a stale set rather than to never loading one at all. The cleanup re-arms the barrier, so a callback identity change re-snapshots around its own gap in coverage.

**Delta replay — already in the branch** (`f547d0fb`, previous round). Changes seen while a fetch is in flight are recorded in a map and replayed on top of the resolved snapshot instead of being overwritten by it. The barrier narrows that window but doesn't close it — an event can still land between "listener live" and "snapshot resolved" — so both mechanisms stay.

The ordering rule is now written down in `docs/architecture/invariants.md` under Events, since it applies to any listener-plus-snapshot pair, not just this one.

PR #525, opened 2026-08-18
fix(release): unbundle libwayland-client from the AppImage

InstaZDLL:
The AppImage doesn't start on any Linux whose Mesa is newer than the build runner's. WebKit's WebProcess dies before painting anything, the Rust side carries on regardless, and the user gets an empty window:

```text
Could not create default EGL display: EGL_BAD_PARAMETER. Aborting...
```

## Cause

Tauri's bundler copies the webview's dependency closure into the AppDir, and that closure contains `libwayland-client.so.0`. That library is on the [official AppImage excludelist](https://github.com/AppImage/pkg2appimage/blob/master/excludelist), with this comment:

> `libwayland-client.so.0`
> `# https://gitlab.freedesktop.org/mesa/mesa/-/issues/11316`
> `# New version of Mesa has some dependency issues with libwayland-client if it is bundled`

We build on Ubuntu. Fedora 44 ships Mesa 26.1.6 and libwayland 1.25, the bundled copy wins the lookup, and EGL refuses.

## Evidence

Run against the **released v1.7.0 AppImage** on Fedora 44 / GNOME Wayland / Mesa 26.1.6 / webkitgtk 2.52.5:

| Variant | Result |
| --- | --- |
| As shipped | `EGL_BAD_PARAMETER` ×2, splash safety-net fires at 15 s, empty window |
| `GDK_BACKEND=wayland` instead of the forced x11 | identical — the backend is not involved |
| All four bundled `libwayland-*` removed | WebProcess alive, **0 EGL errors**, `app://ready` fires |
| Only `libwayland-client.so.0` removed | **0 EGL errors** — culprit isolated |
| Repacked by this script, run end to end | profile created, audio device opened, `app.db` + `data.db` written, **UI visible on screen** |

The Rust side was never affected — it created its profile and opened cpal at 44.1 kHz on every single attempt, including the broken ones. That asymmetry is what made this look like a WebKit version problem.

## What this retires

The standing answer to affected users has been that the AppImage is fundamentally incompatible with WebKitGTK 2.52+, that it's a Tauri limitation rather than something we can fix, and that they should install a native package instead. All three are wrong. The engine was never the problem, and the fix is one file.

## The script

[`scripts/fix-appimage.sh`](scripts/fix-appimage.sh) runs right after `tauri build` in both `release.yml` and `test-appimage.yml`:

- removes the library, and **no-ops with a message** if a future Tauri stops bundling it;
- repacks with `mksquashfs` + the AppImage's own runtime header — nothing to download, and no FUSE, which the runner images don't all carry;
- reads the compressor and block size back from the original's superblock rather than assuming: forcing gzip cost 11 MB on a 95 MB bundle, matching zstd lands byte-for-byte on the same size;
- **verifies before overwriting** — the repacked image must re-extract, must have its `AppRun`, and must no longer contain the library, or the script fails the build;
- **re-signs** when an updater `.sig` is present. This matters: the signature covers bytes the repack rewrites, so a stale one would make every updater client reject the release. It refuses to leave a stale `.sig` behind if the signing key is missing.

`squashfs-tools` added to both workflows' apt list, and `*.sh text eol=lf` to `.gitattributes` so a CRLF checkout can't break the shebang on a runner.

## Scope and limits

Validated on one machine. The upstream excludelist entry says why it generalizes, but the repacked artefact produced *by CI* hasn't been run yet — I validated the script against the released AppImage locally on the affected box. Worth doing before the tag: run `test-appimage.yml` from this branch and launch the artifact on Fedora.

One thing I could not test from an SSH session: mounting the AppImage through FUSE. The original fails there in exactly the same way, so it's not a regression from the repack, but "double-click the file in a desktop session" remains unverified.

## Summary by CodeRabbit

  * Synchronisation en temps réel des favoris, du volume, du mode aléatoire et de la répétition entre les fenêtres de l’application.
  * Les listes de morceaux favoris se mettent automatiquement à jour après toute modification.

* **Corrections**
  * Amélioration de la compatibilité des AppImage Linux en supprimant une bibliothèque susceptible de provoquer des erreurs graphiques.

* **Documentation**
  * Mise à jour des informations sur la synchronisation de l’interface et la préparation des versions Linux.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -17,6 +17,9 @@
 *.md        text eol=lf
 *.html      text eol=lf
 *.css       text eol=lf
+# Shell scripts run on CI runners: a CRLF checkout makes bash fail on the
+# shebang with a `\r` nobody can see in the error message.
+*.sh        text eol=lf
 
 # Lockfiles change frequently and benefit from native diff/merge handling,
 # but should still stay LF so CI builds on Linux runners don't see noisy
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -128,6 +128,7 @@ jobs:
             libayatana-appindicator3-dev \
             librsvg2-dev \
             libfuse2 \
+            squashfs-tools \
             libasound2-dev
 
       - name: Setup Rust toolchain
@@ -262,6 +263,19 @@ jobs:
           else
             bun run tauri build --bundles "$BUNDLES"
           fi
+
+      - name: Unbundle libwayland-client from the AppImage (Linux)
+        if: matrix.platform == 'linux'
+        shell: bash
+        env:
+          TAURI_SIGNING_PRIVATE_KEY: ${{ secrets.TAURI_SIGNING_PRIVATE_KEY }}
+          TAURI_SIGNING_PRIVATE_KEY_PASSWORD: ${{ secrets.TAURI_SIGNING_PRIVATE_KEY_PASSWORD }}
+        # Tauri's bundler ships a library the official AppImage
+        # excludelist forbids, and the result doesn't start on any host
+        # with a newer Mesa than this runner. Runs before "Prepare release
+        # assets" so the renamed copy and its .sig are taken from the
+        # fixed bundle. See the script header for the full diagnosis.
+        run: bash scripts/fix-appimage.sh
 
       - name: Scrub Authenticode secrets (Windows)
         if: always() && matrix.platform == 'windows'
--- a/.github/workflows/test-appimage.yml
+++ b/.github/workflows/test-appimage.yml
@@ -45,6 +45,7 @@ jobs:
             libayatana-appindicator3-dev \
             librsvg2-dev \
             libfuse2 \
+            squashfs-tools \
             libasound2-dev
 
       - name: Setup Rust toolchain
@@ -93,6 +94,12 @@ jobs:
         # the build fails after producing the AppImage, so the upload
         # step never runs.
         run: bun run tauri build --bundles appimage -c '{"bundle":{"createUpdaterArtifacts":false}}'
+
+      - name: Unbundle libwayland-client from the AppImage
+        # Same post-processing the real release applies, so an artifact
+        # downloaded from here behaves like the shipped one. No updater
+        # signature is produced above, so the script skips re-signing.
+        run: bash scripts/fix-appimage.sh
 
       - name: Upload AppImage artifact
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
--- a/docs/RELEASING.md
+++ b/docs/RELEASING.md
@@ -242,6 +242,20 @@ pre-release; the beta manifest's payload URLs point there.
 > creates the tag, you don't). The rule's intent (don't bypass
 > release-please for _stable_ versions) is preserved.
 
+## The AppImage is post-processed, and it has to stay that way
+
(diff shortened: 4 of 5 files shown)


InstaZDLL:
@coderabbitai review

InstaZDLL:
Validation de bout en bout terminée — l'artefact **produit par la CI** (run [32192875412](https://github.com/InstaZDLL/WaveFlow/actions/runs/32192875412), branche de cette PR) testé sur la machine qui cassait.

**Étape CI**, sortie brute :

```text
[fix-appimage] unpacking WaveFlow_1.7.0_amd64.AppImage
[fix-appimage] removing usr/lib/libwayland-client.so.0
[fix-appimage] repacking (squashfs, zstd, 131072B blocks)
[fix-appimage] verifying the repacked image
[fix-appimage] WaveFlow_1.7.0_amd64.AppImage rewritten (96M)
[fix-appimage] no updater signature alongside the bundle — nothing to re-sign
```

**Artefact lancé sur Fedora 44 / GNOME Wayland / Mesa 26.1.6 / webkitgtk 2.52.5** :

```text
INFO waveflow_lib::state: created default profile profile_id=1
INFO waveflow_lib::audio::output: cpal output stream opened sample_rate=44100 channels=2 sample_format=F32
=== EGL_BAD_PARAMETER === 0
=== splash 'never fired' === 0
=== DB === app.db, data.db
```

Zéro erreur EGL, le handoff `app://ready` se déclenche (donc React a rendu), et la fenêtre affiche bien l'interface — confirmé de visu par l'auteur du dépôt sur son écran.

Le seul chemin qui reste non testé est inchangé : le montage FUSE en double-cliquant le fichier depuis une session graphique. L'AppImage d'origine échoue identiquement depuis SSH, donc ce n'est pas une régression du repack, mais ça vaut un essai manuel avant le tag.

InstaZDLL:
Dernier trou de validation comblé : **le double-clic depuis une session graphique fonctionne** (montage FUSE, pas extraction), confirmé par l'auteur du dépôt sur Fedora 44.

Le premier essai s'était soldé par un splash qui disparaissait aussitôt, mais la cause n'avait rien à voir avec le repack — cette machine porte un profil créé par un build de dev, et la v1.7.0 refuse une base contenant une migration qu'elle ne connaît pas :

```text
Failed to setup app: error encountered during setup hook: migration error:
migration 20260802120000 was previously applied but is missing in the resolved migrations
```

Avec un profil neuf, l'AppImage montée en FUSE démarre et affiche son interface. Cette PR est donc validée sur les deux chemins d'exécution : contenu extrait et image montée.

Le comportement de rétrogradation est suivi séparément dans #526 — sqlx a raison de refuser, mais le panic silencieux n'aide personne.

InstaZDLL:
**Correction : cette PR était empilée sur #524 par erreur.** Sa branche avait été créée depuis `fix/cross-window-state-drift-523` au lieu de `main`, si bien que son diff embarquait toute la correction de dérive inter-fenêtres — encore en draft et non validée. La merger aurait fait passer l'autre PR avec elle, en silence.

Rebasée sur `main` : un seul commit, cinq fichiers, tous liés à l'AppImage. #524 est inchangée.

Ce que ça implique pour la validation déjà publiée plus haut : l'artefact que j'ai testé sur Fedora avait été construit depuis la branche empilée, donc il contenait aussi le code de #524. Ça ne remet pas en cause la conclusion — l'échec EGL vient d'une bibliothèque embarquée dans le bundle, sans rapport avec la synchronisation du volume ou des likes — mais par rigueur j'ai relancé `test-appimage.yml` sur la branche propre.

PR #528, opened 2026-08-19
fix(db): explain a downgrade instead of dying on the splash

### LINKED ISSUE #526: bug: downgrading to an older version dies with a silent panic instead of explaining itself
Running an older build against a profile a newer one created kills the app during startup with a Rust panic. The user sees the splash screen appear and everything vanish — no dialog, no message, nothing in the UI to act on.

Found while testing the AppImage on Fedora 44 (#525): that machine had a profile from a dev build, and the v1.7.0 AppImage refused it.

## What it looks like

```text
thread 'main' panicked at tauri-2.11.5/src/app.rs:1425:11:
Failed to setup app: error encountered during setup hook: migration error:
migration 20260802120000 was previously applied but is missing in the resolved migrations
```

The splash window is created before the `setup` hook runs, so it paints, then the process dies with it. From the user's side that's indistinguishable from a crash on launch.

## Why it will reach real users

None of these are exotic:

- someone tries a **beta**, then reinstalls the stable channel;
- an **older AppImage** still sitting in `~/Downloads` gets double-clicked;
- a distro package gets **downgraded** after a bad update;
- two versions on the same machine (an AppImage kept alongside the native package) share `~/.local/share/app.waveflow`.

Sqlx refusing is the **right** behaviour — running a newer schema through older code is how databases get corrupted. The refusal isn't the bug. The silence is.

## Where

Both migration paths panic identically because the error just propagates out of `setup`:

- [`db/app_db.rs`](../../src-tauri/crates/app/src/db/app_db.rs) — `migrator.run(&pool)`
- [`db/profile_db.rs`](../../src-tauri/crates/app/src/db/profile_db.rs) — same

Note this is *not* the case [`migration_heal`](../../src-tauri/crates/app/src/db/migration_heal.rs) already handles: that one reconciles checksum drift from CRLF/LF churn on a migration we *do* ship. Here the migration genuinely doesn't exist in this binary, and there's nothing to heal — only something to explain.

## Suggested shape

Distinguish "this profile is from the future" from every other migration failure, and tell the user instead of panicking:

1. Match the `VersionMissing` case sqlx reports (a version in `_sqlx_migrations` the embedded migrator doesn't have) rather than string-matching the message.
2. Show a native dialog — the app already depends on `tauri-plugin-dialog` — along the lines of: *"This library was last used by a newer version of WaveFlow. Install version X or later, or start from a different profile."* The applied-version timestamp is right there in `_sqlx_migrations` and is enough to name what happened.
3. Exit cleanly after the dialog rather than unwinding through the setup hook.

Worth checking whether the splash can be closed before the dialog, otherwise it may sit on top of it.

InstaZDLL:
Closes #526.

## What went wrong

Migrations only ever get appended, so a database records the newest build that ever opened it. Point an older binary at it and sqlx finds a `_sqlx_migrations` row it has no migration for:

```text
migration 20260802120000 was previously applied but is missing in the resolved migrations
```

Refusing is **right** — replaying a newer schema through older code is how databases get corrupted. The bug was that the refusal left `AppState::init`, and an error out of the Tauri `setup` hook is a panic. The splash window is created before `setup` runs, so it painted and the process died with it: indistinguishable from a crash on launch, nothing on screen to act on.

None of the ways in are exotic — a beta then back to stable, an older AppImage still in `~/Downloads`, a distro package rolled back, two versions sharing one data directory.

## The guard

`db::schema_guard` looks for applied versions this binary doesn't ship, and reports the highest one. It runs **before the migrator, and before the checksum heal pass** — the heal writes, and there's no reason to rewrite checksums into a database this build has already decided not to touch. Wired into both `app_db::open` and `profile_db::open`, so it covers every path that opens a database: importing a profile archive exported by a newer build now says so instead of failing on a checksum.

Distinct from `migration_heal`, which reconciles drift on a migration we *do* ship. Here the migration genuinely isn't in this binary and there is nothing to repair, only something to explain.

## Presenting it

This turned out to be the hard half, and all of it follows from one fact: **`setup` runs from inside the event loop** (`RuntimeRunEvent::Ready`), so nothing queued from there is dispatched until `setup` returns — and a fatal path never returns. Three consequences, each measured on Windows rather than assumed:

| | |
|---|---|
| `tauri-plugin-dialog` | Dispatches through `run_on_main_thread` → queued, never drained. `blocking_show` documents itself as unsafe from the main thread for that reason. |
| `splash.close()` | Silent no-op — a close arrives as an *event*. `hide()` takes the runtime's same-thread fast path and applies immediately. The splash has to go: it's `alwaysOnTop` and would sit over the dialog. |
| Dialog on the main thread | `TaskDialogIndirect` creates its window and never shows it — `WS_VISIBLE` stays clear — and the process hangs there forever. A spawned thread carries its own pump and displays normally. |

So: `rfd` called directly (the dialog plugin's own backend), on its own thread, after hiding the splash. The message is English like the tray's seed labels — there's no i18next yet, and no readable language preference, because that setting lives in the database we just refused to open.

`rfd` is pinned to the version and features `tauri-plugin-dialog` already requests, so Cargo unifies it into the copy already in the lockfile: the only `Cargo.lock` change is one line adding it to this crate's dependency list. No new package, and `check-sources.py` passes without regenerating the Flatpak sources.

## Validation

- **6 unit tests** on `schema_guard`, run against the **real** migration trees. Includes `the_guard_fires_on_exactly_what_sqlx_refuses`, which pins the guard to sqlx's own `VersionMissing` — if sqlx ever stops refusing, the guard has started deciding policy by itself and that test says so.
- **End to end on Windows**, against a throwaway app identifier so no real library was touched: fresh profile boots normally with zero errors (the guard is a no-op on a healthy database), then with a forged future migration the splash disappears and this comes up:

```text
WaveFlow can't open this library

This profile's library was written by a newer version of WaveFlow.

Opening it with this older build could corrupt it, so WaveFlow stopped
before touching anything.

Install WaveFlow again from the version you were using before, or start
WaveFlow with a different profile.

Database schema 29990101000000, applied on 2099-01-01 00:00:00.
```

`typecheck` / `lint` untouched (no frontend changes); `cargo clippy --all-targets -D warnings` clean on both feature configs; `cargo check --workspace --all-targets` clean.

**Not verified on macOS or Linux.** The three findings above are Windows measurements. The guard itself is platform-independent, and the dialog is the same `rfd` call the app already makes for file pickers, but the "hidden window / own thread" behaviour was only observed here.

## Summary by CodeRabbit

  * Détection des bases de données créées par une version plus récente de l’application.
  * Affichage d’une boîte de dialogue native indiquant la version concernée et la marche à suivre.
  * Arrêt propre de l’application afin d’éviter toute modification ou corruption de la base.

* **Documentation**
  * Clarification de la politique de migration : les rétrogradations sont refusées pour les bases globales et les profils, y compris lors de l’import d’archives.

### DIFF
--- a/docs/architecture/invariants.md
+++ b/docs/architecture/invariants.md
@@ -80,6 +80,14 @@ sqlx records a SHA-384 checksum in `_sqlx_migrations.checksum` at apply time, so
 
 **Line-ending drift is a non-event.** [`db::migration_heal`](../../src-tauri/crates/app/src/db/migration_heal.rs) reconciles stored checksums against the compiled-in migrator before each `Migrator::run`: when the stored hash matches the LF or CRLF variant of the same SQL (the Windows `core.autocrlf=true` regression), it silently rewrites the row to the canonical hash and logs a warning. A real SQL change still panics, because neither LF nor CRLF normalization will rescue it.
 
+**A database from a newer build is refused, out loud.** Because migrations only ever get appended, a database records the newest build that ever touched it, and an older binary finds a `_sqlx_migrations` row it has no migration for. Refusing is right — replaying a newer schema through older code corrupts it — but the refusal used to leave `AppState::init`, and an error out of the Tauri `setup` hook is a panic: the splash painted, the process died with it, nothing explained (#526). [`db::schema_guard`](../../src-tauri/crates/app/src/db/schema_guard.rs) now detects the case ahead of the migrator, and startup turns it into a native dialog and a clean exit.
+
+Three rules come out of that, in order:
+
+- **Guard before heal, heal before run.** The guard runs first because the heal pass _writes_ — no rewriting checksums into a database this build has already decided not to touch.
+- **The guard must not outlive its reason.** `the_guard_fires_on_exactly_what_sqlx_refuses` pins it to sqlx's own `VersionMissing`; if sqlx ever stops refusing, the guard has started deciding policy on its own and that test says so.
+- **Nothing you queue from `setup` will happen.** `setup` runs from inside the event loop (`RuntimeRunEvent::Ready`), so while it blocks, the loop dispatches nothing — and a fatal path never returns. Three consequences, all measured, all in [`report_fatal_and_exit`](../../src-tauri/crates/app/src/lib.rs): `tauri-plugin-dialog` is unusable because it dispatches through `run_on_main_thread`; the splash must be `hide()`n, not `close()`d, because a close is delivered as an _event_ while `hide` takes the runtime's same-thread fast path; and the dialog needs its own thread, because `TaskDialogIndirect` on the main thread creates its window and never shows it (`WS_VISIBLE` stays clear) and the process hangs there. The splash has to go regardless — it's `alwaysOnTop` and would sit over the dialog. Text is English like the tray's seed labels: the language preference lives in the database we just refused to open.
+
 ### Single writer to SQLite
 
 WAL mode allows concurrent reads but only one writer.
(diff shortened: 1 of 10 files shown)


InstaZDLL:
@coderabbitai review

InstaZDLL:
@coderabbitai review

PR #530, opened 2026-08-19
fix(db): explain a downgrade before the event loop, not inside it

### LINKED ISSUE #529: bug: on Linux the downgrade guard shows nothing and hangs the process
Follow-up to #528 (which closed #526). The guard itself is fine on Linux; the **presentation half is not** — it shows nothing at all and leaves the process hung.

#528 is explicit that its three findings are Windows measurements and that macOS/Linux were not verified. This is the Linux verification. The conclusion inverts: **the spawned thread is not the fix on Linux, it is the cause.**

## Environment

Fedora 44, `webkit2gtk-4.1` 2.52.5, GTK 3.24.52, `rfd` 0.16.0 (`gtk3` + `common-controls-v6`, as pinned).
`main` at `846fb42d`, debug build. Xvfb :99 (1280x900), no window manager. Throwaway `XDG_DATA_HOME`, so `app_data_dir()` resolves to a scratch tree.

## What works

- **6/6 unit tests** on `db::schema_guard`, including `the_guard_fires_on_exactly_what_sqlx_refuses`. The detection half is platform-independent, as claimed.
- **Healthy boot is untouched**: `app://ready` fires, the real splash handoff runs, main window `IsViewable`, splash closed, no warnings. The guard is a no-op on a healthy database.

## What breaks

Forge a future migration into a profile database and relaunch:

```sql
INSERT INTO _sqlx_migrations (version, description, installed_on, success, checksum, execution_time)
VALUES (29990101000000, 'from the future', '2099-01-01 00:00:00', 1, X'00', 0);
```

The guard fires correctly:

```
ERROR waveflow_lib::db::schema_guard: database carries a migration this build does not ship — refusing to open it
      version=29990101000000 installed_on=Some("2099-01-01 00:00:00") scope=Profile
ERROR waveflow_lib: fatal startup error, exiting
      err=This profile's library was written by a newer version of WaveFlow (database schema 29990101000000)
```

Then nothing reaches the screen. Measured with `xwininfo` against the X server:

| #528 expects | Linux does |
| --- | --- |
| `splash.hide()` applies immediately | **no-op** — the splash stays `IsViewable` |
| the dialog displays on its own thread | **never mapped** — no dialog window exists at all |
| `std::process::exit(1)` | **never reached** — killed by hand after 178s |

The process does not die. For a Linux user this is worse than the pre-#528 behaviour: the panic at least terminated; this hangs on a frozen splash.

## Mechanism

`eu-stack` on the hung process. Two threads waiting on each other:

The thread `report_fatal_and_exit` spawns:

```
#4  rfd::backend::gtk3::utils::GtkGlobalThread::run_blocking
#5  <rfd::backend::gtk3::message_dialog::MessageDialogImpl>::show
#6  rfd::message_dialog::MessageDialog::show
#7  waveflow_lib::report_fatal_and_exit::{closure}
```

...blocked on a condvar, waiting for rfd's own global GTK thread:

```
#1  g_cond_wait
#2  g_main_context_wait_internal
#3  g_main_context_iterate_unlocked
#4  g_main_context_iteration
#5  gtk_main_iteration
#6  rfd::backend::gtk3::utils::GtkGlobalThread::new::{closure}
```

...which is waiting for the **default GLib main context**. That context is already held by the main thread — visible in its own stack:

```
#3  __pthread_clockjoin_ex                     <- the .join() in report_fatal_and_exit
#7  waveflow_lib::report_fatal_and_exit
#11 tauri::app::setup
#16 glib::MainContext::with_thread_default     <- tao owns the default context here
#18 tao...EventLoop::run
```

`rfd`'s gtk3 backend routes every dialog through a dedicated global thread that iterates the default main context. `tao` acquired that context on the main thread, and the main thread is parked in `.join()` waiting for the dialog. Neither can proceed.

This is not the Windows failure mode with a different shape — under Windows the dialog window was at least created (`WS_VISIBLE` clear). Here nothing is created at all.

## Control

The **same** `rfd` call, same version and features, same Xvfb, in a standalone binary — spawned thread included — displays correctly (465x118, `IsViewable`, right title). So this is neither the environment, nor GTK, nor `rfd`. It is the interaction with the running `tao` event loop.

## Suggested direction

Move the decision out of `setup`: run the guard **before `tauri::Builder::run`**. No GLib main context is held at that point, so `rfd` works on all three platforms and the three Windows-specific workarounds in the `report_fatal_and_exit` doc-comment stop being necessary.

The paths resolve early without an `AppHandle` — `app_data_dir()` is `dirs::data_dir()/app.waveflow`, and `AppPaths::from_handle` only needs the handle for that plus the bundled-plugin resource dir.

## Release note

PR #486 (`chore(main): release 1.8.0`) would ship the current behaviour. On Linux that trades a crash for a hang.

## Not covered

- No window manager on the test display. It cannot explain a `g_cond_wait` deadlock — that is independent of X, and the standalone control displays fine in the same environment — but a real desktop session is still unverified.
- macOS unverified.
- Only the `DbScope::Profile` path was exercised; `app.db` goes through the same `report_fatal_and_exit`.

InstaZDLL:
Closes #529.

#528 shipped the detection and the sentence. The detection was platform-independent, as claimed; the sentence was not. On Linux it never reached the screen and left the process hung on a frozen splash — worse, for a Linux user, than the crash it replaced.

## Why it hung

`setup` runs from *inside* the event loop, so the main thread was parked in `join()` while `tao` held the default GLib main context. That is the same context `rfd`'s GTK backend needs: it hands every dialog to a global thread that iterates it. Two threads waiting on each other, nothing mapped, no exit. The `eu-stack` trace is in #529.

Reading `rfd` 0.16's sources for the third platform says macOS deadlocks the same way, and would have even without a display: `run_on_main` dispatches to a main queue nobody is draining when called off-thread, and panics outright when `NSApplication` isn't running yet. So the spawned thread that fixes Windows breaks the other two.

## What changed

The explanation moved ahead of `tauri::Builder::run` — no main context held, no window created, and therefore no splash to hide either, which was the other half of what made the old shape fragile.

- **`schema_guard::preflight`** vets exactly the two databases startup is about to open: `app.db`, then the profile `resolve_target_profile` picks. Not every profile on disk — one left behind by a newer build is no reason to ground a launch that won't touch it.
- **`resolve_target_profile` is now a free function** taking `&SqlitePool`, so the pre-flight and `bootstrap` ask the same question. If they drifted, the pre-flight would vet one database and the app would open another.
- **It creates nothing** (`create_if_missing(false)`): a pre-flight that materialized `app.db` would turn "no install yet" into "install with an empty database" and the first-run path would never run again.
- **It defers on anything short of a verdict.** No database, a file it can't read, a query that fails → normal startup, where the guards inside the real opens are still the authority. A pre-flight refusing on its own uncertainty would be a new way to brick a healthy install.
- **The `setup` guard stays as a backstop and may only exit**, never present. It's reachable when the pre-flight couldn't read what startup then could.
- **Headless is handled explicitly.** `rfd` has no way to report a failed `gtk_init_check` — `run_blocking` parks on a condvar nobody will notify — so the dialog is skipped when neither `DISPLAY` nor `WAYLAND_DISPLAY` is set. Otherwise the fix would reintroduce the hang everywhere GTK can't start.

`AppPaths::from_root` / `root_for_identifier` are the plumbing: the pre-flight has no `AppHandle` to resolve against, and the identifier is read from the generated context so `tauri.conf.json` stays the single source of truth (`app_data_dir()` is `dirs::data_dir()/<identifier>`, per `tauri::path::PathResolver`).

## Verified

- `cargo test -p waveflow --lib` — **348 passed**, including 6 new pre-flight tests against the real migrations: a future `app.db`, a future profile `data.db`, a profile this launch will *not* open (must not ground it) versus the one it will (must), an unreadable file (must defer), and a fresh install left untouched on disk.
- `cargo check --workspace --all-targets`, `bun run typecheck`, clippy — clean, no new warnings.
- **The fatal path, headless, on Fedora 44**: forged future migration → guard fires → `no display server; the error above is the whole story` → **exit 1 in under a second**. No hang, which is the regression this closes.

## Watched on a real desktop — Linux and macOS

The parked check is done — Fedora 44, GNOME on Wayland, against a throwaway `XDG_DATA_HOME` so no real library was opened:

1. **Fresh profile boots normally**, zero `ERROR` lines. The guard costs a healthy install nothing.
2. **Forged future migration → the dialog appears on screen**, confirmed visually by the maintainer. This is the half that headless could not establish.
3. **The process blocks on it** rather than exiting: killed by `timeout` at 25 s with 124 while the dialog was still up. Dismissing it exits instead.
4. **The fatal line reaches the log file**, for the runs that ended through `std::process::exit(1)` — the flush fix, checked on the path it exists for rather than in isolation.

One note for whoever tests this next: `wmctrl` reports nothing here, and that is not a failure. It speaks X11/EWMH only, while the GTK dialog on Wayland is a native surface it cannot see. Machine-side window proof needs a Wayland-aware tool or a human.

GTK renders `set_title` as both the window title and the bold heading, so the sentence appears twice — cosmetic, and rfd gives no way to separate them. The toolkit also localises the button ("Valider"), while our text stays English for the reason above.

### macOS, no longer inferred

Same protocol on macOS 26.6.1 (arm64), against a throwaway bundle identifier since `XDG_DATA_HOME` means nothing there:

1. Fresh profile boots in 7 s, zero `ERROR` lines.
2. Forged future migration → **the dialog appears**, confirmed visually. `lsappinfo` shows the process checked in with the window server (`type="UIElement"`), so it reached the GUI session rather than dying headless.
3. **No panic.** This is the claim that mattered: the failure this PR was written to avoid was read out of `rfd`'s source and never run. It does not happen through the pre-flight path.
4. Process blocks on the dialog and exits ~1 s after dismissal, with the fatal line in the log file.

macOS renders it as a plain alert — no title bar, so no duplicated heading. That confirms the doubling is GTK's handling of `set_title`, not our text.

The docs were corrected before this run and now need a second pass in the other direction: what they call inferred for macOS is measured for the *fix*. Still genuinely unverified is the **old** shape's failure on macOS — nobody has run a dialog from inside `setup` there, and nothing in this branch does.

## Testing caveat worth recording

The throwaway-identifier sandbox does **not** isolate logs on macOS: `logging.rs` hardcodes `APP_IDENTIFIER = "app.waveflow"` instead of reading the Tauri identifier, so the run appended two lines to the real `~/Library/Logs/app.waveflow/`. Harmless, and not a product bug — the identifier never changes in production — but the isolation is weaker there than on Linux, where the path derives from `XDG_DATA_HOME` and the sandbox was complete.

## Summary by CodeRabbit

  * L’application vérifie les bases de données avant le démarrage.
  * Les bases créées avec une version plus récente affichent une boîte de dialogue d’erreur, puis l’application se ferme proprement.
  * Le profil actif est identifié automatiquement lors de cette vérification.

* **Corrections**
  * Les vérifications couvrent les bases de l’application et des profils, y compris lors de l’importation d’archives.
  * Les installations vierges et les bases illisibles continuent de fonctionner comme prévu.
  * Les journaux sont désormais correctement vidés à la fermeture de l’application.


InstaZDLL:
@coderabbitai Les trois findings sont valides, tous corrigés dans `18abf828`.

**1. `journal_mode` dans le pré-vol — corrigé, et plus sérieux que « trivial ».** La source de sqlx est explicite : *« Don't set `journal_mode` unless the user requested it. WAL mode is a permanent setting for created databases and changing into or out of it requires an exclusive lock that can't be waited on with `sqlite3_busy_timeout()`. »* Donc l'appel n'était pas seulement inutile : il faisait **écrire** la passe dont le rôle entier est de regarder sans toucher, et lui faisait prendre un verrou exclusif non temporisable — précisément le mauvais comportement pour un pré-vol qui doit s'abstenir en cas de doute. L'ouverture réelle du démarrage continue de poser WAL, à sa place. J'ai laissé un commentaire expliquant aussi pourquoi on ne passe pas en `read_only` pour autant : une base laissée avec un WAL chaud par un crash a besoin d'une récupération à l'ouverture, et une connexion read-only échouerait sur exactement les installations les plus en difficulté.

**2. Vidage du journal avant `exit` — corrigé aux deux sites.** Le garde du writer non-bloquant passe en état de module avec un `logging::flush()` explicite. `run` conserve un `FlushOnDrop`, donc le retour ordinaire **et** un dépliage de panique continuent de vider tout seuls — la propriété d'aujourd'hui n'est pas perdue au passage. Le site à l'intérieur de `setup` est celui qui comptait le plus : il n'affiche rien à l'utilisateur, le fichier de log y est le seul témoignage.

**3. Affirmation macOS — corrigée, et à deux endroits de plus que celui signalé.** Vous avez raison, c'est déduit des sources de `rfd` 0.16, jamais exécuté, et il n'y a aucun job macOS en CI. Au-delà de `storage.md`, la même surclaim était dans `docs/architecture/invariants.md` — dont le « Measured, one failure per platform » couvrait les trois OS — et dans le commentaire de `report_fatal_and_exit`. Les trois disent maintenant ce qui est mesuré (Linux #529, Windows #528) et ce qui est inféré.

CI relancée. La vérification visuelle de la boîte de dialogue sous Linux reste à faire, elle est signalée comme telle dans le corps de la PR.

PR #531, opened 2026-08-21
feat(scanner): play AIFF, and say why the other formats are absent

InstaZDLL:
A user asked for Opus. Opus isn't the cheap one — **AIFF was**, and the comment that lumped the two together is what kept it closed.

## Why AIFF was one flag away

AIFF is not a codec. It's Apple's container for the same PCM `wav` already carries, and symphonia reads it from `symphonia-format-riff` — the crate the `wav` feature already pulls in:

```toml
aiff = ["dep:symphonia-format-riff", "symphonia-format-riff/aiff"]
wav  = ["dep:symphonia-format-riff", "symphonia-format-riff/wav"]
```

So enabling `aiff` adds a sub-feature of a crate already compiled into the binary. **`Cargo.lock` is untouched** — no new dependency, no new licence surface, no Flatpak source regeneration. `lofty` already mapped `FileType::Aiff` in `extract.rs`, so the scanner was waiting on the extension gate and nothing else.

`.aifc` stays out deliberately. symphonia's AIFC path handles the PCM-shaped compression types (`none`, `sowt`, `twos`, `fl32`, `alaw`, …) and returns `unsupported` for the rest, so accepting the extension would index files that cannot play — the failure this list exists to prevent.

## The comment was the actual bug

The old text read:

> Opus / WMA / AIFF are intentionally absent — symphonia doesn't ship a mainline decoder for Opus, WMA is Microsoft proprietary, and AIFF isn't in the default feature set.

Three formats, one sentence, and a reader comes away thinking the answer is licensing. It isn't:

| | Why it's absent |
| --- | --- |
| **Opus** | Purely technical — symphonia ships no decoder. It's an IETF standard (RFC 6716), royalty-free by design, `libopus` is BSD. The opposite of proprietary. |
| **WMA** | Genuinely proprietary. Stays out. |
| **AIFF** | One feature flag. |

Calling Opus proprietary next to WMA is the kind of sentence that keeps a question shut for two years.

## The gap this does not close

The comment now records it rather than leaving it to be rediscovered: `.ogg` is accepted for Vorbis, an extension cannot tell Vorbis from Opus inside the container, so **an Opus file named `.ogg` is indexed here and fails at play time** — which is exactly what the user hit. Closing that needs the codec read at scan, not a longer extension list. Separate change, and it's the one that fixes their actual complaint.

## DLNA

Indexing a format and then serving it as `application/octet-stream` is half a feature, so `audio/x-aiff` is added to both MIME tables and to the `ProtocolInfo` advertisement. `x-aiff` rather than `aiff` because it's what Apple emits and what renderers match on.

Note the two `mime_for_path` functions in `dlna/cds.rs` and `dlna/http.rs` are byte-identical and had to be edited in lockstep. Not folded here — worth its own change.

## Test

`an_aiff_file_probes_and_decodes` builds a real AIFF byte by byte — `FORM`/`COMM`/`SSND`, 16-bit big-endian stereo, 44100 Hz as an 80-bit extended float — and decodes it through the same `symphonia::default::get_probe()` path the app uses, asserting every frame written comes back out. A ramp rather than silence, so a decoder returning the right count of zeroes wouldn't pass.

Remove the feature and it fails with `Unsupported("core (probe): no suitable format reader found")` — which is what makes it a test of the flag rather than of a string in a list.

`the_extension_list_admits_aiff_and_still_refuses_what_cannot_play` pins the policy at its edges, `.aifc` and `.opus` included.

## Checks

`cargo check --workspace --all-targets`, `cargo test --workspace` (all green), and the `--no-default-features` opt-out build, matching what `ci.yml` runs. Frontend untouched.

Two things I found and deliberately did **not** touch, both pre-existing and neither blocking: `cargo clippy` reports 7 `useless_borrows_in_formatting` errors on current stable in code this PR doesn't go near (CI doesn't run clippy), and `cargo fmt --all` on current stable reformats ~40 unrelated files. Both look like the toolchain having moved under an unpinned `toolchain: stable`.

## Summary by CodeRabbit

  * Prise en charge de la lecture et de l’analyse des fichiers audio AIFF.
  * Les extensions `.aiff` et `.aif` sont désormais reconnues, tandis que `.aifc` reste exclue.
  * La diffusion audio AIFF via DLNA est maintenant annoncée et prise en charge.

* **Tests**
  * Ajout de vérifications couvrant le décodage, l’analyse et la reconnaissance des fichiers AIFF.

### DIFF
--- a/src-tauri/crates/app/Cargo.toml
+++ b/src-tauri/crates/app/Cargo.toml
@@ -138,6 +138,10 @@ symphonia = { version = "0.6", default-features = false, features = [
     "mp3",
     "flac",
     "wav",
+    # Same crate as `wav` — `symphonia-format-riff`, already in the graph —
+    # under a different sub-feature. AIFF is Apple's container for the PCM
+    # that `wav` already decodes, so this adds a format, not a dependency.
+    "aiff",
     "vorbis",
     "ogg",
     "aac",
--- a/src-tauri/crates/app/src/dlna/cds.rs
+++ b/src-tauri/crates/app/src/dlna/cds.rs
@@ -497,6 +497,9 @@ fn mime_for_path(path: &str) -> &'static str {
         Some("mp3") => "audio/mpeg",
         Some("flac") => "audio/flac",
         Some("wav") => "audio/wav",
+        // `x-aiff` rather than `aiff`: it is what Apple emits and what
+        // renderers actually match on.
+        Some("aiff" | "aif") => "audio/x-aiff",
         Some("ogg" | "oga") => "audio/ogg",
         Some("m4a" | "mp4" | "aac") => "audio/mp4",
         _ => "application/octet-stream",
--- a/src-tauri/crates/app/src/dlna/http.rs
+++ b/src-tauri/crates/app/src/dlna/http.rs
@@ -69,7 +69,7 @@ pub fn router(ctx: ServerCtx) -> Router {
 <s:Envelope xmlns:s="http://schemas.xmlsoap.org/soap/envelope/" s:encodingStyle="http://schemas.xmlsoap.org/soap/encoding/">
   <s:Body>
     <u:GetProtocolInfoResponse xmlns:u="urn:schemas-upnp-org:service:ConnectionManager:1">
-      <Source>http-get:*:audio/mpeg:*,http-get:*:audio/flac:*,http-get:*:audio/wav:*,http-get:*:audio/ogg:*,http-get:*:audio/mp4:*</Source>
+      <Source>http-get:*:audio/mpeg:*,http-get:*:audio/flac:*,http-get:*:audio/wav:*,http-get:*:audio/x-aiff:*,http-get:*:audio/ogg:*,http-get:*:audio/mp4:*</Source>
       <Sink></Sink>
     </u:GetProtocolInfoResponse>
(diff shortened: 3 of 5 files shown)


InstaZDLL:
Rebased onto `main` after #533 and #534.

Two things happened in the rebase, one mechanical and one not.

**Mechanical.** #533 added its own helpers and test to the same place in
`extract.rs`'s test module. Both sets are kept —
`an_aiff_file_probes_and_decodes`,
`an_ogg_is_judged_by_its_stream_and_not_by_its_extension` and
`the_extension_list_admits_aiff_and_still_refuses_what_cannot_play` now
sit side by side and all three pass.

**Not mechanical.** The doc comment this PR writes described the `.ogg`
gap as still open — *"an Opus file named `.ogg` is indexed here and fails
at play time. Closing that needs the codec read at scan, not a longer
list."* #533 made exactly that codec read. Left alone, this PR would have
shipped a comment contradicting `is_scannable_audio`, two blocks below
it. The bullet now points at that function instead, and the `.aifc`
paragraph no longer refers to "the `.ogg` problem above", which no longer
exists.

`cargo fmt --check`, `cargo clippy -D warnings` and the core test suite
pass on the rebased branch — the first two being the gates #534 just
added.

PR #532, opened 2026-08-22
ci(release): codesign the macos bundle so tcc grants persist

InstaZDLL:
## Summary

- **macOS re-asks for folder permission on every launch.** macOS keys a TCC grant (the *"WaveFlow would like to access files in your Downloads folder"* consent) to the app's designated requirement. The shipped bundle only ever carried the ad-hoc signature the linker adds automatically on Apple Silicon — `Identifier=waveflow-36c95c4d36c8f6a6`, `flags=0x20002(adhoc,linker-signed)`, `Info.plist=not bound`, `Sealed Resources=none`, and no `Contents/_CodeSignature` at all. With no sealed bundle and no stable identifier there is nothing to anchor the grant to, so it never persists. Anyone whose library lives under `~/Downloads`, `~/Desktop`, `~/Documents`, an external drive or a NAS share re-consents at every single launch.
- **Sign the bundle in the release job, with graceful degradation.** With `APPLE_CERTIFICATE` + friends set it is a real Developer ID signature plus notarization — stable across versions, and it clears Gatekeeper too. Without them it falls back to ad-hoc, which still produces sealed resources, a bound `Info.plist` and the `app.waveflow` identifier: enough for the grant to survive between launches, though it resets on each update. So this lands useful *today* and gets better the day a cert exists, with no further code change.
- **Signing runs inside `tauri build`, not as a post-build `codesign` pass.** The bundler builds the DMG and the `.app.tar.gz` updater payload *from* the `.app`, so re-signing afterwards would leave both wrapping an unsigned copy.

Supporting pieces:

- `Info.plist` next to `tauri.conf.json` (auto-merged by the bundler) carries the `NS*UsageDescription` strings, so the consent dialog explains itself instead of showing the generic prompt.
- `waveflow.entitlements` declares `com.apple.security.cs.allow-jit`. Hardened runtime defaults to on and is mandatory for notarization; without the entitlement the wasmtime plugin host can't map its Cranelift output and crashes on the first plugin load (Web Radio ships in the bundle, so that's every user).
- A `Verify macOS signature` step asserts sealed resources, the `app.waveflow` identifier and a satisfied designated requirement. A silent signing failure would otherwise look like a clean release and quietly reintroduce the per-launch prompt.
- Partial credential sets fail the build instead of silently skipping signing or notarization.

## How I tested

The signing path itself can't be exercised without pushing a tag, so I validated the mechanism directly against a real bundle:

- **Reproduced the root cause** on the installed `/Applications/WaveFlow.app` (v1.7.0): `codesign -dv` reports `adhoc,linker-signed`, `Sealed Resources=none`, and `Contents/_CodeSignature` does not exist. The affected profile had `library_folder` = `/Users/…/Downloads/IM NAYEON`, which is exactly the TCC-protected case.
- **Proved ad-hoc signing is sufficient for the TCC half.** Applied the exact invocation the bundler will run — `codesign --force --sign - --identifier app.waveflow --options runtime --entitlements waveflow.entitlements` — to a copy of that bundle. Result: `Identifier=app.waveflow`, `flags=0x10002(adhoc,runtime)`, `Info.plist entries=16`, `Sealed Resources version=2`, entitlements embedded, and `codesign --verify --strict` → *"satisfies its Designated Requirement"*. Hardened runtime composes fine with an ad-hoc identity.
- **Ran the verify step's logic** against both the signed copy (passes) and an unrelated signed app (correctly rejected on the identifier assertion), so the guard fails closed rather than rubber-stamping.
- **Confirmed the config keys against the CLI actually pinned here** (`@tauri-apps/cli` 2.11.4) rather than from memory: `bundle.macOS.{entitlements,hardenedRuntime,signingIdentity}` exist, `hardenedRuntime` already defaults to `true`, the bundler auto-merges an `Info.plist` sitting next to the config, and the env it reads is `APPLE_CERTIFICATE` / `APPLE_CERTIFICATE_PASSWORD` / `APPLE_SIGNING_IDENTITY` / `APPLE_ID` / `APPLE_PASSWORD` / `APPLE_TEAM_ID`.
- `bun run lint`, `bun run typecheck`, `cargo check -p waveflow` (which re-runs `tauri-build`, so it validates the edited `tauri.conf.json`) — all pass. YAML and both plists lint clean.

## Checklist

- [x] Title uses Conventional Commits (`type(scope): subject`, kebab-case scope)
- [x] `bun run lint` + `bun run typecheck` pass locally
- [x] `cargo check --manifest-path src-tauri/Cargo.toml --all-targets` passes locally
- [x] If UI strings changed: every locale in `src/i18n/locales/` updated — n/a, the `NS*UsageDescription` strings are bundle metadata, not i18next keys (localizing them needs `.lproj/InfoPlist.strings`, which the bundler doesn't generate; noted in the file header)
- [x] If cross-cutting pattern changed: `CLAUDE.md` and `docs/` updated — new invariant line in `CLAUDE.md`, new *macOS signing and the TCC prompt* section in `docs/RELEASING.md`, plus the secrets table and the two now-stale "not Apple-Developer-signed" passages
- [x] Breaking change? No

## Follow-up (not in this PR)

Buying the Apple Developer account is planned for end of month. Once the cert exists, the only work left is pushing the six `APPLE_*` secrets — the workflow switches modes on its own. The secrets table in `docs/RELEASING.md` documents exactly where each value comes from.

Worth knowing: ad-hoc identity changes with every build, so **the grant still resets on update** until the Developer ID lands. Users whose library sits outside the protected folders (e.g. `~/Music`) are unaffected either way.

## Summary by CodeRabbit

  * Ajout de la signature et de la notarisation des applications macOS, avec signature ad hoc par défaut.
  * Ajout des autorisations nécessaires pour l’accès aux fichiers, volumes et réseaux locaux.
  * Conservation des autorisations système lors de la distribution de l’application.
  * Configuration des entitlements requis pour assurer la compatibilité avec le runtime de l’application.

* **Documentation**
  * Mise à jour des instructions de publication, de signature et de configuration des secrets macOS.
  * Clarification des contrôles de validation des versions macOS, notamment la signature, les ressources et Gatekeeper.


InstaZDLL:
@coderabbitai Finding vérifié contre la branche et fondé — corrigé dans `9daa8c8b`.

La validation était **à l'intérieur** du `if [ -n "$APPLE_CERTIFICATE" ]`, donc le certificat décidait si le reste était inspecté. Deux configurations à moitié faites tombaient dans le repli ad-hoc et partaient en release : des identifiants de notarisation sans certificat, et une identité + un mot de passe sans certificat. Dans les deux cas l'opérateur demande sans ambiguïté une release signée, et il en obtenait une non signée derrière un warning.

Un détail qui renforce le finding : **`docs/RELEASING.md` promettait déjà l'inverse** — « providing only part of either group fails the build on purpose, rather than silently shipping a half-signed bundle ». Le code ne le faisait simplement pas. J'ai aussi ajouté à la doc le cas croisé qu'elle ne nommait pas non plus : la notarisation exige le certificat, faute de quoi il n'y a rien de signé à soumettre.

Vérifié en table plutôt qu'à la lecture — le corps de l'étape extrait et exécuté sur sept configurations, contre les deux versions :

| Configuration | Avant | Après |
|---|---|---|
| aucun secret | exit 0, adhoc | exit 0, adhoc |
| Developer ID complet + notarisation complète | exit 0, developer-id, notarize | idem |
| Developer ID complet, sans notarisation | exit 0, developer-id + warning | idem |
| **notarisation sans certificat** | **exit 0, adhoc** | **exit 1** |
| **identity + password sans certificat** | **exit 0, adhoc** | **exit 1** |
| certificat sans password | exit 1 | exit 1 |
| notarisation partielle + Developer ID | exit 1 | exit 1 |

Exactement deux verdicts bougent, et ce sont les deux silencieux. Le chemin « aucun secret » — celui que le dépôt emprunte aujourd'hui — est inchangé, ce qui compte : il ne fallait pas transformer un durcissement en blocage de release.

InstaZDLL:
@coderabbitai review

Issue #406, opened 2026-07-19
Deezer/Last.fm similar artists come back empty for artists that previously had a list


InstaZDLL:
Split out of #403 (reported by @jo-el414 on `1.6.4-beta.1`), which bundled two unrelated problems.

## Symptom

> first many artists that did have info befor, now no info come up at all (sample image 1 - abba). second the ones that do fetch not all images view (use to view before)

Two parts: the similar-artists list comes back **empty** for artists that previously had one (ABBA among them), and where a list does come back, some entries render without a picture.

He suspects the recent Deezer artist-name work.

## Hypotheses tested and ruled out

Both of the obvious suspects were checked against the code and neither holds:

**1. The fuzzy name matcher (`cdc8d6ac`, `fix(metadata): reuse fuzzy name match for Deezer artist enrichment`).** This is the change he points at. The concern would be `select_by_name` comparing a *normalised* candidate against a *raw* search term, which would drop nearly everything. It doesn't — both call sites normalise before passing:

- `deezer.rs`: `select_by_name(hits, &normalize_name(&artist_name), …)`
- `theaudiodb.rs`: `let searched = normalize_name(name);`

"ABBA" normalises to "abba" on both sides and matches exactly. Not this.

**2. The shared picture resolver (`983ec2a9`).** Scoped to `fetch_custom_similar` — the *manually curated* similar-artists path from #323. A user with no curated overrides never reaches it, so it cannot explain a general regression.

## Leading hypothesis: Last.fm key, surfacing late because of the cache

[`get_similar_artists`](../blob/main/src-tauri/crates/app/src/commands/similar.rs) reads the Last.fm API key and returns an **empty list** when there isn't a usable one. Similar artists are a Last.fm feature; Deezer only backfills the pictures afterwards.

What makes this fit the "it worked before" framing: results are cached in `app.lastfm_similar` with an `expires_at`. Artists he had already viewed kept serving from cache long after a key stopped working. Once each entry's TTL lapsed, the refresh needed the key again and silently produced nothing — so the breakage surfaces gradually and lines up with whatever update happened to be installed at the time, without being caused by it.

This is a hypothesis, not a diagnosis. Asked him to confirm the key under **Settings → Integrations → Last.fm**.

## The missing pictures are probably a separate, known limitation

[...]

PR #539, opened 2026-08-23
fix: the tag editor and audio-path defects found in the audit

InstaZDLL:
An audit of the tag-writing and audio paths turned up ten shipped defects. Three of them destroy user data, and no rescan recovers any of the three, because the value is gone from the file as well as from the database. This is all of them.

One commit per defect, each with its own reasoning in the message.

## Data loss

**Saving a track's properties erased its genre.** The dialog sends every field on save, and its genre input opened empty behind a comment promising to surface the real value later. The backend read a present-but-empty genre as "the user cleared it" — `remove_genre()` on the file, `DELETE FROM track_genre` in the row. Opening a track's properties and fixing a typo in the title took its genre off the disk. The genre lives in `track_genre` rather than on the `Track` row, so it now takes its own `get_track_genres` call; until that lands the field is omitted from the payload entirely rather than sent empty.

**Changing a cover deleted every other embedded image.** `update_track_cover` cleared the picture list before pushing the new one — the comment said this was to clean up ID3 cover spam, and what it did was throw away the booklet, the back cover and the artist shot of any release that shipped them. Only `CoverFront` and the untyped `Other` are removed now.

**Editing a FLAC dropped its non-standard Vorbis comments.** `lofty::read_from_path` hands back generic `Tag`s built by splitting each concrete tag; `Id3v2Tag` keeps the remainder in a companion slot and restores it, `VorbisComments` does not. Every non-standard comment on a FLAC / Ogg / Opus / Speex file went away on the first save — including the `SYNCEDLYRICS` our own lyrics editor writes, so editing a FLAC's title erased its synced lyrics. All writes now go through one `patch_file` that reads the concrete file type and merges the remainder back.

## Silent wrongness

**The year never reached an ID3v2 file.** It was written to `ItemKey::Year`, which has no ID3v2 mapping at all, so it was dropped on the way to the file while `sync_db` recorded the new value. The dialog showed a year no other player could see, and the next rescan took it back. It rides on the date item now — which is what the scanner reads through `Accessor::date` — and a save that doesn't change the year leaves a full release date alone instead of truncating `2024-03-15` to `2024`.

**`Bit-perfect` was shown for streams going through the system mixer.** The pill checked every DSP stage and never checked whether the stream owned the device. `player_get_state` now carries `exclusive_active` (the engine already tracked it; it just never left the settings card), and a clean pipeline over a shared device gets its own line instead — "Sortie partagée (mixeur système)" — so the reason the green pill is absent is on screen rather than left to guess.

**DoP on Linux gave up on every desktop.** PipeWire and PulseAudio hold every card from login, so the raw `hw:` open returned `EBUSY`, we logged a `warn!` and fell back to DSD → PCM: the user turned the setting on, nothing happened, nothing said why. `device_reservation` takes the `org.freedesktop.ReserveDevice1.Audio<N>` name on the session bus — the protocol both servers watch in order to release a device — then retries the open while the server finishes letting go. Additive by construction: no session bus, a refused name, or a card that stays busy all land exactly where the code was before.

**A downgrade could still crash on launch.** #526 and #529 taught the app to refuse a database from a newer build and explain itself before the event loop starts, but only for the shape where the migration *version* is unknown. When two builds share a migration id whose SQL differs, sqlx answers `VersionMismatch` instead — and that one still panicked out of the Tauri `setup` hook, which is the crash-on-launch shape #526 set out to remove. `ensure_no_foreign_migration` gives it the same pre-flight, dialog and clean exit. Line-ending drift is deliberately excluded, since `migration_heal` repairs that a moment later.

**Deezer refuses with HTTP 200.** A rate-limited call answers `{"error":{…,"message":"Quota limit exceeded"}}` with a 200 status. Our types only knew `data`, so that body failed to deserialize and every caller logged it and carried on empty — a throttled client and an artist Deezer had never heard of were indistinguishable, which is the fog #406 was diagnosed through.

**Editing tags on a DSD file failed with "unknown format".** lofty has no DSD reader, so `.dsf` / `.dff` are refused up front with a message that says so, instead of failing later in a way that reads like a corrupt file.

**A failed save looked like a successful one.** Every failure went to `console.error` and nothing else. The dialog shows it now — without which the DSD refusal above would have been invisible for the same reason.

## Verification

The app crate's tests don't run on Windows (the test binary links Tauri and dies with `STATUS_ENTRYPOINT_NOT_FOUND`), so three throwaway crates carried the parts that needed executing rather than compiling:

- **Tag writes** — WAV/ID3v2 and FLAC files synthesized on the fly. Confirms what the old path did (the FLAC custom comment is gone; the cover write leaves exactly one picture; `ItemKey::Year` never reaches the file) and what the new one does (custom frames survive a fields edit *and* a cover change, back cover and booklet survive, the year reads back, a second edit still lands, DSD is refused by name).
- **The schema guard**, against the real profile migrations: a tampered checksum is refused, the CRLF form of the same SQL is not, an unknown version is left to the other guard — and sqlx does answer `VersionMismatch` on the row the guard catches, which is the contract it exists to get in front of.
- **The Linux-only code**, compiled and clippy-checked under WSL against zbus and ALSA. That is what caught `DBusProxy::new` and `release_name` returning two different error types, which would otherwise have reached CI unbuilt.

Deezer's new error arm is covered by ordinary unit tests in `waveflow-core` (a refusal, a genuinely empty result the error arm must not swallow, a populated one). `typecheck` / `lint` / `cargo clippy --workspace --all-targets -- -D warnings` / `cargo fmt --check` are clean.

A local CodeRabbit pass raised four findings; three are fixed in the last two commits (an out-of-range year that would have written 65535 to the file while the row kept the real value, a wasted 50 ms before the first open attempt, and a release reply that logged "released" for all three of its outcomes). The fourth — that nothing subscribes to `NameLost` — is answered in the commit message: its stated consequence doesn't occur, and what remains is a belief nothing reads.

## Not covered

The reservation is not tested against a real sound server — that needs a Linux box with a DAC, and it changes nothing on a machine where the card was free to begin with. `zbus` becomes a direct dependency on Linux, but not a new tree: notify-rust and two Tauri plugins already pull the same version, so the lockfile gains one edge and no packages.

## Summary by CodeRabbit

  * Affichage plus précis du mode bit-perfect et des sorties audio partagées.
  * Lecture exclusive ALSA améliorée sous Linux, avec nouvelle tentative après libération du périphérique.
  * Chargement séparé des genres et messages d’erreur détaillés lors de l’édition des morceaux.

* **Correctifs**
  * Préservation améliorée des métadonnées et des pochettes lors de la modification des tags.
  * Validation des années et refus explicite des fichiers DSD non modifiables.
  * Meilleure détection des erreurs Deezer, notamment les dépassements de quota.
  * Protection contre les migrations de base incompatibles.

* **Documentation**
  * Documentation mise à jour sur l’édition des tags, les migrations, Deezer, la lecture DoP et le mode bit-perfect.


InstaZDLL:
@coderabbitai all six are addressed in `bb4a9aa5`. Taking them in order:

**Multiple genres merging on save** — valid, and the better of the two options you offered. The genre now rides along only when the user actually changed the field, so a track that legitimately carries two genres and is saved untouched keeps both rows. Splitting on `"; "` backend-side was the other candidate and I passed on it: the scanner stores the raw tag string as one genre, so the editor and a rescan would then disagree about the same file. Worth noting nothing in the app writes two rows today — but that is an assumption, not a guarantee, and the dirty check removes the need for it.

**Genre typed while the fetch is failing** — valid, and worse than it looked, because the save reported success. The field is disabled until `genres` resolves: a box we cannot compare against is a box we will not send.

**Bit-perfect with an unknown source rate** — valid. `isResampling` and `isDownmixing` both answer "no" when the source format is unknown, so the verdict could be reached without the comparison ever happening. It now requires a known source rate equal to the output rate, and known channels the output doesn't reduce, with `isDopNative` exempt since DoP ships at `dsd_rate / 16` and never matches nominally. No test for it: this repo doesn't ship frontend unit tests.

**HTTP status in `fetch`** — valid, and it defeated the point of the previous commit. A 429 from the edge has no JSON to read, so it reached `json()` and came back as exactly the decode error the new arm exists to stop producing. Status is checked first now: `RateLimited` for 429, `error_for_status()` for other non-2xx, and `is_quota_exceeded` covers both arms.

**Untagged variant precedence** — added. A body carrying both `data` and `error` reads as the refusal, which is the precedence we want, and a reorder now fails a test instead of flipping it quietly.

**`Общий вывод` → `Общий выход`** — applied; that is the word the file already uses for an audio output.

On the doc paragraph: you're right that no caller consumes `is_quota_exceeded`, and the paragraph read as though app behaviour had changed. It hasn't — every enrichment path still degrades to an empty result by design, since a missing picture must not fail an album page. What changed is that the reason reaches the log the enrichment paths already write. The paragraph now says that, and says the classifier is there for a caller that wants to back off, with none doing so yet.

One unrelated note for anyone reading: **`Flatpak sources` is red on `main`**, not from this branch. `14fac7de chore(deps): update dependencies` bumped the lockfiles without regenerating `packaging/flatpak/generated/`, leaving 90 crates uncovered and eight npm specs adrift. This PR adds no package — one dependency edge to a `zbus` already in the lock — and the check fails identically on `main` at `14fac7de`. It needs the `flatpak-sources-refresh` workflow, separately.

InstaZDLL:
@coderabbitai both addressed in `03065940`.

**Genre normalisation** — applied. One correction on the reasoning: the divergence you describe doesn't occur, because the backend trims on both sides — `apply_patch` tests `g.trim().is_empty()` before `remove_genre()`, and `sync_db` tests the trimmed string before re-inserting, so a whitespace-only genre already behaved exactly like an empty one and the row and the file agreed. What the finding does catch is a save that rewrites the file to change nothing: "Rock " against a stored "Rock" counted as a change. Normalised once now, and used for the comparison, the payload and the local state alike.

**The DoP shortcut** — fully valid, and it was doing more than I intended. `isDopNative ||` waved through unknown and zero source values along with the rate comparison it was meant to excuse. Only the rate is exempt now, and only because a DoP stream rides at `dsd_rate / 16` so the nominal rates are never equal by construction. Channels also have to match outright rather than merely not shrink — you're right that an output wider than the source means the engine is routing or padding, which isn't "untouched" whatever the rest of the chain looks like.

InstaZDLL:
@coderabbitai valid, fixed in `79be05c8` — and it undercut the verdict the previous three commits were spent making honest, so thanks for catching it.

The effect had `[]` deps, and the popover is mounted for as long as the pointer rests on the footer. A track ending underneath it swapped `track` for the next one while every output-side field still described the previous stream, so the pill was decided by pairing new metadata with an old measurement. The sharp edge is `dopActive`: carried over from a DSD track it exempts the *next* track from the sample-rate comparison, which is exactly the exemption that lets something be called bit-perfect without one.

- Re-reads on `track.id`.
- Also re-reads on `player:audio-mode-changed`, because the output is rebuilt *after* the track changes (a DoP engage, a WASAPI exclusive re-open at the new native rate) — the track identity alone would sample too early. Every engine path that stores a new output mode emits it. The listener is attached before the first read, per this repo's "subscribe first, then snapshot" invariant, since a rebuild landing between the two is the one event we can't lose.
- The snapshot records the track it was read for, and one belonging to a different track counts as absent. Everything downstream already renders "loading" and withholds the verdict without a snapshot, which is the right thing to show between a track change and the read that follows it.

A monotonic token drops an older read resolving after a newer one — same guard as `deviceRefreshTokenRef` in `PlayerContext`, which exists for the same reason.

I did not route this through `usePlayer`'s `deviceSampleRate` / `deviceChannels`, which are already refreshed on `player:track-changed`: they don't carry `dop_active` or `exclusive_active`, so the popover would still need its own read for the two fields the verdict now turns on.

InstaZDLL:
@coderabbitai confirmed and fixed in `f137d444`.

Traced the sequence: the old effect's `await listen()` resolves after `trackId` changed, it releases its listener, then calls `read()` anyway — which bumps a token that is **per-component, not per-effect**, so it lands past the token of the read the new effect had already started. That newer read then fails its own `token !== readTokenRef.current` check and drops its result. No wrong value is displayed (the snapshot stays absent and the verdict is withheld), but the popover sits on "loading" until the next output rebuild or a fresh hover, which is not what it should do after a plain track change.

Guarded at the top of `read` rather than at the call site: that covers the tail above *and* any listener callback already in flight, with one check instead of one per caller.

InstaZDLL:
@coderabbitai valid — that's the hole on the other side of the stamp, fixed in `5d3aca17`.

The stamp records which track a measurement belongs to, but it recorded what the **prop** said rather than what the engine answered. A track advancing after a read starts returns the new stream's output, and the prop only catches up once `player:track-changed` has been through the context and a render; in that window `cancelled` is still false and the token still matches, so the read landed under the old id — one track's metadata paired with another's measurement.

I took it narrowly, though, and not quite as written: **only a positive disagreement counts**. Two reasons, both in `player_get_state`:

- `current_track` is `None` for a Web Radio session — negative sentinel id, no library row, and the comment there is explicit that the frontend is meant to hydrate radio through `get_current_radio_metadata` instead. Skipping on a null would leave the popover permanently on "loading" for every radio stream.
- It is also `None` when `restore_state` disagrees with `SharedPlayback.current_track_id` (the `_ => (None, …)` arm), which is likely during exactly the window this finding is about — so a strict equality gate would throw away good reads at the moment it matters most.

So the guard fires only when the engine *names* a different track, and the read the new effect fires for itself fills the gap otherwise. `typecheck` and `lint` clean.

InstaZDLL:
@coderabbitai skipping this one — the branch it asks for is unreachable, and it would weaken the guard rather than relax it.

`current_track.id` is **always positive when present**. Two filters, both in the backend:

- `queue::restore_state` opens with `if let Some(last_id) = … if last_id > 0` and then queries the `track` table, so it can only ever return a library row id.
- `player_get_state` branches on `SharedPlayback.current_track_id`: `> 0` yields `Some(t)` only when `t.id == active_id`; `< 0` (the radio / Spotify sentinel from `next_radio_track_id`, a negative monotonic counter) yields `None` **deliberately**, with a comment saying the frontend must hydrate radio through `get_current_radio_metadata` instead; `== 0` falls back to `restore_state`, filtered as above.

So there is no path that puts a negative id in `current_track`, and `reported > 0 &&` would be dead code. The negative-sentinel case is already covered by the `reported != null` half of the guard, which is exactly why it is written that way — the commit message on `5d3aca17` spells that out.

And if it ever did become reachable, letting a non-positive reported id through is the wrong direction: it would stamp a radio session's snapshot onto a library track, which is the pairing this guard exists to prevent.

One note on the framing: this guard skips storing a snapshot in a hover popover. Nothing in this file touches transport, so no value of it can cancel playback.

PR #550, opened 2026-08-26
feat(library): browse both sources from one album list

InstaZDLL:
Second half of lot 1. The catalogue mirror (#547) landed the server's albums locally and the cover cache (#549) made them paintable; nothing showed them. This does.

`list_library_albums` returns both halves as one sorted list tagged by `source`, and the source becomes a **filter inside** the list rather than a section beside it. That is the whole difference between a unified library and two tabs.

## Decision 1 is untouched

An album held on both sides appears **twice**, tagged twice. Unifying the navigation is not deduplicating the catalogue — that is reserved for its own RFC — and the chip on the cover is what says so.

## Two things only a run against a real database showed

- **The sort keys have to be normalised on both sides.** The local half sorts on `canonical_title` / `canonical_name`, forms produced by `normalize_name`: lowercased, diacritics folded, punctuation dropped. SQLite reproduces none of that — `COLLATE NOCASE` is ASCII-only — so sorting the remote half on its raw display name put "Björk" and "bjork" in two different places and **split one artist in half down the middle of the list**. `remote_album` now carries `sort_title` and `sort_artist` through the same function, written by the mirror. Rows from before those columns existed fall back to their display title, and one walk fills them in.
- **A local library filter has to exclude the remote half.** The picker chooses among *local* libraries and a server album belongs to none of them; leaving those rows visible while the user has narrowed to one library reads as the filter having failed.

## What a server album does not get

No playlist, no cover picker, no context menu — none of them can accept it. It opens the remote detail view. The filter hides itself when no server is bound: a filter whose second option is permanently empty is worse than no filter, and most profiles are local-only.

## Why the ordering is its own clause

`album_order_clause` sorts on the inner tables' own columns (`al.canonical_title`, `MIN(t.added_at)`), which do not exist outside the local half of a union. The new clause sorts on the columns the union projects, so both halves obey one comparison and one collation. A list sorted differently depending on where a row came from would be worse than two lists.

## Checks

`cargo fmt --check`, `typecheck`, `lint`, `cargo clippy --workspace --all-targets -D warnings` on the default feature set **and** `--features sync_v2`, `cargo test --workspace --features sync_v2` (388 + 226, zero failures) — including 2 new tests on the sort keys. The union query was exercised against a real SQLite database with all 33 profile migrations applied: every ordering, both filters, and the fallback for rows mirrored before the columns existed.

Localised across all 17 locales.

## Not in this PR

The Tracks, Artists and Playlists tabs on the same pattern, then retiring the three twin views. The pattern is proven end to end; the rest is replication.

## Summary by CodeRabbit

  - La bibliothèque regroupe les albums locaux et distants dans une vue unifiée, sans doublons masqués.
  - Filtrez par source : toutes, cet appareil ou serveur distant.
  - Triez les albums selon des critères communs et ouvrez directement un album distant.
  - Les illustrations distantes se chargent à la demande avec mise en cache et vignette de secours.
  - Les actions de modification restent réservées aux albums locaux.

- **Améliorations**
  - Les états vides et la navigation au clavier sont mieux pris en charge.

- **Traductions**
  - Le filtre de source est disponible dans les langues prises en charge.


InstaZDLL:
@coderabbitai review

PR #558, opened 2026-08-30
feat(playlist): one playlist page, not a poorer copy for the server's

InstaZDLL:
The last of the three twins from lot 1, and the only one where the copy was the **richer** view. `RemotePlaylistView` (1 140 lines) had an inline rename, a remove button on every row and an add-tracks panel backed by a live catalogue search — three affordances the local playlist has no equivalent for, because a local playlist renames through a modal that also sets a colour, an icon and a cover, and adds and removes through the track context menu.

So the absorption carries them over rather than levelling them off. `PlaylistView` takes either identifier; those three are remote-only the same way shuffle, M3U export and the edit modal are local-only.

## The one real asymmetry is a write, not a capability

The drag handle. Locally a playlist entry is named by its track's rowid — a track is in a playlist zero or one times — so a move travels as `(track, destination)`. The server keys entries by position and may hold the same track twice, so there is no id to name one by and the move travels as `(from, to)`.

The table already resolved a drag to a pair of indices, so one implementation drives both writes and the branch is a single call site. Dragging works on a duplicate because the projection hands each row a **distinct negative sentinel id per position** at load time, and rows are moved rather than re-projected — so the handle survives an optimistic reorder.

## What is absent, and why

| Absent for a server playlist | Reason |
| --- | --- |
| Shuffle | a mode of the local queue; the remote one has none |
| Export M3U | an M3U is a list of file paths |
| Edit modal (colour / icon / cover) | writes through a rowid; renames in place instead |
| Selection, context menu, ratings | all keyed by rowid |
| `added_at` / `filename` sort | a remote track carries neither field — dropped rather than shown inert |

A persisted sort mode is checked against the list *actually on screen*, not just against the union, so a server playlist cannot restore a mode its menu no longer offers.

What it gains is everything the local view had grown since the twin was written: the skeleton, the per-playlist sort memory (`profile_setting`), the virtualized table, the portalled drag overlay.

## Two defects the compiler could not see

- `pending_creation` was only ever written on the remote branch, so **“not sent to the server yet” survived a hop to a local playlist**.
- A remote playlist **refetched from the server whenever any local playlist's `updated_at` moved** — `playlistsSignature` cannot concern it.

The identity stamp from #555 is reused, and the effects that read the snapshot check it, not only the render — the correction #556 needed.

## i18n

`reorder` and `remove` move out of `remote.playlist.*` into `playlistView.actions.*` with their existing translations, which also retires a **hardcoded English `aria-label`** on the local row (`"Drag to reorder"`). Two new keys carry the empty and not-found advice a server playlist needs — the local wording points at a `+` button that is not there. The keys the twin alone used are dropped from all seventeen locales.

## Verification

Not by reasoning alone this time. A sandboxed `XDG_DATA_HOME` with a seeded projection — a playlist of five entries **holding one track twice** — rendered the full page, and both hard points behaved: the reorder moved the row that was grabbed, and the remove dropped the entry that was clicked, not its duplicate. Confirmed against the projection afterwards, with the matching `update_playlist` mutations queued.

## Known, and deliberate

A remote row loses the **hover ▶ that replaced the track number** in the twin. The local row does not have it and this unifies on the local row; double-click and the header Play cover the gesture. Restoring it would mean adding it to the local row too, which is a change to the local playlist and outside this PR.

`role="button"` on a row containing nested buttons (grip, heart, and now the remove ×) makes those descendants presentational for assistive tech — the defect #556 named. It is pre-existing on the local row and on `LibraryView`, and fixing it properly means restructuring the row; flagged rather than silently propagated.

---

`typecheck` / `lint` / `prettier` green. `AppLayout` carries one pre-existing Prettier drift (line 366, unrelated) left untouched.

## Summary by CodeRabbit

  * Les playlists distantes sont désormais intégrées aux vues principales des playlists.
  * Recherche et ajout de morceaux disponibles dans les playlists distantes.
  * Navigation vers les albums et artistes associés depuis les morceaux distants.
  * Réorganisation par glisser-déposer et synchronisation des favoris prises en charge lorsque compatibles.

* **Améliorations**
  * Messages dédiés pour les playlists distantes vides ou introuvables.
  * Actions de réorganisation et de retrait de morceaux ajoutées aux playlists locales.
  * Actions et libellés de playlist mis à jour dans les langues disponibles.


InstaZDLL:
@coderabbitai Good catch, and it was broader than the two lines you flagged in the sense that mattered: the sweep that retired `remote.playlist.trackCount` was written from the key set in `en.json`, which only has `_one` and `_other`. Russian carries `_few` and `_many`, Arabic adds `_two` — CLDR categories English has no equivalent of — so a literal key list could not see them.

Fixed in a3352f08 by stripping on the **base name** instead of the literal key, which is the form that survives a locale having plural categories the source language does not. I also scanned every locale in both directions afterwards — no other orphan under `remote.playlist.*` or `remote.common.*`, and no key used in code missing from any locale.

PR #559, opened 2026-08-30
feat(remote): ask the server to transcode, and never be refused for it

InstaZDLL:
**Lot 2 of the unified library, and it needed nothing from the server.** `/api/v2/stream/{ticket}` has always accepted `format` (`raw` / `mp3` / `opus`) and `bitrate` (32–512), and waveflow-server#143 added the announced ceilings. The desktop simply never sent the parameters, so every remote track arrived as the original file and no setting existed to change that.

## Shape

Per profile, **off by default**, **remote-only**.

Off by default because transcoding trades fidelity for bandwidth, which is not a trade to opt someone into silently. Remote-only because a local file is already on the machine — re-encoding it would cost quality and buy nothing.

The parameters belong to the *request*, not to the ticket, so the whole feature resolves at one seam (`remote::stream::ticket_url`) and the rest of the playback path is untouched: it receives a URL either way.

## The 429, and why it is avoided rather than handled

The server returns `429` at its transcode concurrency ceiling. **It cannot be handled where it arrives**: the refusal lands when the *decoder* opens the URL, on the audio thread, where there is nothing sensible to retry with and no layer that knows what a fallback would even be.

So it is not handled — it is *avoided*. `GET /api/v2/transcode/status` is read first (`available`, `active`, `global_limit`, `per_user_limit` — announced for exactly this purpose), and the original bytes are chosen **before** asking rather than the refusal being discovered after. A saturated server costs the user bandwidth instead of a track that will not play; a server built without FFmpeg says so in Settings instead of leaving the preference looking broken.

The pre-check is not a reservation — two clients can pass it at once and one still be refused. That residual case is now the rare failure rather than the common one, and it is written down rather than implied.

## Three defects, none of them visible to the compiler

| | Found by |
| --- | --- |
| Switching codec carried the bitrate over. 320 is roughly transparent for MP3 and wasteful for Opus — the two scales do not mean the same thing. | reading |
| The network probe shared a `try` with the visibility probe, so an **unreachable server hid the whole card** — exactly when someone is most likely to be setting it, before leaving the network. | reading |
| With no stored bitrate the hook fell back to a fixed 192 while the backend resolved the **current codec's** default (128 for Opus). The picker showed one number while the URL carried another. | the sandbox |

The third is the interesting one: the picker is the only place a user can see what is being asked for, so a divergence there is the interface lying about the request. Both sides now resolve identically — and it was invisible until the preference was written and read back from a real database.

## Verification

A sandboxed `XDG_DATA_HOME` with a seeded projection and the bound server **deliberately unreachable** — which is precisely the case that used to make the card disappear. Card renders, all three choices selectable, bitrate picker and the degradation notice appear on Opus, and the preference lands in `profile_setting`.

What is *not* verified: an actual transcode. That needs a reachable server with FFmpeg, which this machine does not have.

---

3 new unit tests (unknown format is "original" rather than a guessed codec; `Off` emits no query at all, since `format=raw` with a bitrate is a 400 server-side; "at the ceiling" is not headroom, and neither is "idle but unable"). 404 lib tests green, `cargo fmt` / `typecheck` / `lint` / `prettier` green. 12 keys × 17 locales. RFC-005 updated.

## Summary by CodeRabbit

  - Ajout du transcodage des morceaux distants en Original, Opus ou MP3.
  - Sélection du débit audio depuis les paramètres, désactivée par défaut.
  - Affichage de la disponibilité et de la capacité de transcodage du serveur.

- **Améliorations**
  - Retour automatique au fichier original lorsque le transcodage est indisponible ou saturé.
  - Les fichiers locaux ne sont pas concernés par le transcodage.

- **Traductions**
  - Ajout des libellés correspondants dans les langues prises en charge.


InstaZDLL:
@coderabbitai Three of five applied in 5532534b, two skipped with reasons.

**Applied.**

- **`lib.rs` cfg guard** — correct, and worth more than consistency. Every other `remote_auth` registration carries `#[cfg(feature = "sync_v2")]`; mine was the lone exception. The module is not itself gated so nothing failed to compile, but the command was invokable on a build with no remote support, where it can only answer "not signed in to a remote server". Verified both ways: `--no-default-features --features updater` and the default set.
- **`status()` offline check** — a genuine breach of the repo's process-wide offline invariant. `ticket_url` checked; `status` is *also* called straight from the settings card, so the offline switch had a way past it.
- **Italian `opusHint`** — applied. Both forms are grammatical, but the apocope reads better with a preceding article and there is none here.

**Skipped.**

- **16px select on mobile** — this is a Tauri desktop application; there is no mobile viewport and no iOS focus-zoom to avoid. Every other settings `<select>` in the codebase (e.g. `ArtistBioSourceCard`) is `text-sm`, so the change would make this one control typographically inconsistent to solve a problem the app cannot have.
- **Do not hide the card when `remoteGetStatus` fails** — the premise does not hold: `remote_get_status` is a local IPC command that reads `remote_binding` and `auth_credential` from SQLite. It performs no network I/O, so there is no transport failure to survive. Its throwing is precisely the self-hiding contract every remote surface uses — TypeScript cannot see a Cargo feature, so an unregistered command is how the UI learns `sync_v2` is absent. `CatalogueMirrorCard` and `RemoteServerCard` both do `catch { setVisible(false) }` for that reason; deviating would leave this card alone on screen in a build whose backend does not exist.

Worth noting the second one was already half-addressed before your review: the *capability* probe (`remoteTranscodeStatus`, which does hit the network) was split into its own `try` in an earlier commit, so a sleeping or unreachable server no longer hides the card — only an absent backend does.

PR #561, opened 2026-08-30
feat(remote): keep a server track on disk, and play one whose file is gone

InstaZDLL:
**Lot 4, both halves.**

## Keeping a track

A download is not a cache entry. The cache is filled without asking and evicted without asking; a download is *chosen*, and leaves only when its owner says so. It lands in a managed folder the scanner never sees.

It is deliberately **not** a local library track: no `track` row is created for it, which also means `remote_track_link` cannot describe it — that table keys on `local_track_id REFERENCES track(id)`. It is recorded as what it actually is, a remote track that happens to be on this disk (`remote_track_download`).

### That resolves a contradiction the plan was carrying

The arbitration called the managed folder *invisible to the scan* **and** called the reconciliation link *free*. Both cannot be true of the same row: a link needs a local track, and an invisible folder produces none.

What is genuinely free is the **hash**. The file is hashed in the same pass that writes it, so the server's own `full_hash` — not the library's `file_hash`, which covers a file the server has never seen — is in hand the moment a `track` row exists to attach it to. The link is paid *forward*, when someone imports into a scanned folder, not here.

### Always the original bytes

Whatever the transcode preference says. That preference exists to spend less bandwidth on a stream heard once; baking a lossy re-encode into a file someone chose to keep would make a bandwidth decision permanent, and silently.

Playback prefers, in order: **reconciled local file → download → cached stream → network.**

## Playing one whose file is gone

Server-to-local has always resolved. **Local-to-server never did** — a library track whose file had been moved or deleted just failed with `file not found` (`player.rs:1922`). With a *confirmed* link it now plays from the server instead.

Only confirmed. A stale link is a guess, and guessing which recording to substitute is worse than reporting the missing file.

### The decoder was answering two questions with one variable

It decided **"finite file or endless radio?"** from the *global remote-session state*. That is right for the remote queue and wrong for everyone else: a library track falling back has no session, so it would have been opened with ICY — forward-only, and parsing metadata blocks out of a FLAC.

Finiteness is a property of *this stream* and now travels on the command. Whether a remote session is running stays what it was, and still drives the metadata the bar shows. Separating them is what made the fallback possible at all, and it removes a coupling that would have broken for the next caller too.

## Verification

- The migration was applied to a **real profile database**, not a fixture, with `foreign_keys` on; both CHECK constraints were probed and both bite (`length(full_hash) = 64`, `size > 0`).
- 417 lib tests green.
- **Both feature configurations build clean** — the new path helpers are gated, since nothing without `sync_v2` can reach them.

## Not verified

A real download against a live server, and therefore the fallback end to end. The sandbox has a projection but no bytes and no reachable host. The path-safety and naming logic is unit-tested; the wiring is not.

## Surface

Per-playlist **"keep offline"** in the remote header — sequential on purpose, since the wait is the bytes rather than the round-trips, and one at a time is what lets the count mean anything while it runs. Settings reports and clears the copies, apart from the stream cache, because the two have opposite lifetimes.

i18n: 5 keys × 17 locales, plural sets taken from **each locale's own** categories with the script failing loudly on a missing body.

## Summary by CodeRabbit

  - Téléchargez des titres de playlists distantes pour les écouter hors connexion.
  - Suivez la progression, le nombre de titres conservés et l’espace utilisé.
  - Consultez, supprimez individuellement ou effacez tous les téléchargements.
  - Les téléchargements sont prioritaires à la lecture, y compris lorsque le fichier local lié est indisponible.
  - Les fichiers audio finis prennent en charge la navigation, tandis que les radios conservent leur fonctionnement habituel.

- **Internationalisation**
  - Nouveaux libellés disponibles dans toutes les langues prises en charge.

### DIFF
--- a/docs/rfcs/RFC-005-remote-source-and-sync-v2.md
+++ b/docs/rfcs/RFC-005-remote-source-and-sync-v2.md
@@ -456,6 +456,43 @@ cached file that will not decode falls back to the server once instead of
 failing that track forever. Offline, the ticket is simply not minted and the
 cached file plays alone, which is the point of having it.
 
+**A track can be kept**, which is a different thing from being cached. The
+cache is filled without asking and evicted without asking; a download is
+chosen, lands in a managed folder the scanner never sees, and disappears only
+when its owner says so. It is deliberately **not** a local library track: no
+`track` row is created for it, so `remote_track_link` cannot describe it either
+— that table keys on `local_track_id REFERENCES track(id)` — and a download is
+recorded as what it is, a *remote* track that happens to be on this disk.
+
+Downloads are always the **original** bytes, whatever the transcode preference
+says. That preference exists to spend less bandwidth on a stream heard once;
+baking a lossy re-encode into a file someone chose to keep would make a
+bandwidth decision permanent, and silently. The file is hashed while it is
+written, in the same pass that copies it, which is what makes the plan's "free"
+reconciliation proof real: the server's own `full_hash` — not the library's
+`file_hash`, which covers a file the server has never seen — is in hand the
+moment a `track` row exists to attach it to, with no re-read.
+
+Playback prefers, in order: a reconciled local file, a download, a cached
+stream, the network.
+
+**And the direction that never existed now does.** Server-to-local has always
+resolved — `preferred_local_playback` plays the local file when there is one —
+but local-to-server did not, so a library track whose file had been moved or
+deleted simply failed with `file not found`. When such a track carries a
+*confirmed* link, it now plays from the server instead. Only confirmed: a stale
+link is a guess, and guessing which recording to substitute is worse than
+saying the file is missing.
+
+Making that work required separating two questions the decoder had been
+answering with one variable. It decided "finite file or endless radio?" from
+the **global remote-session state**, which is right for the remote queue and
+wrong for anyone else: a library track falling back has no session, and would
+have been opened with ICY — forward-only, and parsing metadata blocks out of a
+FLAC. Whether a body is finite is a property of the stream, so it now travels
+on the command; whether a remote session is running stays what it always was,
+and still drives the metadata the bar shows.
+
 **The queue panel** switches to a dedicated `RemoteQueueView` while a remote
 session plays (keyed on `isRemoteTrack`), reading `remote_get_play_queue` (an
 in-memory snapshot) and jumping with `remote_queue_jump` — the local
--- a/docs/rfcs/RFC-005-remote-source-and-sync-v2.md
+++ b/docs/rfcs/RFC-005-remote-source-and-sync-v2.md
@@ -469,9 +469,13 @@ says. That preference exists to spend less bandwidth on a stream heard once;
 baking a lossy re-encode into a file someone chose to keep would make a
 bandwidth decision permanent, and silently. The file is hashed while it is
 written, in the same pass that copies it, which is what makes the plan's "free"
-reconciliation proof real: the server's own `full_hash` — not the library's
-`file_hash`, which covers a file the server has never seen — is in hand the
-moment a `track` row exists to attach it to, with no re-read.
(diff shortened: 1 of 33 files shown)


InstaZDLL:
@coderabbitai Five of six applied in a3ab56a8, one deferred with its consequence stated. All verified against the code first.

**The two that were real bugs.**

- **Concurrent downloads of the same track shared one working file.** The name derived from the track alone, so a second window or a double click had both callers writing the same path. This is precisely the defect the stream cache already fixed by making its working name unique *per call* — and I did not carry the lesson across two modules I wrote in the same session. Both parades now: a per-call name so nothing can interleave even if a guard is bypassed, and a single-flight claim held from the existence check through the rename and the row insert, so the second download does not happen at all. Re-checked under the claim, since the copy may land while waiting.
- **Nothing checked what came back.** `remote_track.full_hash` records what the server says the bytes hash to, and we hash while writing, so the comparison is free — and it is the only thing between a truncated or substituted body and a file later offered as an *exact reconciliation proof*. A track mirrored before that column was populated has no digest, and that accepts the download rather than refusing it: an unverifiable copy is still the copy the server just sent.

**Two more that were mine, and are the same omission twice.**

- **The resume snapshot flattened every stream to radio** — `seekable_file` was matched and discarded, so a finite server track resumed after a device change came back forward-only and ICY-parsed.
- **The duration was lost on the way to the fallback** — the decoder took it from the remote session alone, which knows nothing about a library track streaming from the server, so the seekbar lost its total exactly where the track's own row had it. What the caller declares now wins over the session.

Both are the same shape: I moved a fact onto the command and then let two paths keep reading the old global source.

**The RFC wording** conflated *the hash is available* with *the link can be made*. Corrected: the digest is known the moment the write completes, with no re-read and with no `track` row needed — it is stored on `remote_track_download` there and then. Only the *link* waits for a row.

**Deferred: extending the fallback to `player_next` and auto-advance.** You are right that it is a gap. It is not a minimal fix: four other sites dispatch a library track, and `LoadRemoteFileAndPlay` carries no `source_type` / `source_id`, so routing them through it would silently change how those plays are attributed in analytics. That is a design question about play attribution, and I would rather answer it deliberately than fold it into this branch. Today the fallback covers an explicit play; the other four fail exactly as they did before, so nothing regresses.

417 lib tests green, `cargo clippy --workspace --all-targets -- -D warnings` clean in both feature configurations, `cargo fmt` clean.

InstaZDLL:
@coderabbitai Skipped, after verifying it — and the fact you cite is correct, so it deserves a precise answer rather than a one-liner.

`0` is passed, and it is not a lost duration. It is the decoder's **live-stream sentinel**, and the comment sits at the call site: `play_track` reads `stream.duration_ms > 0` to gate crossfade prefetch (`decoder.rs:1434`) and the end-of-track guard. Replacing it would not fill in a missing number, it would switch those two behaviours on for every remote stream — and the remote queue advances on EOF through `AnalyticsMsg::RemoteTrackEnded`, not through that guard, so it would be a change to an RFC-005 seam rather than a value fix.

The implied consequence — that the seekbar loses its total — is not the mechanism either. **The bar reads the event, not the stream**: `PlayerContext.tsx:497` does `setDurationMs(e.payload.duration_ms ?? 0)` from `player:radio-metadata`. That payload is precisely what the previous commit corrected, so the total is already right for the fallback case; the two values answer different questions and only one of them is displayed.

Two more checks before concluding:

- The `0` is **not** something this branch introduced — it is on `main` at `decoder.rs:739`, with the same comment.
- I looked for a real consequence to report and did not find one: `completed` is indeed always false at duration 0, but `RemoteTrackEnded` is matched and ignored in the scrobble arm (`analytics.rs:209`), so remote plays never take that path. It would have been easy to flag that as a latent gap; it is not one.

So: no change. If crossfade across remote-queue tracks is wanted, that is a deliberate feature with its own advance-seam questions, not a parameter to widen here.

PR #574, opened 2026-09-06
feat(remote): correct a server track, and read the feed that reports it

InstaZDLL:
Lot 6 of the unified library, in two commits: a server track can now be **corrected**, and the desktop now **reads the server's change feed**. The two belong together — the second is what makes a correction made on another device show up here.

---

# 1. Correcting a server track

A track on the bound server could be browsed, played, downloaded and imported, but never corrected. The tag editor writes a file, and there is no file here — so a badly tagged server track stayed badly tagged.

The correction now goes to the server, which stores it *beside* the track rather than in the file: it survives that library's rescans, and every member of the library sees it. Nobody's collection gets rewritten.

## The endpoint inverts the queue's usual rule

`PATCH /api/v2/tracks/{id}` is **wholesale**: the body is the complete set of corrections the track should carry afterwards. A field left out is not a field left alone — it is a correction *withdrawn*, and the track falls back to what its file's tag says.

Every other update in the outbound queue is incremental, and needs a `clear_*` flag to empty a field because the server coalesces absent with null. This one is the opposite, so `UpdateTrackMetadata` is the single mutation with **no** `clear_*` flags — and callers may never send a diff. Sending only what changed, which is exactly what the local editor does, would silently drop every correction it did not mention.

**Only the six fields the editor shows are sent.** The server also stores `sort_title`, `comment` and `musicbrainz_recording_id`; under wholesale semantics, sending a field we cannot show the user is indistinguishable from clearing it. That is sound **only while this application is the sole writer of `track_override`** — which it is today: the server's own web client sends no such patch (its two `PATCH` calls are playlist and admin). The day a second writer appears, this quietly erases their corrections, and the fix is a server route returning the **raw** overrides. `GET /tracks/{id}` cannot stand in for it — it answers the merged values, in which a correction is indistinguishable from what the file said. **This is a fifth server capability to ask for**, written down in the code and in RFC-005 rather than left as folklore.

**The reply is the only place the underlying tag can be learnt.** Withdrawing a correction hands back a value this device has never seen — it only ever saw the correction masking it. So the drain writes the patch's reply into the mirror through `projection::apply_track_metadata`, which writes its columns **flat** rather than coalescing like `cache_song` does. Coalescing would keep showing the year the user just deleted, with no later pass able to put it right; `a_patch_reply_can_blank_what_the_upsert_would_have_kept` pins that down.

## Shape

It rides the existing outbound queue rather than calling the network directly, so a correction made offline waits and lands later, like every other gesture. `set_track_metadata_in_tx` keeps the module's rule: the mirror row and the queued mutation are written in **one transaction**.

A separate dialog rather than a mode of the local Properties one — Properties is a file inspector, and for a track on somebody else's disk almost none of it exists. Labels are reused from the local editor rather than duplicated: a second set would drift against the first across seventeen locales before anyone noticed.

The load is **stamped** with the track it filled the form from. A failed load leaves the form blank, and blank is not neutral under a wholesale patch — saving it would withdraw every correction the track carries — so saving waits for a form actually filled from this track.

---

# 2. Reading the server's change feed

`AlbumItem::song_count` is what makes the catalogue walk cheap, and it is also a blind spot with a precise shape. A track added or removed moves that count, so the walk notices. A **correction does not.** Somebody retitles a track — from another device, or from the dialog above — the album's count is unchanged, the album is skipped, and the mirror goes on showing the old title until something unrelated forces a re-fetch.

`GET /libraries/{id}/events` is what sees those. This is **not a faster sweep**: it reports a class of change the sweep was never able to.

It runs **before** the walk, not after. What it produces is a set of invalidations — an album whose `mirrored_at` it cleared, a library whose cursor the server refused — and the walk is what acts on them. Reading it afterwards would leave every one of them for the next walk.

## Three things that are load-bearing

**A track upsert carries `full_hash`, and that is the point.** Nothing else tells a client that a file was retagged *outside* the API: the track keeps its identifier while its bytes move, so an `exact_full_hash` link built on the old bytes stays `confirmed` while being wrong. Reading the feed is what marks it `stale` — the consumer RFC-006 explicitly said the desktop was missing.

**A refused cursor is not an error.** The server purges old events behind a watermark and refuses a cursor below it, rather than handing back the surviving tail — which would look like a successful catch-up while silently skipping the gap. The answer is to forget the cursor **and the sweep date**, so the walk rebuilds the mirror. The failure mode is "no faster than before", never "wrong".

**The refusal arrives as `conflict`, not `cursor_expired`.** The sync journal answers the second code for the same situation; this feed maps its refusal through `ServiceError::Conflict` and answers the first. A reader watching for `cursor_expired` here would never match, and would retry a permanent refusal forever. `RemoteFailure` keeps the code structured rather than folded into the message for exactly this reason — and `is_conflict` loses its `#[allow(dead_code)]`, since it now has a real caller.

## Notes

- Our own writes come back through the feed and are skipped on `origin_device_id`. The drain has already applied them from the patch's reply, which is the better source — the state the server settled on, not a notification that it changed. Skipping them also stops an upload arriving as a track this device has just discovered.
- The cursor advances only behind an event that landed, and is never lowered. A fetch failing mid-page leaves the rest for the next pass rather than skipping it: the cursor is the only record of what was read, so a gap in it is a gap nothing can detect afterwards.
- A feed failure never fails a walk — it is an optimisation over a sweep that still runs.
- The mirror card counts feed events apart from walked albums, and stops saying "unchanged" when the feed changed something: a correction applied through it changes what the library shows while leaving every album's count alone.
- `forget_track` moved into `mirror` and is shared, so the "nothing else references it" predicate stays written once — the macro it uses exists for that reason.

---

## Validation

`cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `bun run typecheck` and `bun run lint` are green locally.

**The tests could not be run locally**, and this is worth stating rather than glossing: the app crate does not start under Windows (`STATUS_ENTRYPOINT_NOT_FOUND`), and the usual WSL fallback failed — no DNS there, and the Windows cargo cache has never resolved for Linux, so the Linux-only crates are absent from it. The Rust (ubuntu) job is the only place they execute. It ran on the first commit: **452 passed, 0 failed**, six more than `main` — so those tests did execute rather than merely compile. The second commit's tests are checked the same way rather than assumed.

## Summary by CodeRabbit

  - Les métadonnées des pistes distantes peuvent être corrigées depuis la bibliothèque, sans modifier les fichiers locaux.
  - Les corrections sont enregistrées côté serveur, conservées après les rescans et visibles par tous les membres.
  - Vider un champ rétablit la valeur issue du fichier distant.
  - La synchronisation applique les mises à jour du catalogue et affiche le nombre de corrections traitées.

- **Améliorations**
  - Ajout de libellés et d’explications dans les langues prises en charge.


InstaZDLL:
@coderabbitai Thanks — five of the eight are in `04a2f335`, including the one that mattered most.

**Kept**

- **Stale reply overwriting the projection** (`write.rs:147`). Real, and worse than the comment supposed: `drain` had no lock at all, so `spawn` being edge-triggered meant two quick edits started two passes that each read the queue before the other removed anything. Harmless while a reply carried nothing — the operation identifier makes the server treat the second send as a replay — but this branch is the first where a reply writes back. Fixed by serialising the pass (`PASS`, a process-wide `tokio::Mutex`) rather than by versioning each mutation: strict ordering is what the FIFO queue already promises, and two concurrent passes are exactly what took it away. Waiting rather than `try_lock`, so a pass started for B still runs after A's.
- **`"; "` and not `';'`**. Correct — `scanner::upserts::split_artist_name` splits the local half on exactly that string, and `split(';')` made "AC;DC" two artists that never existed. Under a wholesale patch that would have become the truth. Fixed, with a test.
- **Reset before each load**. Real. Fixed by re-keying the modal on the identifier, which is the strategy `TrackPropertiesModal` already uses — closing sets the key to `"none"`, so reopening mounts a fresh dialog rather than one still holding the previous form.
- **Escape during save**. Fixed: the hook now gets `handleClose`.
- **Actions column width**. Fixed, `2.5rem` → `5.5rem` in both `gridCols` and the skeleton. Worth noting it already overflowed with the two buttons that predate this branch; the third made it visible.

**Declined, with what I checked**

- **Render through a portal.** None of the repository's eighteen modals do (`grep -l createPortal src/components/common/*Modal.tsx` returns nothing). The invariant is about overlays whose ancestor is transformed or backdrop-filtered capping the stacking context — a `fixed inset-0` dialog mounted at the app root is not that case, and this one follows the same shape as the seventeen others. Changing it here alone would make this modal the odd one out without fixing a symptom anyone has.
- **Nested `overflow-y-auto`.** Twelve of those eighteen modals carry it, `BatchTagEditModal` — the closest neighbour, a tag form — included. It is the house pattern for a panel bounded at `max-h-[90vh]`. The invariant it cites is about virtualised lists and `usePageScroll()`, whose failure mode is a virtualiser pointed at a nested scroller; a six-field dialog has neither.
- **Reveal the action on touch.** True, and true of every action in this table — the two server buttons that predate this branch, the local "+", all `opacity-0 group-hover:opacity-100`. Fixing only the button this PR adds would leave one visible button beside two invisible ones, which is the inverse inconsistency. It deserves one deliberate pass over the table's action affordances rather than a third of one smuggled in here.

InstaZDLL:
@coderabbitai Second round: six kept in `f2547abc`, one declined. The first one was the best catch of either round.

**Kept**

- **A permanent failure wedged the feed** (`events.rs:225-234`). Any error stopped the pass without advancing the cursor, so an event that can never succeed blocked that library's feed on *every* pass from then on — and the commonest shape of it is ordinary rather than exotic: an upsert whose track was deleted before this device read the feed answers 404 when re-fetched. Permanent failures are now warned and stepped over, counted as `skipped_permanent`; transient and local ones still stop the pass, since those may work next time and the cursor is the only record of what was read. This mirrors what the outbound drain already does with `mark_failed` vs `record_attempt`.
- **The feed could fail a walk** (`mirror.rs:338-341`). Correct, and it contradicted the module's own doc comment: `catch_up` was propagated with `?`, so a failure loading the cursors or writing one back took the whole catalogue walk with it. Per library it was already swallowed; this covers the parts around that loop.
- **A `null` mirror row filled the form with blanks and marked it ready.** Under a wholesale patch blank is not neutral — a Save there would have withdrawn every correction the track carries. Now reported like any other load failure, which leaves saving disabled, with its own localized message.
- **`parseInt` truncated in silence.** A `type="number"` input accepts "1.9" and "1e3"; `parseInt` read both as `1`. `Number` + `Number.isSafeInteger` reads the whole string or refuses it.
- **The zero-walk message.** With nothing walked and the feed having applied something, the card paired "0 albums, 0 tracks" with the feed's count, which reads as a contradiction rather than as two facts. It now shows the feed's line alone.
- **Double-click on a remote action** no longer bubbles to the row and starts playback behind the dialog — the same guard the rating column beside it already carries.

**Declined**

- **Give the row's nested buttons their own semantics.** Real, and it predates this branch: the row is a `div[role="button"]`, so *all* of its children are presentational to assistive technology — the heart, the "+", and the two server buttons that were already there. Fixing it means the overlay-button pattern from #551 (a real `<button>` layered over the row, declared before the nested controls in the DOM), which is one deliberate pass over the table's row semantics. Doing a seventh of it behind a third icon would leave the table in a half-converted state that is harder to reason about than the current uniform one. Worth its own change.

Also worth recording here: the previous round's fixture failed CI on `NOT NULL constraint failed: track.file_hash`, and I had reported it green before that job had spoken. Rebuilt on `remote::hashing`'s fixture — real migrator, `foreign_keys` on, a real `library` row behind the foreign key — because merely adding the missing column would have left a fixture that enforces none of the constraints it is meant to be testing against. `977dad93` is green: 460 passed on ubuntu.

InstaZDLL:
@coderabbitai Kept the first half, declined the second, in `483a8460`. The finding was right about something I had actively claimed the opposite of.

**Kept — invalidating the albums.** The recovery did not recover anything. Freshness is decided **per album** (`remote_album.mirrored_at` against `song_count`, read once by `known_albums`), so clearing the *library's* sweep date changed nothing about which albums the walk fetches — and the events a refused cursor made us miss are corrections, which leave `song_count` untouched. The walk skipped precisely the albums that needed re-reading. The comment above `forget_cursor` asserted the opposite; it was describing an intention, not the code. Second time in this PR, so it is worth saying plainly.

Albums with `library_id IS NULL` are invalidated too: the column is nullable, and an album we cannot attribute to a library is one we cannot prove is unaffected. Re-walking it costs a request; skipping it would leave wrong metadata in place with nothing left to correct it. Test asserts all three cases, including that another library's albums are untouched.

The library's own sweep date now survives, which it should have all along: it is what the interface shows as "last copied", and the library *has* been copied.

**Kept — the loop, which the finding is right to point at even though I have fixed it differently.** A feed merely forgotten is asked from zero on the next pass, refused again, and invalidates the mirror again: a full re-walk every pass, forever. That is now impossible — a `CURSOR_UNREACHABLE` sentinel takes the library out of the feed, and the selection predicate skips it. Emptying the mirror clears the mark, which is the way back, so `clear` resets the cursor alongside the sweep date.

**Declined — persisting a reference cursor after the re-walk.** There is no such cursor to persist. The watermark (`library.events_purged_through`) is not exposed on any route: `GET /libraries/{id}/events` answers 409 without it, and no other endpoint carries it. Any value we could invent is wrong in one of two directions — too low is refused again, and too high **silently skips every event below it**, which is the exact failure this path exists to prevent (the server refuses a stale cursor rather than handing back the surviving tail for that reason). Naming the feed unreachable is the honest version of the same intent, and it is recoverable by a gesture the user already has.

Worth recording for whoever picks this up: exposing the watermark, or a current-cursor field, is the server change that would make a real reference cursor possible. It is the sixth thing on that list.

InstaZDLL:
@coderabbitai One kept in `eaa44ba6`, one already shipped, two declined.

**Kept — validate the numbers before sending.** Real, and the failure mode is the nastiest kind: `toNumber` folded anything unparseable into `null`, and under a wholesale patch `null` means "no correction". So typing "1.5" into the year — which a `type="number"` input holds quite happily — *withdrew* the correction the user believed they were making, silently. `parseWhole` now answers one of three things rather than two: blank is no correction, a whole number is that number, anything else is refused. Save is disabled while a field is unparseable and guarded again on submit.

On the regression test: this repository ships no JS test suite, so the behaviour is proven with a throwaway node script rather than by committing one — `""` and `"   "` blank, `"1995"` and `"-3"` pass, `"1.5"`, `"abc"`, `"1e400"` and `"9007199254740993"` refused, `"1e3"` surviving as 1000 since it *is* a whole number.

Adjacent and deliberately not done here: the server also bounds `year` to 1–9999 and refuses negative track/disc numbers, so "-3" is enqueued and then permanently refused with nothing shown to the user. Same family, wider blast radius (it wants the error surfaced from the queue, not just an input guard), so it deserves its own change rather than being folded in behind this one.

**Already shipped — invalidate album freshness on cursor refusal.** That is `483a8460`, pushed before this round: `forget_cursor` now clears `remote_album.mirrored_at` for the library (including `library_id IS NULL`, since an album we cannot attribute is one we cannot prove is unaffected), and leaves the library's own sweep date alone. You were right that the original did nothing — freshness is per album against `song_count`, and the missed events are corrections, which leave `song_count` untouched.

The other half — *persist the current server-provided watermark* — is declined, and the reason is factual rather than a preference: **the server provides no watermark**. `library.events_purged_through` is read internally by `library_changes` and exposed on no route; the 409 body carries only `{code, message}`. There is nothing to persist. Any value invented instead is wrong in one of two directions: too low is refused again, too high **silently skips every event beneath it**, which is precisely the failure the server's refusal exists to prevent. The library is marked unreachable instead, and emptying the mirror clears the mark.

**Declined — gate the edit button on `remoteAvailable`.** The two server actions that predate this branch (keep offline, import) are gated the same way it is: on the row being remote, not on the capability. A row with `source === "remote"` only exists when the mirror holds rows, which only happens under `sync_v2`. Adding the check to the third button alone would leave two buttons that can still be shown beside one that cannot — the inverse inconsistency. If the guard is wanted it belongs on all three, which is a change to existing behaviour and not this PR's.

**Declined — an offline check in `RemoteClient::refresh`.** `refresh` is untouched by this branch, and every entry point into it is already behind the invariant: `mirror_catalogue` returns early on `offline::is_offline()` (so does the feed, which it calls), and `drain` does the same. Adding a guard inside `refresh` is defensible as defence in depth, but it is a change to shared pre-existing code with no path in this PR that reaches it unguarded.

InstaZDLL:
@coderabbitai Already fixed — this is `eaa44ba6`, pushed before this round landed. No change made, and here is the check rather than the claim:

- `toNumber` no longer exists. It is `parseWhole`, returning `{ ok: true, value: number | null } | { ok: false }` — three answers instead of two, which is the whole point.
- **`null` is produced only for a blank field**: it is returned from the `trimmed === ""` branch and from nowhere else.
- A non-empty non-integer answers `{ ok: false }`, and `handleSave` returns at line 167 — **before** `remoteUpdateTrackTags` — showing `remote.tags.invalidNumber` instead. Save is also disabled up front via `numbersValid`, so the button says "not yet" rather than the dialog saying "no" afterwards.
- Valid whole numbers pass through unchanged.

The reasoning matches yours exactly, and it is written into the doc comment so the next reader does not have to rediscover it: folding an unparseable value into `null` sends "no correction" under a wholesale patch, silently withdrawing the correction the user believes they have just typed.

Verified with a throwaway node script rather than a committed suite, since this repository ships no JS tests: `""` / `"   "` blank, `"1995"` / `"-3"` pass, `"1.5"` / `"abc"` / `"1e400"` / `"9007199254740993"` refused, `"1e3"` surviving as 1000 since it is a whole number.

PR #575, opened 2026-09-06
feat(remote): paint the server's Canvas behind a remote track

InstaZDLL:
Lot 7 of the unified library. A Canvas is the short looping clip the now-playing view paints behind the cover; a track playing from the bound server never had one.

**Not for want of a server route.** Those have been there since PR #154–165. What was missing is that nothing downstream could *name* the track to the server: a remote track plays under a **negative sentinel minted per playback**, meaningful only inside this process. `useTrackCanvas` held an id it could not ask about — `get_track_canvas` would have been querying the local database for a rowid that does not exist.

## Making the identifier travel

By the route `artwork_hash` already took, which is why this half is small: `RemoteStreamMeta` carries the server id out of the queue entry **under the same lock** as the duration and the artwork hash (separate reads could straddle a `clear()` and emit a track's id with another's duration), `player:radio-metadata` carries it to the frontend, and the projected `Track` carries `remote_id` — absent meaning local, as everywhere else in the unified library.

## The clip rides a ticket

A `<video src>` cannot send a Bearer header. That is the same problem remote *audio* has, and the server answers it the same way, so `remote::canvas` is `remote::stream`'s shape applied to a second kind of media — `POST /tracks/{id}/canvas-ticket`, then a self-authorising `/api/v2/canvas-stream/{ticket}`.

It keeps the rule that matters: **the URL is built from our base and a relative reply is required**. An absolute URL from the server would point the webview at a host the user never authenticated against — and unlike a fetch, a `<video>` element loads one without a murmur. `is_relative` is its own function so the test exercises the predicate production uses rather than a copy of it, which is how a restated rule drifts while staying green.

## Two things that fall out rather than being arranged

**The ticket expires within the hour and must not be cached durably** — and it cannot be. `useTrackCanvas` keys on the local track id, which is per-playback here, so replaying a track mints a fresh ticket instead of reusing a stale one.

**A track with no Canvas is the ordinary case, not a failure.** The server answers 404, which arrives classified as permanent and resolves to `None`, falling through to the plugin rung. Only a transient failure is reported as an error, so a server that is merely unreachable does not read as "no Canvas" to a caller that might remember it.

## Precedence, and what needed no guard

A server clip takes the rung the manual local one holds: somebody put it on that library deliberately. Below it, unchanged — **server/manual > plugin > motion cover > slideshow > static cover**.

The picker needed nothing: it already required `currentTrack.id >= 0`, a guard put there for radio, so a server track was never offered a "replace this clip" it could not honour. `isRemoteCanvasUrl` likewise already sorted URL from path — it now has two remote origins instead of one, and the three comments that named only the plugin were corrected rather than left to rot.

## Validation

`cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `bun run typecheck` and `bun run lint` green locally. The app crate's tests do not run under Windows, so the Rust (ubuntu) job is checked on this PR rather than assumed.

**Not exercised against a real server** — as with every lot before it, nothing here has yet talked to one. The paths that can be tested without a network are; the ticket round-trip is not one of them.

## Summary by CodeRabbit

  - Prise en charge des Canvas pour les pistes diffusées depuis un serveur distant.
  - Génération d’URL de lecture authentifiées via des tickets temporaires, renouvelés à chaque lecture.
  - Repli automatique vers les Canvas des plugins lorsqu’aucun Canvas distant n’est disponible.
  - Transmission des identifiants des pistes distantes jusqu’à l’interface de lecture.
  - Les flux radio en direct ne sont pas concernés par cette identification.

- **Documentation**
  - Documentation mise à jour pour décrire la résolution des Canvas distants et leur priorité de lecture.


InstaZDLL:
@coderabbitai The two guard findings were the important ones, and both were right: I had shipped a **dead path**. Fixed in `729d3c37`.

**The lookup guard** (`useTrackCanvas`, line 233). It skipped every negative track id — a rule written for radio and Spotify, which have no library row and nothing to resolve a plugin Canvas against. A server track is negative too, but it is not in that position: `remote_id` names it on the server, which is exactly what the remote lookup asks about. The guard is now on having something to ask *with* rather than on the sign of the id.

**`remote_id` was never passed to `lookup`** (lines 234-240), so even past the guard the branch could not have been taken. Passed now, with `remoteId` in the effect's dependency array.

**And a third one you did not have in view, found by auditing the rest of the path** — which is the lesson I should have applied before opening this: `ImmersiveView` used a single value, `canvasTrack`, for two different questions, and the stricter one won. *Displaying* a Canvas needs no library row; the **picker** does, since it writes `track_canvas` keyed on a rowid. Conflated, the "Show Canvas" toggle went unrendered for a server track — and that toggle **defaults to OFF**, so the clip could never have been revealed at all, even with the two fixes above. Split into `canvasShowable` / `canvasEditable`.

The remainder of the path checks out as it stands: `NowPlayingPanel` gates on `canvasAvailable` alone, both `CanvasStage` mounts take `canvasPath` unguarded, and `isRadioTrack` is false for a server track since it requires `codec === "Web Radio"`.

**Declined — remove the generic catch in `lookup`.** The fail-soft is deliberate and load-bearing: a Canvas is decoration, and the `canvas` fanout is documented as never breaking playback. Propagating would also change nothing observable, since the sole caller already handles rejection with `() => apply(null)`. The fair half of the criticism was that it was swallowed *without trace* — "no Canvas" and "the lookup broke" looked identical on screen — so it now logs before returning null.

**Declined — require HTTPS for the ticket base.** `auth.rs:121` accepts `http` **or** `https` deliberately: a self-hosted server on a LAN is a supported case. The Canvas ticket is no more sensitive than what already crosses that same base — the audio stream, the artwork, and the OAuth tokens themselves. Restricting the scheme is a policy change for the whole binding, not for one media route, and it would break existing installations. Worth its own discussion; not something to smuggle in behind a Canvas.

Issue #485, opened 2026-08-05
refactor: profile-scope set_profile_setting + factor out the duplicated preference-hook pattern


InstaZDLL:
Two related weaknesses surfaced during the review of #483 (artist hero, #482). Both are pre-existing and shared across the codebase, so they were deliberately left out of that PR rather than fixed for one hook only — fixing a single call site would have given the illusion of coverage while 20+ others stayed exposed.

## 1. `set_profile_setting` is not scoped to a profile (TOCTOU)

[`commands/profile.rs::set_profile_setting`](src-tauri/crates/app/src/commands/profile.rs) takes `key` / `value` / `value_type` and resolves the pool itself:

```rust
pub async fn set_profile_setting(state, key, value, value_type) -> AppResult<()> {
    let pool = state.require_profile_pool().await?;
    // INSERT ... ON CONFLICT DO UPDATE
}
```

Every JS caller guards the write against a profile switch by comparing a captured id with the live one — the pattern in [`useCoverSlideshow`](src/hooks/useCoverSlideshow.ts) and its copies:

```ts
const write = writeChainRef.current.then(async () => {
  if (activeProfileIdRef.current !== profileId) return;
  await setProfileSetting(KEY, next ? "true" : "false", "bool");
  ...
});
```

That check happens **before** the `await`. A `switch_profile` landing between the guard and `require_profile_pool()` acquiring the lease writes the preference into the **wrong profile**. The window is small and the blast radius is one UI preference (no data loss), but it's a genuine correctness gap and it exists on every write path.

**Fix:** take an expected profile id on the command and validate it **while holding the same lock used to acquire the active pool** (see [`state.rs`](src-tauri/crates/app/src/state.rs)) — a JS-side guard fundamentally cannot close this, only the backend can. Mismatch should be a no-op or an explicit error the caller can ignore, not a write.

**Scope:** 22 call sites across 18 files (`setProfileSetting` in `src/`), including every preference hook, `ThemeContext`, `SkinContext` and `lib/shortcuts.ts`. `get_profile_setting` deserves the same treatment for symmetry — a read racing a switch returns another profile's value.

## 2. The preference-hook pattern is copy-pasted 7×

These all reimplement the same machinery — serialized write chain, profile-switch guards, rollback to the last backend-confirmed value, a `window` event to re-read across mounted consumers:

- [`useScrollLongTitles`](src/hooks/useScrollLongTitles.ts)
[...]

PR #625, opened 2026-09-12
fix(player): write the resume point on exit and every ten seconds

### LINKED ISSUE #624: bug: the last-played position is never restored, the app always reopens at 0:00
Found while smoke-testing #621. Not a regression from it — the path has never worked.

## What happens

WaveFlow is supposed to reopen on the last-played track, at the position it stopped at. It always reopens at **0:00** instead, on whatever track the queue cursor points at.

## Why

`queue::persist_resume_point` has exactly one call site: `WindowEvent::Destroyed` on the `main` window, gated by `QuitGate`. Neither way of leaving the app reaches it.

- **Tray → Quit** calls `request_quit` → `app.exit(0)`. Tauri 2.11.5 documents `App::exit` as triggering `RunEvent::ExitRequested` and `RunEvent::Exit` — *app* events, not `WindowEvent::Destroyed`.
- **The window's X** arms the gate only when `minimize_to_tray` is off. With it on, the close is intercepted and the window is merely hidden, so nothing is armed and nothing is written.

`persist_resume_point` is therefore dead code in practice, and `queue::restore_state` always falls through to its second branch — the queue's current track at position 0.

## Measured

On 2026-09-12, with `app.minimize_to_tray = true`:

- two consecutive sessions were ended through the tray's **Quit**, each with a track playing (`track_id = 174`);
- `player.last_track_id` and `player.last_position_ms` stayed at `0`, last written on 6 September;
- the same sessions did write `queue.current_index = 7` and `player.volume = 20`, so the profile database was reachable and writable throughout.

## Consequence beyond the resume point

The media session published to the OS overlay at launch (#609, shipped in #621) advertises the restored track at 0:00 for the same reason — the position it would show is never saved.

## Direction

Hook the persist on the app-level exit events (`RunEvent::ExitRequested` / `RunEvent::Exit`) so every way out is covered, and persist periodically during playback as well: a crash, a forced kill or a power loss all lose the position today, and a periodic write is the only thing that covers them.

InstaZDLL:
Closes #624.

WaveFlow never came back where it stopped. The resume point had a single writer — `WindowEvent::Destroyed` on the main window, behind the `QuitGate` — and no way out of the app reaches it:

- the tray's **Quit** calls `AppHandle::exit`, which triggers `RunEvent::ExitRequested` and `RunEvent::Exit`, not a window event;
- the window's **X** only hides the window while close-to-tray is on, so nothing is destroyed and the gate is never armed.

`restore_state` therefore always fell through to its second choice, the queue's current track at 0 ms. Measured on my own profile database: `player.last_track_id` and `player.last_position_ms` still sat at their initial `0` while `queue.current_index` and `player.volume` had been written the same day.

## What changes

- The write now hangs off **`RunEvent::Exit`**, which every route reaches. That needs `.build(context)?.run(|app, event| …)` instead of `.run(context)`, since the Builder's own `run` takes no callback.
- A **ten-second ticker** persists it during playback as well — a crash, a kill or a power loss leaves no exit event to hook at all. It skips the write when neither the track nor the second it sits on has moved, so a paused or idle session doesn't wake the single SQLite writer for nothing.
- Both go through one helper, and writing the same pair twice is harmless, so no caller has to coordinate with the others. The window path keeps silencing the output and shutting the decoder down, as before.
- `current_resume_point` returns `None` for anything that isn't a real library track: radio and the remote queue use negative ids, nothing loaded is `0`, and `restore_state` has no `track` row to find for either.

## Test

`cargo fmt --check` and `clippy -D warnings --all-targets` clean on Windows.

Manual, on the dev build: play a track, quit through the tray, relaunch — the track and its position come back, and Play picks it up from there instead of restarting it.

## Summary by CodeRabbit

  - Le point de reprise de la lecture est restauré au démarrage ou à la demande.
  - La position de lecture est sauvegardée automatiquement toutes les dix secondes et lors de la fermeture de l’application.
  - La reprise est conservée après une reconstruction liée à une perte de périphérique audio.
  - Les sessions sans lecture active, les radios et les files distantes ne génèrent pas de point de reprise.

- **Documentation**
  - Ajout d’une documentation sur la persistance et la restauration du point de reprise.


InstaZDLL:
@coderabbitai All three are real, and all three are fixed in the commit above.

**Profile pin.** Agreed, and there was already a precedent to match: the device-error rebuild pins its own resume-point write the same way (#485). `write_resume_point` now captures the active profile id *before* the await with a non-blocking `try_read` — a lock it can't take means a switch is already under way — and resolves the pool through `require_profile_pool_for(Some(profile_id))`. No id, no write.

**Retry after a failed write.** Correct, and the consequence was worse than a single missed tick: the marker moved anyway, so the same point was then skipped as *unchanged* and never retried. A track paused right after a failure would have kept a stale position until it moved again. `write_resume_point` returns whether the point reached the database and the ticker only advances `last` when it did. The exit path is unaffected — it writes unconditionally.

**Atomicity.** Agreed. Three callers can be writing (ticker, exit event, device-error rebuild), so an interleaving could store one track's id beside another's position — which resumes a real track at the wrong place rather than failing visibly. Both rows now go in a single transaction; `write_setting_i64` is generic over the executor, so the two `queue.current_index` callers keep passing the pool and are unchanged.

`cargo fmt --check` and `clippy -D warnings --all-targets` are clean on Windows. The behaviour is documented in `docs/features/playback.md` § Queue → Resume point.

Issue #627, opened 2026-09-12
Follow the OS default output device when none is pinned


InstaZDLL:
Split out of #612, which covers the two halves that could be fixed without new platform plumbing: the picker now flags the device that actually opened rather than the one that was requested, and re-selecting the active row forces a fresh open. This is the third half, and it is a different size.

## What happens

With no device pinned, `device_name = None` opens whatever the OS default is **at that moment**, and the stream stays bound to that endpoint for its lifetime. Nothing subscribes to default-device changes: neither WaveFlow nor cpal 0.17.1 registers an `IMMNotificationClient`, and there is no polling. The only thing that ever reopens the stream is the device becoming invalid (`DeviceNotAvailable`).

So when Windows moves the default — earphones reconnecting, a headset waking, a monitor's HDMI sink appearing — playback keeps going on the previous endpoint. After #612 the menu at least tells the truth about which one that is, but "follow the OS default" still does not follow anything.

## What it would take

- **Windows** — `IMMDeviceEnumerator::RegisterEndpointNotificationCallback`, then rebuild on `OnDefaultDeviceChanged` for the render/console role. cpal exposes none of this, so it means the `windows` crate and raw COM, on the exclusive side as well as the shared one.
- **macOS** — a property listener on `kAudioHardwarePropertyDefaultOutputDevice`.
- **Linux** — the sound server already migrates streams on a default change in most setups, so this may be a no-op; worth measuring before writing anything.

Only rebuild when the pin is `None`. A user who pinned a device asked for that device, and moving them because the system moved would undo the pin.

## Care needed

The rebuild machinery already has two gates that exist because this kind of event storms: `RebuildGate` (one rebuild per burst) and `FlapWindow` (a device that resets on every exclusive grab gives up on exclusive for the session). A default-change notification arrives on an OS thread, can fire several times for one physical event, and our own reopen can trigger another one — so it must go through the same debounce rather than calling the rebuild directly.

Worth pairing with #597: when a rebuild moves playback to a different endpoint, the user should be able to see that it happened.

Issue #612, opened 2026-09-11
bug: the device menu can show one output while audio plays on another


InstaZDLL:
Reported by email by a Windows 11 user:

> Sometimes Windows, or an audio device such as my earphones, changes which output device should be used. WaveFlow's UI may show the expected output device, but playback does not work correctly through that device. To fix it, I have to select a different output device in WaveFlow, then select the original output device again.

This is the same family as #597 (silent fallbacks) and #604 (its exclusive-mode variant). #606 does not address it.

### 1. "Default device" doesn't follow the default

`device_name = None` opens whatever the default endpoint is at that moment, and the stream **stays bound to that endpoint**. Nothing listens for changes: neither WaveFlow nor cpal 0.17.1 registers an `IMMNotificationClient`, and there is no polling.

The only reopen trigger is the stream being invalidated (`DeviceNotAvailable`). When Windows switches the default back to the earphones, for instance after they reconnect, the stream keeps playing on the old endpoint. Meanwhile the device menu marks the *new* default as active, because `None` is displayed as "whatever `is_default` is now" (`player_list_output_devices`).

### 2. A pinned device that isn't found is replaced silently, but still displayed

When the requested name isn't enumerated, both open paths fall back to the default device with only a log line: shared mode in `output.rs`, exclusive mode in `pick_device` (`wasapi_exclusive.rs`).

The resulting `OutputHandle` still records the **requested** name (`device_name` is passed through as-is). `is_active` compares against that name, so the menu ticks the pinned device while audio plays elsewhere. When the pinned device comes back, nothing reopens the stream on it.

### 3. Why "another device, then back" is the only fix

- The active row in `DeviceMenu.tsx` is disabled, so it can't be clicked again.
- `set_output_device` returns early when the name is unchanged.
- Picking a different device, then the original one, forces a real lookup against an endpoint that is present again.

## Expected

- The menu shows the device audio actually goes to.
- In default mode, playback follows Windows' default device.
- A pinned device that was replaced by a fallback is reopened when it reappears, or the fallback is at least visible (#597).
- Re-selecting the active device forces a reopen.

## Directions

- Record the device actually opened in `OutputHandle`, separately from the one requested.
- Watch default-device and device-state changes on Windows (`IMMDeviceEnumerator::RegisterEndpointNotificationCallback`) and rebuild when the relevant endpoint changes.
- Let a click on the active row force a rebuild.

Issue #604, opened 2026-09-10
bug: Audio output device issue


zipedguy:
### WaveFlow version

1.7.0

### Operating system

Windows 11

### How did you install WaveFlow?

GitHub Releases (NSIS .exe)

### What happened?

when starting waveflow for first time it switches to my laptop speaker without any audio output then when i select my wireless headphones it still doesn't play then when i quit the app from task manager and relaunching it again it plays fine 

### Steps to reproduce

quit waveflow
disconnect bluetooth audio device
open waveflow
select a flac file to play
no audio
manually select the audio device still no audio

### Relevant logs / errors

```shell
2026-09-10T04:58:49.065915Z  WARN waveflow_lib::audio::wasapi_exclusive: wasapi exclusive output thread lost the device; requesting rebuild reason=wasapi wait_for_event failed: EventTimeout
2026-09-10T04:58:49.377094Z  INFO waveflow_lib::audio::engine: rebuilding cpal output after DeviceNotAvailable device="Headphones (wireless)" exclusive=true
2026-09-10T04:58:49.378865Z  INFO waveflow_lib::audio::decoder: play_track interrupted finished_track_id=282 listened_ms=168710 will_credit=true
2026-09-10T04:58:49.378941Z  INFO waveflow_lib::audio::analytics: insert play_event track_id=282 listened_ms=168710 completed=false source_type="device-rebuild"
2026-09-10T04:58:49.382508Z  WARN waveflow_lib::audio::wasapi_exclusive: wasapi exclusive: requested device not found, falling back to default requested="Headphones (wireless)"
2026-09-10T04:58:50.107192Z  INFO waveflow_lib::audio::wasapi_exclusive: wasapi exclusive negotiation succeeded sample_rate=48000 channels=2 format="S24_4LE" layout="endpoint"
2026-09-10T04:58:50.107236Z  INFO waveflow_lib::audio::wasapi_exclusive: wasapi exclusive stream opened sample_rate=48000 channels=2 buffer_frames=480
2026-09-10T04:58:50.107246Z  INFO waveflow_lib::audio::output: audio output: WASAPI Exclusive Mode engaged
2026-09-10T04:58:50.107369Z  INFO waveflow_lib::audio::decoder: decoder picked up new ring producer
2026-09-10T04:58:50.173010Z  INFO waveflow_lib::audio::decoder: decoding start dst_sample_rate=48000 dst_channels=2 track_id=282 is_live=false
[...]

PR #628, opened 2026-09-12
fix(audio): report the device that opened, not the one that was asked for

InstaZDLL:
Addresses two of the three halves of #612. The third is split out as #627 — see below.

## The defect

The device menu could tick one output while the audio came out of another, and the only cure was to pick a different device and then pick the original one back.

Three independent causes, all confirmed in the current code rather than reproduced from the report:

1. **Every open path falls back to the default endpoint silently.** `build_stream_inner` in shared mode, `pick_device` under WASAPI, `resolve_device` under CoreAudio — each leaves nothing behind but a `warn!`.
2. **`OutputHandle` keeps the requested name**, on purpose, so the user's pin survives until the device returns. The picker read that field, so after a fallback it flagged a device that was playing nothing.
3. **Re-selecting the active device is a no-op**: `set_output_device` returns early when the pick equals the current pin, and the menu disabled the active row on top of that.

## What changes

**The handle carries both names.** `device_name` is still the pin; `opened_device` is what the backend actually opened. Each backend fills it with what it can honestly claim, and they are not symmetric:

| Backend | `opened_device` | Why |
| --- | --- | --- |
| cpal shared | name read back off the chosen device | a fallback names the endpoint it landed on |
| WASAPI | `Device::get_friendlyname()` | same, through the endpoint we were handed |
| ALSA | the pin | `resolve_hw_device` places it on a real card or **fails** — it never quietly takes another. An absent pin lands on card 0, which has no name in the picker's terms → unknown |
| CoreAudio | the pin when it matched, else unknown | after the fallback only a device id is left, and inventing a name would be the very bug |

**The picker tells both apart.** `is_active` comes from what opened (falling back to the pin, then to the OS default); the new `is_pinned` comes from the pin. They differ exactly during a fallback, so the menu marks the pinned row unavailable instead of claiming it plays.

**The active row is clickable.** It calls `player_reopen_output_device` → `force_rebuild_output`, the path device-error recovery already uses. No preference changes, nothing is persisted.

## Deliberately not here

Following the OS default as it moves — cause 1 in the issue's own list. It needs `RegisterEndpointNotificationCallback` and raw COM on Windows (cpal 0.17.1 exposes nothing), a property listener on macOS, and it has to route through the existing `RebuildGate` because a single physical event fires several notifications and our own reopen triggers another. That is feature-sized and Windows-shaped, so it is **#627** rather than a half-measure buried in this PR. #612 stays open until it lands.

## Verification

`cargo fmt --check`, `clippy -D warnings --all-targets`, `typecheck` and `lint` all clean.

Manual check worth doing on Windows: pin a USB or Bluetooth output, disconnect it, start playback again. The menu should tick the device that is really playing and mark the pinned row unavailable — where it used to tick the absent one. Then click the active row: playback should reopen on it.

i18n: two new `deviceMenu` keys propagated to all 17 locales.

## Summary by CodeRabbit

  - Le menu audio distingue le périphérique sélectionné du périphérique réellement utilisé.
  - Les périphériques épinglés mais indisponibles sont signalés.
  - Cliquez sur le périphérique actif pour le rouvrir et reconstruire le flux audio sans modifier votre préférence.
  - Le périphérique effectivement ouvert est indiqué en cas de basculement vers une autre sortie.

- **Documentation**
  - Le fonctionnement selon les plateformes audio prises en charge est précisé.

- **Traductions**
  - Les nouveaux libellés sont disponibles dans les langues prises en charge.


InstaZDLL:
@coderabbitai Two of the three are right, and the first one would have left the feature dead in the case it exists for. Fixed in the commit above. The ALSA one I'm declining, with reasoning.

**Pinned device absent from the listing — agreed, and worse than stated.** The map only ever walks enumerated devices, so when the pin is gone no row carries `is_pinned` at all — in exactly the fallback this PR is meant to report. `pinnedUnavailable` could never render. The list now appends the missing device as its own row (`is_pinned: true`, `is_active: false`, guarded against duplicating an enumerated one). It is never marked active: it isn't there to be playing.

**`reopen_output_device` bypassing the deliberate-rebuild protections — agreed on both halves.** I checked `force_rebuild_output`: it owns neither guard, its recovery caller computes them (`exclusive = pref && !exclusive_suppressed`, and the gate). So the reopen now opens `begin_deliberate_output_change()` first, cancels it when the rebuild returns `Err` so a genuine error still reaches recovery, and computes `exclusive` with the `exclusive_suppressed` term. I deliberately do **not** call `reset_exclusive_suppression()` here: that is for a re-toggle or a device pick, which are fresh chances for exclusive to work. Asking for the same device again is not.

**ALSA publishing the pin rather than the resolved `hw:` string — declining.** `opened_device` is compared against `d.id` in `player_list_output_devices`, so it has to be expressed in the *picker's* identity space. `resolve_hw_device` returns an ALSA device string (`hw:0,0`), which is generally not a row in the hint listing the picker is built from — publishing it would match no row and leave the menu with nothing ticked, which is the failure this PR removes rather than a fix for it.

The premise also doesn't hold the way it's phrased: one card is reachable as `default`, `plughw:0,0` and `hw:0,0`, so a device string is not an identity — the same point that settled the endpoint-comparison question in #604. And ALSA is the one backend that never falls back silently: `resolve_hw_device` either places the pinned name on a real card or returns an error, so a pin that reached the handle *is* what is playing, under the name the user picked it by. The genuinely unnameable case — no pin at all, landing on card 0 — already reports `None`.

InstaZDLL:
@coderabbitai Agreed — fixed in the commit above.

The two accessors each took the output lock on their own, and every path that replaces the stream swaps the whole `OutputHandle` under that same lock: the reopen this PR adds, a device switch, and the post-device-error recovery. So the pair really could come from two different streams — the old one's endpoint beside the new one's pin — and the menu would tick one row while marking another as pinned, each true of a stream that is no longer playing.

`current_output_devices()` now returns `(opened, pinned)` under a single lock, and `player_list_output_devices` reads the pair from it. That removed the reason `current_output_opened_device()` existed, so it is gone rather than left unused.

Same class as the profile-snapshot race this codebase already had, which is what made the argument easy to accept: an invariant that holds between two values cannot survive being read through two separate locks.

InstaZDLL:
@coderabbitai The race is real and it is fixed in the commit above — though not by passing the guard down, which turned out not to be necessary.

`force_rebuild_output` consumes the device name in exactly one place, when it spawns the new stream, and it already holds the lock from its first line. So the fix is to stop handing it a name read from outside: it now takes a `RebuildDevice`, where `Pinned` resolves from the handle it is already holding — under the same acquisition it swaps that handle with — and `Explicit` preserves the device-error recovery's deliberately computed target. `reopen_output_device` reads nothing beforehand, so there is no window left to lose.

That keeps the recovery path's logic untouched, which mattered here: it carries the #322 / #405 / #604 history, and restructuring it to thread a guard through would have been a much larger change than the defect warranted.

One thing your finding led me to that is worth stating: **the same window exists in `set_exclusive_output`**, which captures `current_output_device()` before taking the lock it rebuilds under. That one predates this PR and does its teardown inline rather than through `force_rebuild_output`, so it is **#629** rather than a widening of a green branch mid-review. `set_output_device` holds the lock throughout and is unaffected; the recovery path computes its own target and is unaffected. Those three are the whole surface.

InstaZDLL:
@coderabbitai The observation is correct, but I'm filing it rather than patching it here — as **#630**.

You're right that `self.output` can be empty: `publish_output_lost_if_gone` exists for exactly that state, and both `force_rebuild_output` and `set_output_device` reach it when a spawn fails after the old handle was released. In that state `RebuildDevice::Pinned` resolves to `None`, because the engine keeps the user's choice **only** inside `OutputHandle` — `profile_setting['audio.output_device']` is read once at startup and never again.

Two things put it outside this PR, though.

**It isn't reachable through the UI.** With no handle, `player_list_output_devices` gets `playing = None` and `pinned = None`, so `is_active` falls back to the OS-default row and no pinned row is produced at all. The frontend only calls `player_reopen_output_device` for the active row, so a user cannot ask for "reopen my pin" in that state; clicking their own device goes through `player_set_output_device`, which works. The outcome you describe needs the command to be invoked directly.

**It isn't new.** Reading the pin off the handle is how `current_output_device()` has always worked — my first commit on this branch did the same thing through that accessor. The gap is that the pin doesn't outlive the handle, which also costs the picker its `is_pinned` flag after a failed open (the very marker this PR added) and makes MPD report no device for the same reason.

So the fix belongs where it can be done once and properly — the engine holding the pin as its own state, fed at construction and by a successful pick, used as the fallback wherever the handle is absent. That keeps the locking discipline from the previous round intact: the handle is still read under the rebuild's own lock, and the fallback only applies when there is no stream to race with. #630 has the detail, and notes it is worth doing alongside #629 since both are about state around the handle being read as if the handle were always there.

PR #631, opened 2026-09-12
fix(player): order playback loads by the intent that started them

### LINKED ISSUE #622: bug: a slow track load can overwrite a newer playback intent
Raised while reviewing #621, and deliberately left out of it.

## What happens

A track load is prepared asynchronously — profile snapshot, queue reads, ReplayGain lookup — and only then handed to the decoder. Two loads started close together therefore reach the decoder in the order their *preparation* finished, not the order the user asked for them.

```text
resume()  starts          intent 41
  └─ slow database work…
next()    starts          intent 42
  └─ sends LOAD B
resume()  finishes
  └─ sends LOAD A         ← overwrites the newer selection
```

Reachable by pressing Play (from a stopped player) and then Next within a few tens of milliseconds, or by a client sending two commands back to back — `mpc play; mpc next`.

## Why a local fix is not enough

There are **17 places** in `crates/app` that send a load command to the decoder:

| Module | Sends |
| --- | --- |
| `player_actions` (`resume_last`, `load_and_play`) | 2 |
| `commands::player` (play tracks, jump, resume, radio, remote) | 6 |
| `audio::engine` (output rebuild, device switch, exclusive toggle) | 6 |
| `remote::playback` | 4 |
| `audio::analytics` (auto-advance) | 1 |

Serialising only the two that a review happens to point at would leave the other fifteen reordering exactly the same way, while reading as though ordering were guaranteed. That is worse than the current, honest absence of a guarantee.

## Direction

Give every load a **monotonic generation, taken when the intent starts** — not when the command is finally sent — and let the single point they all pass through, `AudioEngine::send`, accept only the newest one. An older asynchronous preparation that finishes late is then dropped instead of overwriting a newer playback intent.

`send` is already the boundary that maintains the `radio_resume` snapshot, so the arbitration has one home rather than a scattering of locks.

## Out of scope here

`resume_last` already holds a one-at-a-time guard and publishes `Loading` before its send (#621), which stops it racing *itself*. This issue is about ordering *between* the different load producers.

InstaZDLL:
Closes #622.

## The bug

Handing a track to the decoder is never immediate. A profile snapshot, the queue read, a ReplayGain lookup — and for a remote queue a reconciliation lookup and a streaming ticket — all happen first. Each of the seventeen load sites then sent its command whenever *its own* preparation finished, so two loads started close together reached the decoder in the order they got ready, not the order they were asked for:

```text
resume()  starts          intent 41
  └─ slow database work…
next()    starts          intent 42
  └─ sends LOAD B
resume()  finishes
  └─ sends LOAD A         ← overwrote the newer selection
```

Reachable by pressing Play from a stopped player and then Next within a few tens of milliseconds, or by a client sending `play` then `next`.

## The fix

Every load carries a **`LoadIntent`**: a monotonic number claimed from `AudioEngine::next_load_intent` **when the intent starts** — before the first await, not before the send. The decoder keeps the highest intent it has been handed in `SharedPlayback::newest_load_intent` and drops any load below it, so a preparation that finishes late is discarded instead of taking over.

Three decisions are worth reviewing on their own merits:

**Arbitration at receipt, not at execution.** `drain_commands` returns `ControlFlow::LoadNext`, which stops the current track *before* anything looks at the payload. A stale load discarded later than that would leave the decoder with nothing playing at all — worse than the bug. The three receive points (the idle loop, the drain between packets, the drain inside the pause loop) all go through `accept_load`, and the three identical load arms in each drain collapse into one.

**Compared against what arrived, never against the allocation counter.** An intent can be claimed and never sent — Next on an exhausted queue, a track row that has since vanished. Arbitrating against the counter would let such an intent silence a load that is still on its way; arbitrating against the high-water mark only drops a load when a newer one has actually been delivered.

**Not in `AudioEngine::send`**, which the issue suggested. `send` is not in fact the single point every load passes through: the three output-rebuild resumes and the analytics auto-advance hold a cloned `cmd_tx` and bypass it entirely. The decoder is the real funnel — it receives everything — so that is where the comparison lives.

## Coverage

All seventeen sites, not the two a report happens to name. `LoadIntent`'s field is private, so a load command cannot be built without asking the engine for an intent first: `player_actions` (2), `commands::player` (6), `audio::engine` (6), `remote::playback` (4), `audio::analytics` (1). Where a load is produced behind a guard or an early return, the intent is claimed after it — `player_previous` past the restart-in-place rule, `player_next` past the remote-queue branch — and where several sends share one preparation (`remote::playback::play_current`) they share one intent.

Two things stay outside:

- **`SetNextTrack`** arms the gapless prefetch rather than taking over. Dropping one would cost a gapless hand-off to prevent nothing audible.
- **The decoder's own repair path** (a reconciled file that will not open, falling back to the server) re-sends under the **same** intent it was handed. It is continuing that load, not starting a newer one.

## Validation

- `cargo fmt --check`, `cargo check --workspace --all-targets`, `cargo clippy --workspace --all-targets -- -D warnings` — all clean locally on Windows, so the WASAPI paths are linted too.
- Seven new tests in `audio/decoder.rs` covering the issue's sequence verbatim, the ordinary case (newer intent also arriving later), a re-checked `pending_cmd`, a claimed-but-never-sent intent, and non-load commands never moving the high-water mark. The app crate's test binary cannot run on Windows (`STATUS_ENTRYPOINT_NOT_FOUND` on the Tauri DLLs), so `Rust (ubuntu-latest)` is the only job that executes them — the comparison's direction was sanity-checked standalone in the meantime.

## Docs

[Ordering the loads](docs/features/playback.md) under § Queue, the invariant "every load command carries its intent", and its one-line form in `CLAUDE.md`.

## Summary by CodeRabbit

* **Améliorations**
  * La lecture respecte désormais l’ordre des intentions de l’utilisateur : une action récente ne peut plus être remplacée par le résultat tardif d’une opération précédente.
  * Ce comportement couvre les pistes locales, les téléchargements, le streaming et les reprises après récupération ou changement de périphérique.
  * Les chargements obsolètes sont ignorés sans interrompre la lecture ni modifier l’état de pause.
  * Les commandes de navigation non liées au chargement restent inchangées.

* **Documentation**
  * Ajout d’une documentation sur l’ordonnancement et la gestion des chargements audio.


InstaZDLL:
@coderabbitai All four verified against the current branch. Three of them were this PR's own defect turned back on itself — an intent claimed later than the moment it belongs to — so they are fixed in `2ffadaa1`; the fourth is fixed for the path where it strands the player, and filed for the rest.

**1. The auto-advance's intent (`analytics.rs`) — fixed, as you describe.** You are right that claiming it in `handle_message` puts it after the channel hop *and* after the `play_event` write, the repeat-mode read, the queue advance and the gain lookup. `TrackEnded` and `RemoteTrackEnded` now carry the intent, claimed by the decoder in `handle_playback_outcome` where the track actually ended, and `remote::playback::advance` takes it from its caller instead of minting its own — so a Next is ordered from the key press and the auto-advance from the end of the track. Two consequences worth naming: the allocation counter moved to `SharedPlayback` so the decoder thread can claim one without an engine handle (`LoadIntent::claim`, which keeps the tuple field private to the engine module), and `player_next` / `player_previous` now claim before their remote branch rather than after it, which is where the intent belonged anyway.

**2. The raw `?cmd` in the drop log — fixed at the source.** The leak is worse than the one call site you found: `AudioCmd`'s `Debug` printed the URL verbatim, and that impl is what *every* `?cmd` / `?err` logger reaches, so the redaction belongs there rather than in my log line. `LoadUrlAndPlay` now goes through the existing `http_source::redact_url`, which strips userinfo, query and fragment — the query is exactly where a remote-queue ticket's HMAC signature rides. I kept the variant in the log: with the URL redacted it is diagnostic, not sensitive.

**3. The three rebuilds (`engine.rs`) — fixed, and this was the worst of the four.** The intent was claimed in the resume block at the bottom, past the stop, the device open and the producer swap. An exclusive open costs tens to hundreds of milliseconds, so a track the user picked during a device switch would have been dropped in favour of the resume — the exact inversion this PR is about, with the widest window of any site. It is now claimed beside the `track_id` / `position_ms` snapshot it re-dispatches, in all three methods, and the later claim is gone.

**4. `resume_last`'s effects — fixed for the damage, filed for the rest.** The literal remedy isn't available: acceptance happens on the decoder thread, asynchronously, and the producer cannot await it. But the concern is real and I'd split it in two.

The half that strands the player is `set_state(Loading)`, and it exists only in `resume_last`. Two changes there: it gives up entirely when `load_intent_superseded(intent)` says a newer load has already been handed over — it mutates nothing before its send, so bailing out is total and clean — and the `Loading` write now goes through `SharedPlayback::try_set_state`, a compare-exchange, so it can never overwrite a transition the decoder published for another track. The residual window after the check is a few instructions instead of a database read, and the compare-exchange closes what that window could have done. The failed-send restore is guarded the same way, which it was not before this PR — a blind `set_state(previous)` there could already clobber the decoder.

The other half is the label and the queue cursor, which the remaining producers publish before sending. Those write to SQLite first (`queue.current_index`, and `player_play_tracks` refills the queue), so "check then bail" would leave the cursor moved with nothing playing it — worse than a wrong label. That needs the verdict to gate the effects, or the cursor to follow the accepted load rather than the asking producer, and neither is a review-round change: **#632**.

`fmt` / `check --all-targets` / `clippy -D warnings` clean locally on Windows, so the WASAPI paths are linted too.

InstaZDLL:
@coderabbitai Fondé, corrigé dans `e4879798` — et c'est la troisième fois que ce même défaut se trouve dans mon propre code, ce qui vaut d'être dit.

`play_entries` réservait l'intention à la fin de son travail, alors que ses deux appelants lisent leurs entrées en base avant de l'appeler : `start` une lecture de la projection, et `play_track_ids` **une requête par identifiant de piste** — soit tout un album d'allers-retours pendant lesquels un clic sur une piste locale peut atterrir. L'intention réservée après cela dépassait ce clic, et l'arbitrage jetait donc le choix de l'utilisateur.

Le correctif suit exactement celui d'`advance` au tour précédent : la queue partagée ne réserve plus rien, elle reçoit l'intention de son appelant, et `start` / `play_track_ids` la réservent avant leur premier await. Je les ai gardées là plutôt que de remonter dans les commandes Tauri, parce que `remote_play_playlist`, `remote_play_tracks` et `remote_queue_jump` sont des enveloppes d'une ligne sans aucun await propre : ces deux points **sont** le clic. `jump_to` réservait déjà au bon endroit, tout ce qui la précède étant synchrone.

Ce qui reste vrai après ce tour : chaque site réserve désormais à l'origine réelle de son intention — la commande pour une action de l'interface, la fin de piste pour l'auto-avance, l'instantané pour une reconstruction de sortie, et l'appelant pour les deux queues partagées du module distant.

`fmt` / `clippy -D warnings` propres en local Windows.

InstaZDLL:
@coderabbitai Fondé, et c'est le retour le plus sérieux de la série : ce n'est pas une étiquette, c'est le transport. Corrigé dans `8603b1dd`.

Installer une session distante n'est pas un acte neutre — tant qu'une session est active, elle prend la main sur suivant / précédent et sur la fin de piste pour **toutes** les surfaces. Et `remote_playback` documente précisément l'invariant que cela casse : « la session est effacée à l'instant où une piste de bibliothèque devient courante (`emit_track_changed`), donc une session périmée ne peut jamais détourner l'avance suivante ». L'ordonnancement des chargements lui a ouvert une nouvelle brèche : un démarrage distant dont le chargement est jeté réinstallait sa session **après** cet effacement, et reprenait le transport pendant que jouait la piste réellement choisie.

Deux moitiés, parce qu'une seule ne suffit pas :

- **Avant l'installation**, `play_entries` abandonne si son intention est déjà dépassée. Rien n'est muté à ce point — les entrées viennent d'être lues — donc l'abandon est total, comme pour `resume_last`.
- **Après**, parce que ce contrôle ne couvre pas `play_current` lui-même : émettre un ticket est un aller-retour HTTP, et un choix local qui atterrit pendant laisse la session installée sur une piste qu'elle n'a jamais jouée. L'installation est donc défaite — mais via `RemotePlayback::clear_if`, qui ne retire la session que tant qu'elle est **celle que cet appel a installée**. Un `clear()` sec serait faux : à ce moment-là une session plus récente peut tenir la file, et la jeter échouerait la file que l'utilisateur écoute.

Je n'ai pas pu suivre la lettre du retour — lier l'installation à l'intention validée par `accept_load` supposerait d'attendre une acceptation qui se produit sur le fil décodeur, sans rappel possible. `clear_if` est l'équivalent atteignable : le numéro d'installation est écrit sous le même verrou que la file, donc « est-ce encore la mienne ? » se répond sans course.

Deux tests couvrent les deux moitiés (`clear_if` défait la sienne ; il laisse intacte une session plus récente), et le module porte maintenant les deux versants de son invariant dans sa propre documentation. `fmt` / `clippy -D warnings` propres en local Windows.

InstaZDLL:
@coderabbitai Fondé, et corrigé dans `650560a1`. C'est un défaut que le tour précédent a **introduit** : mon jeton de retour en arrière nommait l'installation, alors que `jump_to` et `advance` naviguent dans une session sans rien installer — le jeton restait donc valide à travers elles.

Le scénario que tu décris se tient jusqu'au bout : un démarrage distant part, son chargement est dépassé **par un saut dans sa propre session**, et au retour du ticket son `clear_if` correspondait encore. Il effaçait donc la file que ce saut est en train de jouer, et plus rien ne pilotait suivant, précédent ni la fin de piste. Mon correctif du tour précédent avait couvert « quelqu'un d'autre a installé une session » et manqué « quelqu'un a navigué dans celle-ci ».

Le compteur nomme maintenant **l'état** de la session plutôt que son installation : `set`, `seek_to` et `step` l'incrémentent tous les trois, donc toute navigation désarme le retour en arrière d'un démarrage antérieur. J'ai vérifié que ce sont bien les seuls mutateurs du module — `clear` mis à part, qui n'a pas besoin d'incrémenter puisqu'un `clear_if` qui correspond encore ne trouve alors rien à effacer.

Deux tests comme demandé, un par chemin de navigation : `clear_if_leaves_a_session_the_user_navigated_alone` (le saut) et `clear_if_leaves_a_session_that_was_stepped_alone` (l'auto-avance), qui s'ajoutent à la paire couvrant l'installation plus récente. Au passage ils réutilisent le constructeur `entry` que le module avait déjà et que mes tests du tour précédent dupliquaient à la main.

La règle est écrite dans la doc du champ, puisque c'est exactement ce qui m'a échappé : **toute mutation qui fait de la session l'affaire de quelqu'un d'autre incrémente la révision**, pas seulement une installation.

Issue #632, opened 2026-09-12
bug: a dropped load leaves the player bar and the queue cursor on the track that lost


InstaZDLL:
Raised during the review of #631, and deliberately left out of it.

Since #631, a load whose `LoadIntent` is older than one already handed to the decoder is dropped — the audio now follows the newest selection. What the *producer* of that dropped load already did on its way there is not undone.

## What is left inconsistent

Every load path publishes before it sends. Two kinds of side effect survive a dropped load:

- **The player-bar label.** `emit_track_changed` announces the track the producer prepared. When its load is then dropped, the UI names one track while another plays, and nothing corrects it until the next real track change.
- **The queue cursor.** `player_next`, `player_previous`, `player_jump_to_index` and `player_play_tracks` move `queue.current_index` (and `player_play_tracks` refills the queue) *before* sending. A dropped load leaves the cursor on a track that is not playing, so the queue panel highlights the wrong row and the next auto-advance steps from the wrong place.

`resume_last` is already handled: it mutates nothing before its send, so #631 made it give up entirely — `load_intent_superseded` — and write `Loading` through a compare-exchange. That fix does not transfer to the paths above, which is why they were left alone: bailing out halfway through would leave the cursor moved and nothing playing it, which is worse than a wrong label.

## Why a pre-check is not the answer on its own

`load_intent_superseded` is best effort by construction — only the decoder knows what it was actually handed, and it can be handed a newer load the instant after the check returns. For a path that has already written to SQLite, "check then maybe bail" also has to answer what to do about the write it already made.

## Shape of a fix

Two candidates, both worth measuring before choosing:

1. **Let the verdict gate the effects.** `AudioEngine::send` claims the dispatch (a compare-exchange on `SharedPlayback::newest_load_intent`) and reports `Dispatched` or `Superseded`; producers publish *after* the send, only on `Dispatched`. The send is a non-blocking push onto an unbounded channel, so moving the emit after it costs nothing measurable — the "emit before dispatch so the PlayerBar updates without waiting" comment in `player_play_tracks` predates the arbitration and would need revisiting, not deleting. The queue write still needs an answer.
2. **Make the cursor follow the decoder, not the producer.** The deeper version: the queue cursor is advanced by whoever's load was accepted, rather than by whoever asked. That removes the class rather than patching each site, and it touches `queue::advance` / `jump_to` semantics — a much larger change.

[...]

Issue #626, opened 2026-09-12
Launch stalls for 15 s when the splash handoff falls back to its timer


InstaZDLL:
## What happens

The main window is created with `visible: false` and revealed from native code when the frontend emits `app://ready`. A 15 s safety net force-reveals it if that event never arrives. That net is not theoretical — it fires regularly on Windows, and when it does the user stares at the splash for a full 15 seconds before the window appears:

```
WARN waveflow_lib: splash handoff: `app://ready` never fired in 15s, force-revealing main window
```

## Evidence

Occurrences in `%LOCALAPPDATA%\app.waveflow\logs\waveflow.<date>`, which covers every launch:

- `2026-09-06` — 12:54:39, 15:44:35
- `2026-09-11` — 09:30:57, 19:44:37
- `2026-09-12` — 03:06:44, 03:08:20, 03:10:49, 03:11:28

Intermittent rather than systematic: plenty of launches on the same days reveal normally. The same warning appears as far back as May 2026 under its older wording (`splash handoff fallback: ... attempt=0`), so this long predates the current code.

Worth stating explicitly, since the reports came in right after #625 merged: **this is not a regression from #625.** Two of the occurrences above (03:06:44 and 03:08:20) predate that merge (03:08:46), and no frontend file changed in it.

## Where the code is

- Listener and timer: [`src-tauri/crates/app/src/lib.rs`](../blob/main/src-tauri/crates/app/src/lib.rs), in `setup` — a `tokio::select!` between `ready.notified()` and `sleep(15s)`, then up to 10 reveal attempts 250 ms apart.
- Emission: [`src/components/common/ReadySignal.tsx`](../blob/main/src/components/common/ReadySignal.tsx), in a `useEffect` after the first React commit.

History: #26 added the splash, #42 moved the handoff to the backend (Linux WebKitGTK raced the heavy first-launch init), #363 made a single task own the reveal and retry until it succeeds.

## What we don't know

Nobody has measured which half fails. Either React genuinely takes more than 15 s to reach its first commit on these launches, or the event is emitted and lost. The two have different fixes, so the next step is to tell them apart:

- timestamp the emission in `ReadySignal` and the reception in the backend listener, and log the delta;
- if the emission turns out to be late, find what blocks the first commit (provider chain, a synchronous `invoke` on mount);
- if it is emitted but never received, the listener registration is racing the webview's readiness and needs to be established earlier or acknowledged.

[...]

PR #633, opened 2026-09-12
fix: four defects before the 1.8.0 beta — output pin, exclusive-toggle race, splash handoff, motion covers

### LINKED ISSUE #615: bug: a motion cover miss sticks for the whole session, and animated webp covers stutter
Reported by email:

> The Motion Cover feature doesn't seem to work for me even though I installed the plugin. I can kind of get it working by putting a video/WebP file into Change Cover → Local File, but the playback is quite laggy.

There are two separate problems behind this.

## 1. An album with no motion cover stays empty for the whole session

`useAlbumMotionArtwork` deduplicates lookups in a process-wide `resolved` map. A successful lookup is remembered even when its answer is `null` (`rememberResolved(key, motion)`), and it is served from that map for the rest of the session. The only limit is eviction beyond 64 entries. There is **no invalidation**: nothing can clear an entry.

The hook's own doc lists the cases that return `null`: no album, offline, no metadata plugin installed, or no plugin has motion artwork for that album. Any of them, once hit for an album, sticks until restart. That includes:

- installing or enabling the plugin after the album was already shown;
- setting a manual `.mp4` from the album page (`MotionCoverPickerModal`) for an album that was already looked up. The cache key includes the album id, but nothing clears the entry after the picker succeeds.

The automatic source is a metadata plugin matched on artist and album text. A miss there is silent: the static cover simply stays.

## 2. Animated covers set through "Change Cover → Local File" stutter

That picker only accepts `jpg`, `jpeg`, `png` and `webp` (`CoverPickerModal.tsx`). A video is refused, so what plays is an **animated WebP**. It is copied as the album's regular cover and rendered as an ``, so it is decoded on the CPU rather than played like a video.

It is also rendered on several surfaces at once, at full size. The immersive view, for one, draws it as its background with `scale-150 blur-3xl` over the whole screen (`ImmersiveView.tsx`), so a full-screen blur is recomputed for every frame.

## Expected

- A motion cover shows up once the plugin is installed, or once a manual `.mp4` is set, without restarting.
- An animated cover doesn't make the now-playing views stutter.

## Directions

- Give `useAlbumMotionArtwork` an invalidation (per album, and globally when metadata plugins change), and call it from `MotionCoverPickerModal` on success.
- Use a static frame (the existing JPEG thumbnails) for blurred backgrounds, and render the animated source once.
- Point people toward the album page's **Motion Cover** picker (`.mp4`) rather than an animated WebP as a regular cover.

## Not confirmed

- Which plugin the reporter installed, and whether it matched their albums.
- The actual decoding cost of animated WebP in WebView2: not profiled.

### LINKED ISSUE #629: Serialise reading the output device with the rebuild that replaces it
Found while reviewing #628, and pre-existing: `set_exclusive_output` reads the active device with one acquisition of the output lock, releases it, then takes the lock again to rebuild.

```rust
let active = self.current_output_device();   // locks, reads, unlocks
// …
let mut guard = self.output.lock()…;         // locks again
```

Between those two, `player_set_output_device` can install device **B** under the same lock and persist `audio.output_device = B`. The toggle then rebuilds on **A**, the name it captured before the swap. The stream ends up on A while the saved preference says B — the next launch moves the user again, and the picker disagrees with what is playing until something else rebuilds.

Same shape as the race fixed in #628 for `reopen_output_device`, where the cure was to stop reading the device outside the lock and let the rebuild resolve it under its own (`RebuildDevice::Pinned`). `set_exclusive_output` cannot reuse that directly: it does its own teardown inline rather than going through `force_rebuild_output`, so the fix is either to route it through the same helper or to hold one acquisition across read and rebuild.

## Why it is filed rather than folded into #628

It predates that PR and belongs to a different method; #628 only stopped its own new caller from reproducing it. Widening that PR would have meant restructuring the teardown path shared with device-error recovery — the code with the #322 / #405 / #604 history — during a review round on an otherwise green branch.

## Worth checking at the same time

Whether any other entry point mutating the output handle reads state outside the acquisition it later rebuilds under. `set_output_device` holds the lock throughout and is fine; the recovery path computes its target deliberately and is fine. Those two plus `set_exclusive_output` are the whole surface today.

The window is narrow — it needs two output-changing actions landing together, which in practice means two clicks in the device menu or a click racing the Settings toggle. The consequence is not a crash but the exact confusion #612 was about: what plays and what the UI claims are no longer the same thing.

### LINKED ISSUE #630: The engine forgets the pinned output device when no stream is installed
Raised during the review of #628, and pre-existing: the engine stores the user's output-device choice **only inside `OutputHandle`**. `current_output_device()` reads `handle.device_name`, so when no stream is installed the engine answers "no device pinned" — even though `profile_setting['audio.output_device']` still holds the user's choice.

`self.output` really can be empty: `publish_output_lost_if_gone` exists for exactly that state, and both `force_rebuild_output` and `set_output_device` reach it when a spawn fails after the old handle was released.

## What it costs

- **`RebuildDevice::Pinned`** (added in #628) resolves to `None` in that state, so a forced reopen would target the OS default rather than the pinned device.
- **The picker** computes `is_pinned` from the same source, so after a failed open no row is flagged as pinned and the "pinned, unavailable" marker #628 added cannot appear — the case it was written for.
- **MPD** reads `current_output_device()` too (`mpd/commands.rs`), so it reports no device for the same reason.

The persisted value is read exactly once, at startup in `lib.rs`, and handed to `AudioEngine::new_with_device`. Nothing re-reads it afterwards.

## Shape of the fix

Give the engine its own copy of the pin, independent of the handle: set at construction from the persisted value, updated by `set_output_device` when a pick succeeds, and used as the fallback wherever the handle is absent. `RebuildDevice::Pinned` then resolves from the handle first (which keeps #628's locking discipline — the handle is read under the rebuild's own lock) and falls back to that field only when there is no handle, where there is no stream to race with anyway.

Worth doing together with #629: both are about state around the output handle being read in ways that assume it is there.

## Why it is not in #628

It is not reachable through the UI today. With no handle the picker flags the OS-default row as active and produces no pinned row, and the frontend only calls the reopen for the active row — so a user cannot ask for "reopen my pin" in that state; clicking their own device goes through `set_output_device`, which works. #628 did not introduce it either: reading the pin off the handle is how `current_output_device()` has always worked.

InstaZDLL:
Closes #630, closes #629, closes #615. Instruments #626, which stays open.

Four reported or filed defects in one pass rather than four rounds of review. Three commits, one subject each.

## `fix(audio)` — the output pin outlives the handle (#630) and the toggle stops reading outside its lock (#629)

The engine kept the user's device choice **only** inside `OutputHandle`, and that handle really does go away: `publish_output_lost_if_gone` exists for exactly that state, and both `force_rebuild_output` and `set_output_device` reach it when a spawn fails after the old stream was released. In that window the engine answered "nothing pinned" — a forced reopen targeted the OS default instead of the user's device, the picker could not flag the pinned row it had just failed to open (the case #628 wrote the marker for), and MPD reported no device at all. `profile_setting['audio.output_device']` was read once at startup and never again.

`self.output` now holds an **`OutputSlot`**: the handle and the pin, under **one** lock. One lock rather than two because every question worth asking pairs them — what is playing, and what was asked for — and a separate mutex would let a reader match one against a stale view of the other, which is the shape of defect #628 had to fix in the picker. `OutputSlot::pinned_device()` answers from the live handle first and falls back to the seeded pin.

Having the pin under that lock is also what fixes **#629**: `set_exclusive_output` read the active device with one acquisition, released it, then took the lock again to rebuild, so a device pick landing in between installed *and persisted* B while the toggle rebuilt on A. It now resolves its target from the slot after taking the lock — the same cure `reopen_output_device` took with `RebuildDevice::Pinned`.

Two more things the local review caught, both worth stating because the comment I had written promised behaviour the code did not have:

- The pick becomes the pin **only when the switch happened**. `switch_error` means the requested device refused to open and the previous one was reopened in its place, so recording the refused one would name a device nothing is using and aim the next forced reopen straight back at it.
- The same-device early return now only fires **while a handle exists**. Without one `current` is `None`, so a user asking for the OS default matched `None == None` and got a no-op in the exact state they were trying to recover from.

`set_output_device` still compares the *handle*, not `pinned_device()`: with no stream installed, picking the pinned device again has to reach the rebuild.

## `fix(ui)` — the splash handoff (#626)

The 15 s net fires regularly on Windows and the user watches the splash for all fifteen seconds. Nobody had measured which half fails, so this removes the failure class that can be removed and instruments the one that cannot. **The issue stays open.**

The signal's primary transport is now the **`app_ready` command** instead of the `app://ready` event. `generate_handler!` registers a command at build time; the event listener is registered part-way through `setup` — after three blocking database reads and the audio device open — while the `main` webview, created from the config by `Builder::build`, is already running. Events are not replayed, so one handled before that listener exists is gone. The event stays as a second transport, because the rendezvous is one-shot and ignores duplicates, and the frontend retries the command if the call rejects.

Both sides now report their timing: the frontend sends `performance.now()` at its first commit, the backend logs it beside its own elapsed-since-launch. That is what separates "the frontend was genuinely late" from "the signal was lost". The leading suspect for the first is in the issue: `main.tsx` gates its first render on i18next, whose locale chunk is fetched through the asset protocol that `setup` is blocking.

The 15 s net is untouched — it is what keeps a launch merely slow instead of stuck.

## `fix(ui)` — motion covers (#615)

Two problems behind one report.

`useAlbumMotionArtwork` remembers the answer per album, `null` included, to collapse the three or four surfaces that ask at once onto one plugin fan-out. Nothing could clear it, so installing the plugin, enabling it, or setting a manual `.mp4` left the album static until the next restart. It now has a per-album invalidation (called by the picker after a set or a clear) and a global one wired to the **existing** `waveflow:plugin-availability-changed` bus, so enable, disable and uninstall are all covered without a second mechanism. Dropping the entry is only half of it — the surfaces are already mounted — so an invalidation bumps a generation the hook reads through `useSyncExternalStore`, which is what re-runs the lookup. A lookup also carries the generation it started in, so an in-flight request resolving after an invalidation cannot put the entry back.

The stutter is the second half. "Change cover → Local file" takes `webp` and refuses video, so what people reach for is an animated WebP as the ordinary cover — an ``, decoded on the CPU. The immersive view drew it at `scale-150 blur-3xl` across the whole screen, re-blurring every frame. It now renders the pre-resized variant, which the artwork pipeline generates single-frame; the animation still plays once, on the foreground cover.

## Validation

- `cargo fmt --check`, `cargo check --workspace --all-targets`, `cargo clippy --workspace --all-targets -- -D warnings`, `bun run typecheck`, `bun run lint` — clean locally on Windows, so the WASAPI paths are linted too.
- Two tests on `OutputSlot::pinned_device()`. Only the handle-less branch is covered: an `OutputHandle` owns a live output thread and cannot be built in a unit test. That is the branch that regressed.
- **A local CodeRabbit review ran before the first push**, and it took three rounds to come back clean: the conditional pin, the `None == None` early return, the in-flight cache poisoning and a stale sentence in `ui.md` were all fixed before this PR existed.

## Deliberately not here

- **#632** (a dropped load leaves the player bar and the queue cursor on the loser) — the cursor half needs `queue::advance` semantics to change, and fixing only the label would leave the issue open anyway.
- **#627** (follow the OS default output device) — feature-sized, and its macOS half compiles in no CI job. After the beta.

## Summary by CodeRabbit

  - Le périphérique audio sélectionné reste mémorisé lors des reconstructions, pertes de sortie et réouvertures.
  - Les illustrations animées se mettent à jour après un changement de pochette ou de disponibilité.
  - L’arrière-plan flouté utilise une miniature optimisée, tandis que l’animation reste visible au premier plan.

- **Améliorations**
  - Le passage de l’écran de démarrage à la fenêtre principale est plus fiable, avec nouvelles tentatives automatiques.

- **Documentation**
  - Documentation actualisée sur la lecture audio, les illustrations animées et le démarrage de l’interface.


InstaZDLL:
@coderabbitai Les deux sont fondés, corrigés dans `05eb8f36`.

**Le fond flouté de la vue immersive.** Tu as raison et c'était la moitié du correctif #615 annulée par elle-même : je branchais sur `artwork_path` et demandais `size="2x"`, or `resolveArtwork` retombe sur la source complète quand aucune vignette n'existe. Sur un album que le worker de vignettes n'a pas encore traité, la source animée revenait donc exactement là où je venais de l'enlever — décodée image par image et floutée plein écran. Le fond branche maintenant sur les variantes statiques et ne reçoit plus aucun chemin complet : pas de vignette, c'est le dégradé, pas l'animation.

**L'inventaire des mutateurs de la poignée.** Exact aussi : `switch_output_for_track` remplace bien la poignée dans ses deux branches, DoP et PCM, donc « ces trois-là sont tous les points d'entrée qui mutent la poignée » était faux tel qu'écrit. J'ai restreint l'affirmation à ce qu'elle voulait dire — les trois points d'entrée qui changent le **périphérique** — et nommé le chemin DoP, en précisant qu'il respecte la même discipline : il résout sa cible depuis le slot **à l'intérieur** de la prise de verrou sous laquelle il reconstruit. C'est d'ailleurs une des lignes que cette PR a déplacées, donc l'inventaire aurait dû être vérifié à ce moment-là.

Revue locale repassée à 0 retour, `typecheck` / `lint` / `fmt --check` propres.

Issue #634, opened 2026-09-12
An output rebuild can stop a track the user just picked, then resume nothing


InstaZDLL:
Found by review while working on #632 / #627, verified by reading the code rather than reproduced. **Older than #622**, so it is not a regression of the PR that found it — but it is the one hole the load-ordering work does not close.

## What happens

The three paths that replace the output stream — device-error recovery, a device switch, the exclusive toggle, and now the default-device follow of #627 — all end in the same shape (`force_rebuild_output`, `set_output_device`, `set_exclusive_output`):

1. snapshot the session (`current_track_id`, position) and claim a `LoadIntent`;
2. `Stop`, so the decoder unwinds out of `play_track`;
3. open the new device, swap the ring producer;
4. re-dispatch the snapshot as a `LoadAndPlay` carrying the intent from step 1.

A track the user picks between steps 1 and 2 is delivered *before* the `Stop`. So the decoder loads it, the `Stop` then unloads it, and the rebuild's own resume is dropped at receipt by `accept_load` — correctly, because the user's intent is newer.

Net effect: **nothing plays**, while the player bar names the track the user picked. One press of Play recovers it (`resume_last`), so this is an annoyance rather than a loss.

## Why the existing machinery does not cover it

- #622 makes the decoder drop a load older than one already delivered. It does exactly that here, and that is the problem: it drops the only resume anyone was going to send.
- #632 makes a producer claim the dispatch before it publishes. The producer here did nothing wrong — its load is the newest, and it was delivered.
- The rebuild's `Stop` is **not a load**, so nothing arbitrates it.

The rebuild also cannot resume the user's track in place of its own snapshot: only the producer knows the payload (a path, or a URL plus a fallback and a ReplayGain value), and it has already returned by then.

## Repro

Start playback, then trigger a device change (toggle exclusive output, pick another device, unplug the current one) and click a track within the same few hundred milliseconds. An exclusive open costs tens to hundreds of ms, which is the window.

## Shape of a fix — not decided

Two candidates, both touching the most delicate code in `engine.rs` (the #322 / #405 / #604 surface, which wants validation on real hardware across the three exclusive backends):

[...]

PR #635, opened 2026-09-12
fix(audio): claim the dispatch, and follow the OS default output device

### LINKED ISSUE #612: bug: the device menu can show one output while audio plays on another

### LINKED ISSUE #627: Follow the OS default output device when none is pinned

### LINKED ISSUE #632: bug: a dropped load leaves the player bar and the queue cursor on the track that lost

InstaZDLL:
Three commits, grouped on purpose: two of them are the last of the playback-ordering work, and the third is the instrumentation that made the third half of it readable.

Closes #632
Closes #627
Closes #612

## 1. A producer must not publish for a load that will be dropped (#632)

Since #622 the decoder drops a load older than one it has already been handed. The producer of that dropped load had already published, though: the player bar carried its title, and the queue paths had already written `queue.current_index`. So the interface named one track while another played, and the cursor sat on a row nothing was playing — which is where the *next* auto-advance stepped from.

`AudioEngine::claim_dispatch` moves the arbitration one step earlier, to the producer: a compare-and-set on the same high-water mark the decoder reads, called immediately before the first side effect, and a `false` means give up entirely. `queue.rs` grows `peek_step` / `peek_jump` / `commit_index` so "where would the step land" is a separate question from "write the cursor", with the arithmetic in a pure `stepped_index`.

The claim alone does not order what follows it, and the local review was right to insist on this: `commit_index` is a database write, so an older producer can be overtaken *during* it and land its cursor write last; and the runtime is multi-threaded, so "no await between the claim and the send" buys nothing against a producer on another worker. Every local producer therefore holds `AudioEngine::lock_publish` from before it reads the queue through to the send. Under it, claim order **is** publish order — and the claim keeps its own job, because lock acquisition can invert intent order and something has to tell the older one to stop.

Two paths are shaped differently, both to keep a slow thing out from under the lock:

- **`player_play_tracks`** cannot peek — replacing the queue *is* its effect — so it claims before `fill_queue`, holds the lock across the replacement, and hands it back the moment the queue is replaced and the panel told. What follows can be slow in a way no other surface should wait on: a ReplayGain read, and, when the file is gone, an HTTP round-trip to mint a streaming ticket. Both re-take the lock and re-claim immediately before publishing.
- **A remote session** keeps arbitrating by rollback (`load_intent_superseded` + `clear_if`) rather than by lock: minting its ticket is a network call.

## 2. Follow the OS default output device (#627, and with it #612)

With no device pinned, the output opened whatever the system default was *at that moment* and stayed bound to that endpoint for its lifetime — nothing subscribed to default-device changes, and cpal 0.17 exposes no notification API at all.

[`audio/default_device.rs`](src-tauri/crates/app/src/audio/default_device.rs) is the subscription: an `IMMNotificationClient` on Windows, a HAL property listener on `kAudioHardwarePropertyDefaultOutputDevice` on macOS, and a documented no-op on Linux, where PipeWire and PulseAudio migrate a running stream to the new default sink themselves and ours is one of their clients.

What the review of #612 taught, applied here:

- **Only the console role.** Windows fires `OnDefaultDeviceChanged` once per role for one physical change; console is the role both of our open paths ask for (cpal calls `GetDefaultAudioEndpoint(flow, eConsole)`, and so does `wasapi::get_default_device`).
- **The listeners rebuild nothing** — they hand the event to `schedule_default_device_follow`, which takes the same 300 ms backoff and the same `RebuildGate` the device-loss recovery takes. The gate is armed **only once the cheap checks say a rebuild will really happen**: arming for a follow that turns out to be a no-op would open a settle window a real `DeviceNotAvailable` gets swallowed by — dropped, never retried, silence until the user intervenes. That one came out of the local review too.
- **Only while nothing is pinned, decided inside the lock that installs pins.** The engine reads the pin first, but that read is advisory; `RebuildDevice::OsDefaultIfUnpinned` asks again under `force_rebuild_output`'s own acquisition, and that answer is binding.
- **Nothing that changes is left alone, nothing else is touched.** A default that is already the endpoint we are on is not reopened (a rebuild is an audible gap); a default that vanished entirely is not followed (rebuilding onto no device would drop audio we still have, and a device we really lost arrives as `DeviceNotAvailable`, which owns its own recovery). Exclusive output follows the preference minus the #322 session suppression, without resetting it and without recording a flap — the system moving its default is not a device resetting under us.

`windows-core` becomes a direct dependency at the version `windows` itself uses: the `implement` macro expands to `::windows_core::` paths, so the crate has to be nameable from our root. It was already in the lockfile as a dependency of `windows`, so this adds a name, not a crate — `Cargo.lock` gains one edge and no package, so the Flatpak generated sources stay valid.

## 3. Where the startup time goes

The splash handoff logged how long the frontend took to report ready, and a single number cannot say which half was slow. `bundleMs` (the entry module is executing) and `i18nMs` (i18next resolved, and `main.tsx` renders behind it) ride along on the `app_ready` call, so one log line carries the whole timeline. Kept because of how #626 went: three *dev* runs read as release numbers are what made a lost event look like a slow webview.

## Verification

- `cargo clippy --all-targets -- -D warnings`: clean on **Windows** (which compiles the COM half) and on the **macOS machine** (which compiles the CoreAudio half — no CI job does).
- `cargo test --workspace` on macOS: **507** app-crate tests pass, +16 on `c65ee6df` (11 for the claim, 5 for the follow decision). The app-crate test binary cannot start on Windows, so this and the Rust CI job are the judges.
- `cargo fmt --check`, `bun run typecheck`, `bun run lint`: clean.
- Local CodeRabbit review run to the point where the only remaining finding is out of scope, filed as **#634**: an output rebuild can stop a track the user just picked and then resume nothing. It predates #622 — the rebuild's `Stop` is not a load, so nothing arbitrates it — and closing it means restructuring the rebuild resume on the #322 / #405 / #604 surface, with real hardware.

## Not covered

The Windows notification path is not exercised by any CI job (no Windows audio in CI) and was verified by compiling and reading; the macOS path is compiled and linted on the Mac but a default-device change there has not been staged by hand. Linux is deliberately untouched — if a setup turns out not to migrate the stream, that is a measurement to take before writing anything.

## Summary by CodeRabbit

  * Le lecteur suit les changements de périphérique audio par défaut sous Windows et macOS, avec reconstruction différée.
  * Les préférences de sortie exclusive sont préservées lors de ces changements.
  * Les temps de démarrage du bundle et de l’internationalisation sont désormais mesurés.

* **Corrections**
  * Les actions de lecture obsolètes ne modifient plus la file, le curseur ou la lecture.
  * La navigation et l’enchaînement des pistes sont plus fiables lors d’actions concurrentes.
  * La reconstruction audio est correctement protégée, y compris lors d’arrêts anticipés.


InstaZDLL:
@coderabbitai Four of the five are fixed in `ae40db9a`, one is declined. Thanks — three of them were documentation my own change had made untrue, and the fourth was a real hole.

**Fixed — `output.rs`, a default change dropped while the gate is busy.** You were right that this could be discarded permanently, and the reason it matters is that each change is *one* notification, not a repeating signal: moving the default twice inside the two-second settle window left the stream on the intermediate device for good. `follow_os_default_output` now returns `FollowOutcome::Deferred` instead of swallowing the `try_arm` refusal, and `schedule_default_device_follow` comes back after the settle window, three attempts at most. Each retry re-reads the pin *and* the default, so overlapping tasks converge on the current state rather than replaying a stale notification, and a follow with nothing to do costs nothing. I also stopped a panicking blocking task from counting as a settled follow.

**Fixed — `engine.rs`, the `claim_dispatch` rustdoc.** Worse than a style problem: the block was contiguous with the next method's comment, so the whole thing attached to `load_intent_superseded` and `claim_dispatch` had no documentation at all. Moved above the method it describes.

**Fixed — `state.rs`, `newest_load_intent`.** It still claimed the decoder thread was its only reader and writer, which my own `try_claim_load` had made false. It now says there are two kinds of writer — the decoder for every load it accepts, and a producer through `try_claim_load` from `AudioEngine::claim_dispatch` (the Tauri commands and the analytics task) — and states the invariant: release on every write, acquire on every read, and the producer side is a compare-and-set, so a claim can only move it forward.

**Fixed — `playback.md`, what the lock covers.** The paragraph predated the widening and was simply wrong. The lock is taken *before* the `peek_*`, which is the point: it ties the committed index to the queue that index was read from. The ReplayGain lookup is inside the serialized section at the stepping paths, because it happens after the lock is taken. What stays outside is the profile snapshot and the pool lease — plus, at `player_play_tracks` only, the gain read and the streaming ticket, after the lock is deliberately handed back.

**Declined — comparing `cpal::DeviceId` rather than display names.** `DeviceTrait::id()` does exist in cpal 0.17.1, but `opened_device` is matched against the *picker's* ids, and on Linux those come from the ALSA hint database (`snd_device_name_hint`) rather than from cpal at all — cpal enumeration is avoided there on purpose, it opens every card and costs seconds. The persisted `profile_setting['audio.output_device']` is a display name, and all three exclusive backends resolve by name (`Device::get_friendlyname`, `resolve_hw_device`, `get_device_id_from_name`). This is the identity-space argument from the #612 review: `opened_device` has to live in the selector's identity space, and a value from another space means nothing is ticked at all.

For `should_follow_default` specifically, the name comparison is only an optimisation — "don't reopen the endpoint we are already on". A mismatch costs a redundant rebuild on a change that really happened, never a missed one, and that is documented at the function. Comparing an id against a name would make it strictly worse.

**One finding from the local review deliberately not in this PR:** the same protocol applied to the *rebuild resumes*. It is real, but it predates #622 — the rebuild's `Stop` is not a load, so nothing arbitrates it — and closing it means restructuring the resume on the #322/#405/#604 surface with real hardware on three backends. Filed as #634 with both candidate fixes.

PR #637, opened 2026-09-12
feat(library): browse by folder, and find Chinese titles by pinyin

### LINKED ISSUE #578: feat: browse the library by folder tree, not just by library root
Requested in [discussion #557](https://github.com/InstaZDLL/WaveFlow/discussions/557) (item 1), with screenshots.

## What problem does this solve?

WaveFlow has a **Folders** tab, but it lists the library *roots* you configured — `FolderRow` carries `id`, `path`, `last_scanned_at`, `is_watched`, `track_count` ([`browse.ts:232`](https://github.com/InstaZDLL/WaveFlow/blob/main/src/lib/tauri/browse.ts)). That is folder **management**, not folder **browsing**.

People who organise their music on disk — `Artist/Album/`, or by genre, or by rip source — currently lose that structure entirely once the library is scanned. They can reach a track by album, artist, genre or playlist, but not by where it actually lives. For a player whose whole premise is local files, that is a real gap, and it is one of the most common expectations carried over from file-oriented players.

The requester's case: *"I hope we can sort by folder, it would be easier to listen to songs of the same type."*

## Shape

A browsable tree, not a flat list of roots. What that means concretely:

- **Reachable as a destination, not only as a tab.** Somebody who organises by folder navigates that way most of the time; making it a peer of Tracks / Albums / Artists in the sidebar costs nothing and saves two clicks on every use.
- **A breadcrumb and a way back up.** Path from the library root, and an explicit parent entry rather than relying on browser-style back.
- **Files carry their own facts.** Format and size on each entry — in a folder view the file *is* the unit, and "which of these two rips is the FLAC" is the question being asked.
- **Both densities.** A grid for folders with covers, a list for reading names and sizes. The same grid/list toggle the other tabs already have.
- **Play, queue and add-to-playlist at any level** — the actions the album and artist views already offer, applied to a folder and everything under it.
- **An entry point into batch tag work, scoped to the folder being viewed.** A folder is exactly the selection someone wants to act on ("fix this album's tags"), so whatever batch tooling exists should be reachable from here with that scope already applied, rather than requiring a re-selection.

## Notes on scope

- **No migration needed.** `track.path` is already stored, so the tree can be derived by path-prefix grouping at query time. Whether that stays a query-time derivation or becomes a materialised table is an implementation call to make on measured numbers, not up front.
- **Watch the virtual-scroll invariant.** Every list in the app goes through `@tanstack/react-virtual` + `usePageScroll()` with no nested `overflow-y-auto` ([invariants](https://github.com/InstaZDLL/WaveFlow/blob/main/docs/architecture/invariants.md#virtual-scroll-everywhere)). A tree with expandable nodes has to flatten to a virtualised list rather than nest scroll containers.
- **Decide what a folder's counter counts.** Files on disk, or indexed tracks? They differ whenever a folder holds something the scanner skipped, and a counter that says `0` on a folder containing seven files is worse than no counter.
- Sorting inside a folder should reuse the existing track-table sort rather than inventing a second one.
- New UI strings must be propagated to all 17 locales.

### LINKED ISSUE #579: feat: search Chinese titles by substring and by pinyin
Requested in [discussion #557](https://github.com/InstaZDLL/WaveFlow/discussions/557) (item 2) as "support pinyin search". Measuring it first turned up a **larger defect underneath**: Chinese search is broken today for anything that is not the start of a title, pinyin or not.

## The measured defect

`track_fts` is an FTS5 table tokenised with `unicode61 remove_diacritics 2` ([`20260411120000_initial.sql:322`](https://github.com/InstaZDLL/WaveFlow/blob/main/src-tauri/migrations/profile/20260411120000_initial.sql)). `unicode61` does not segment CJK: an unbroken run of Han characters becomes **one token**. The search command appends `*` to each whitespace-separated word ([`commands/track.rs:240`](https://github.com/InstaZDLL/WaveFlow/blob/main/src-tauri/crates/app/src/commands/track.rs)), so prefix matching works — but only from the first character of the run.

Reproduced against SQLite directly, with a title of `中国人民解放军`:

| Query | Position in the title | Result |
| --- | --- | --- |
| `中国*` | start of the run | found |
| `人民*` | middle of the run | **nothing** |
| `稻香` in `周杰伦 - 稻香` | own run, split by ` - ` | found |

So a user with a Chinese library can only find a track by typing it from its first character. Typing correct characters from the middle returns nothing at all — silently, which is the worst shape for a search failure.

## Proposed solution

Two parts, and the first is the one that fixes the defect:

1. **Substring search for CJK.** FTS5's `trigram` tokenizer (SQLite ≥ 3.34) gives substring matching and would make `人民` find the title above. Its cost — a larger index, and a 3-character minimum — needs measuring against a real library before committing to it.
2. **Pinyin, including initials.** Index a transliteration alongside the characters so `zhongguo` and the initials form `zgr` both reach `中国人`. This is what the requester actually asked for, and it removes the input-method switch they described.

## Notes on scope

- `track_fts` is **contentless** (`content=''`) and an FTS5 table cannot gain a column via `ALTER`. This needs a new dated migration that creates the replacement table, rebuilds it, and rewrites the sync triggers — there are triggers on `track`, on `artist.name` and on `album.title` ([`20260428000000_drop_metadata_fks.sql`](https://github.com/InstaZDLL/WaveFlow/blob/main/src-tauri/migrations/profile/20260428000000_drop_metadata_fks.sql)). Migrations are immutable once merged, so this is additive by construction.
- Rebuild cost on a large existing library is the thing to measure early; it decides whether this runs at migration time or as a background backfill like the ReplayGain one.
- Traditional characters matter too — WaveFlow ships both `zh-CN` and `zh-TW`.
- Worth checking whether the same tokenizer choice hurts Japanese and Korean, which are also shipped locales.

InstaZDLL:
The last two items of the v1.8.0 milestone, both features, grouped in one PR.

Closes #578
Closes #579

## Browse the library by folder (#578)

The Folders tab listed the configured **roots** and stopped there — folder *management*, not folder *browsing*. Someone who organises their music as `Artist/Album/` on disk lost that structure entirely once the library was scanned. Clicking a root now descends into it, with a breadcrumb, an explicit parent entry, sub-folders as cover tiles or as rows with names and sizes, and the folder's files below.

The tree is derived at query time from `track.file_path` — nothing materialised, no migration. Three commands: `browse_folders` (the directories directly inside a path, with recursive counts and sizes, plus the breadcrumb's anchors), `list_folder_tracks` (the files of one directory, in the Tracks tab's own row shape), `folder_track_ids` (everything beneath a path, in path order, for the actions that take ids).

Four decisions carry it:

- **A band, not a `LIKE`.** `folder_prefix_range` returns `[dir + separator, dir + separator+1)`, so the scan walks the `UNIQUE (library_id, file_path)` b-tree, `Rock` cannot match `Rockabilly`, and a folder with a `%` in its name needs no escaping — an `ESCAPE` clause would cost the index anyway. The offsets handed to `substr` / `instr` are **character** counts, because that is what SQLite counts.
- **A folder counts indexed tracks, recursively** — not files on disk. The tree is derived from what the scanner indexed, so a directory holding only skipped files never appears, which is why no row can read "0 tracks" (the failure the issue warned about).
- **The cover is the first file's, by path**, through a bare column beside `MIN(rest)` — a documented SQLite guarantee rather than an accident, pinned by a test so a change there fails the build instead of quietly reshuffling covers.
- **No second table.** The existing call became `renderTrackTable(rows, busy, onPlay)`, used by the Tracks tab and by the folder view, so the folder inherits the columns, the persisted sort (the same `library_track_order_clause` on the Rust side), the context menu and the Properties modal. `rows` is also what selection and single-click play range over, so each caller's clicks act on its own list.

Folder actions: play, queue, play next, add to a playlist, create a playlist, and the **batch tag editor scoped to the folder** — usually exactly the selection someone wants to act on. A listing carries the path it was read for, and one whose stamp does not match the folder on screen counts as absent rather than approximate. Root management stays where it was, because that is also where a root is unwatched or removed.

## Find a Chinese title by its pinyin (#579)

The substring half shipped with the trigram rebuild: `人民` finds `中国人民解放军`. This is the half that was actually asked for — typing `zhongguo`, or the initials `zgr`, without switching input method.

Pinyin cannot be derived in SQL, so it is computed in Rust and stored: `pinyin_blob` turns `中国人` into `zhongguoren zgr` — syllables, a space, initials — and `track.pinyin` / `album.pinyin` / `artist.pinyin` hold one blob each.

- **One column per entity.** The two forms share it because a term never contains a space (the planner splits on whitespace), so nothing matches across the boundary, and one column means one `LIKE` per entity instead of two.
- **Written wherever `canonical_*` is written**, which is why it lives in the same module: `upsert_artist`, `upsert_album`, the scanner's track `INSERT`, and the tag editor's title patch. The editor matters more than it looks — the backfill only looks at `NULL`, so a renamed title whose blob stayed behind would be wrong for the life of the library.
- **`track_fts` gained a fourth column** holding all three blobs, fed by triggers that only ever *copy* what Rust computed. One `MATCH` therefore covers text and pinyin alike, with ranking intact, in an index that cannot transliterate anything. The triggers watch `pinyin` as well as the text, or a backfilled row would stay unsearchable until something else touched its title.
- **`NULL` means "not computed", `""` means "nothing to romanise".** The backfill needs that to terminate: most libraries hold no Han at all. It is a background writer, so it parks behind the scan, batches and retries on `SQLITE_BUSY`; what it does not reach stays `NULL` and the next launch resumes there.
- **Both scripts, and a known false positive.** The mapping is by codepoint, so traditional is covered as well as simplified. Japanese kanji romanise as if they were Chinese, which can add a match that has nothing to do with the query — an extra way to find a track, never a missing one. That answers the issue's question about the other CJK locales: the tokenizer change costs Japanese and Korean nothing.

Short initials land on the scanned route by construction (`zg` is two characters, below the trigram floor), so `LIKE_TERM_CLAUSE` grew from three columns to six.

`pinyin` 0.11 is one crate with no dependencies of its own. Its Flatpak source entry is added by hand from the checksum the lockfile already carries, and `check-sources.py` confirms coverage — so the offline build stays whole.

## Verification

- `cargo clippy --workspace --all-targets -- -D warnings`: clean on Windows and on the macOS machine.
- Tests: **526** app-crate and **251** core on macOS, all green. The new ones run against the **real migrator**, which caught two columns I had guessed at in a fixture.
- `cargo fmt --check`, `bun run typecheck`, `bun run lint`: clean.
- 13 new UI keys in all 17 locales, additions only.
- Local CodeRabbit review: 5 findings, all valid, all fixed in the last commit — the one worth naming is a library root stored with a trailing separator (a whole volume is `E:\` on Windows), which found no root at all and cost the browser its breadcrumb anchor.

## Not covered

The folder tree is derived per query rather than materialised; that is the issue's own "implementation call to make on measured numbers", and the numbers say the band scan walks an index over one directory's subtree. If a library ever makes that show, the same three commands can be backed by a table without the UI noticing. Pinyin is not yet in the remote (server) half of the library, which has its own search path.

## Summary by CodeRabbit

  - Parcourez les dossiers de la bibliothèque avec fil d’Ariane, affichage en grille ou en liste, compteurs, tailles cumulées et jaquettes.
  - Lisez, mettez en file d’attente, créez des playlists et modifiez les tags de plusieurs morceaux depuis un dossier.
  - Recherchez les titres, albums et artistes chinois grâce à la romanisation pinyin, y compris les initiales.
- **Documentation**
  - Documentation mise à jour sur la navigation par dossiers et la recherche pinyin.
- **Localisation**
  - Traductions ajoutées pour les commandes du navigateur de dossiers.


InstaZDLL:
@coderabbitai All six are fixed in `3c514f71`. The out-of-diff one was the most valuable — a write site I had missed.

**`scan.rs` — the rescan branch.** Right, and it was the hole in the whole design: the full-rewrite branch sets `title` without `pinyin`, and since the backfill only ever looks at `NULL`, a file whose tags changed on disk would have kept the romanisation of its old title for the life of the library. The blob is written with the title now, and I checked the statement's placeholders against its binds (24 and 24) because a mismatch there only shows up at runtime. The *other* update branch is deliberately left alone: it runs on a file whose bytes are unchanged and never touches the title, so there is nothing to re-romanise.

On the test you asked for: `a_renamed_title_does_not_leave_a_stale_blob` already plays exactly what the scanner now does — `UPDATE track SET title = ?, pinyin = ?` — and asserts the old romanisation stops matching while the new one starts. What no test can reach from here is the scanner *statement* itself, which needs a filesystem fixture and the app-side orchestrator that emits `scan:progress`; the SQL half is what the trigger tests pin.

**The backfill write.** Fixed, and worth stating precisely: the pass reads, then computes, then writes, so a scan or a tag edit landing in that gap has already written a blob for a title this pass never saw. The statement is now `WHERE id = ? AND pinyin IS NULL` and moved into `update_sql`, so the new test plays the very statement the pass runs rather than a copy of it — with a full-pass test beside it (Han filled, Latin marked `""`, a second pass finding nothing, which is what makes this a once-per-database cost).

**`role="menu"`.** Correct — `MenuActionItem` renders a plain button with no `menuitem` role, so the container was promising arrow-key navigation nothing there implements. Removed, along with the trigger's `aria-haspopup`. Worth flagging since you scoped the finding to my container: the two pre-existing menus in this file (`AddToPlaylistPopover` and the album-grid menu) have the same shape. I left them as they are rather than widen the PR, but the fix for all three is the same one line in `MenuActionItem`.

**The breadcrumb with no root.** Real, and the crumbs were worse than useless: with `root_path` null the whole absolute path got split and each crumb's target was built by walking from `""`, so they pointed at relative fragments (`m`, `m/Rock`) that lead nowhere. With no root the trail is now the current folder and nothing else.

**`Pasta anterior`.** My mistake in both Portuguese locales — that says "previous folder", and the button goes to the parent. Now `Pasta superior` in `pt` and `pt-BR`.

**Sizes and the locale.** Done: `formatBytes` takes a locale and formats through `Intl.NumberFormat`, and the browser gets `i18n.resolvedLanguage ?? i18n.language` threaded through `FolderBrowser` and `FolderTile`. The point that decided it is that the browser's own locale is not necessarily the one the user picked in the app, and half the seventeen locales WaveFlow ships write `9,4 GB` rather than `9.4 GB`.

Verified after the fixes: clippy `-D warnings` clean on Windows and on the macOS machine, **528** app-crate tests and 251 core green there, `cargo fmt --check` / `typecheck` / `lint` clean.

InstaZDLL:
@coderabbitai Fixed in `656398a5` — valid, and it was a consequence of the guard I added in the previous commit rather than something that pre-existed it: once the write is conditional on `pinyin IS NULL`, a row somebody else filled between this pass's read and its write is skipped, so the batch's length stopped being the number of rows the database wrote.

`write_batch_once` now sums `rows_affected()`, `write_batch` returns it, and `fill_table` adds that to its total — so the log line reports what happened rather than what was attempted. The loop's termination never depended on the count (it stops when the select comes back empty), so there was no risk of it spinning; what was wrong was only the number it reported, which is exactly the kind of thing nobody checks later.

The guard test now goes through `write_batch` instead of a copy of the statement, so it covers the count as well as the skip: a row taken meanwhile returns zero written.

Verified: clippy `-D warnings` clean, 528 app-crate tests and 251 core green on the macOS machine.

Issue #617, opened 2026-09-11
feat: option to pause when the output device disconnects


InstaZDLL:
Reported by email by a Windows 11 user:

> If music is currently playing and I turn off or put away the earphones, Windows switches the output to the PC speakers and WaveFlow continues playing through them. […] it would be useful if WaveFlow had an option to automatically pause playback when the currently used audio device is disconnected. This could prevent music from suddenly playing through the speakers.

Phones and most desktop players pause when headphones go away. WaveFlow currently does the opposite.

## What happens today

When the stream is invalidated, `notify_device_lost` sets the **displayed** state to `Paused`, but it never sends `AudioCmd::Pause`. 300 ms later, `force_rebuild_output` reopens the output on the current default device and dispatches `LoadAndPlay` at the saved position. For the listener, "paused" is a flicker in the UI and the music moves to the speakers.

Two more gaps:

- **No setting exists** for this.
- **Nothing notices a default-device change** that doesn't invalidate the stream. See #612.

## Proposal

A setting, **Pause when the output device disconnects**. When it is on, a device loss:

1. actually pauses (`AudioCmd::Pause`, so `paused_output` is set);
2. still rebuilds the output on the new device, so pressing Play works immediately;
3. skips the resume `LoadAndPlay` of the rebuild.

Should it default to on? That is a product call. It matches what people expect from headphones, but it changes today's behaviour for speaker setups where a USB DAC glitches for a moment.

## Where

The same place as #611, where a paused session is resumed by that rebuild: `notify_device_lost` / the device-error callback in `output.rs` and `AudioEngine::force_rebuild_output` in `engine.rs`. The two should probably land together.

InstaZDLL:
Shipped in #638, merged as f8eb7b77. Left open only because GitHub reads a closing keyword before the *first* issue number in a comma-separated list and not the rest.

What landed: `audio.pause_on_device_loss`, a per-profile setting, **on by default** — that is what people expect from headphones, and the reported behaviour is the other one.

The distinction it rests on: **a device that disappears is not a default that changes.** #627 had just shipped the opposite behaviour — follow the system default when nothing is pinned — so the session is parked only when the rebuild lands on a *different* endpoint (`endpoint_moved`). A device that flaps and comes back is picked up where it was, which keeps #175 intact. `note_device_loss` plus a three-second ownership window stands the default-follow down while a loss is being recovered; without it the pause fell to chance between the two rebuilds.

[...]

Issue #600, opened 2026-09-09
feat: reopen the exclusive output at each track's source rate


InstaZDLL:
#577 gave Linux and macOS exclusive output for ordinary PCM, so the system mixer is out of the path on all three platforms. It deliberately stopped short of the other half.

## What is still true

The sample rate is a **preference**, not a demand. Every backend opens at a rate the *device* offers and the decoder's rubato resampler meets it. So a 44.1 kHz track played on a device sitting at 48 kHz is resampled by us instead of by the system — better, because it is our resampler and nothing else is mixed in, but not the same thing as the source reaching the DAC untouched.

This is why the word "bit-perfect" was removed from the exclusive-output copy in #577. It was never true on Windows either: that backend has always opened at the endpoint's format and let rubato convert. The audio-pipeline pill can still say bit-perfect, but only because it separately checks that the source rate and the output rate are equal — which is a coincidence today rather than something we arrange.

## What this asks for

Reopen the device at each track's own rate, so a 44.1 kHz track plays at 44.1 and a 96 kHz track at 96, with no resampling at all. That is the audiophile meaning of the term, and it is the only version of it we could honestly print.

The DoP path already does exactly this — it **demands** its rate, because a marker cadence that gets resampled is noise, and `switch_output_for_track` reopens the output per track to get it. The machinery therefore exists; what changes is applying it to PCM, where the rate is currently taken rather than asked for.

## Traps

- **Reopening costs a gap.** DoP tracks never crossfade or gaplessly prefetch for this reason. Doing the same for all PCM would silently disable gapless and crossfade on every rate change, which is a worse trade than resampling for most listeners — so this is a **preference**, not a default, and the interaction has to be decided rather than discovered.
- **Release before reopening on macOS.** Hog mode is registered against a PID, so a second open from our own process is admitted onto a device we still hold and renders nothing ([`must_release_before_reopening`](https://github.com/InstaZDLL/WaveFlow/blob/main/src-tauri/crates/app/src/audio/engine.rs)). Per-track reopening makes that path run constantly instead of rarely.
- **A device that refuses the source rate must fall back**, not fail — and the pill must then stop claiming what is no longer true.
- **The flap guard exists for a reason.** A device that resets on every exclusive grab already gives up for the session after a few attempts; per-track reopening multiplies the number of grabs and will meet that guard far more often.

Issue #597, opened 2026-09-09
feat: show when playback quietly falls back, instead of only logging it


InstaZDLL:
Playback can quietly become something other than what the user asked for, and nothing on screen says so.

## Two halves, both real today

**`player:error` reaches a `console.error` and stops there.** [`PlayerContext.tsx:515`](https://github.com/InstaZDLL/WaveFlow/blob/main/src/contexts/PlayerContext.tsx):

```ts
await listen<PlayerErrorPayload>("player:error", (e) => {
  console.error("[player:error]", e.payload.message);
}),
```

That event carries the message from `output::notify_device_lost` — the device vanished mid-playback, the exclusive backend lost its stream, the HTTP source failed. The player pauses and the user is told nothing at all. A dev console is not a user interface.

**And the output silently degrades.** Exclusive output falls back to shared mode on its own whenever init fails — device busy, no format the driver accepts, no exclusive support. That is the right behaviour, and it is invisible: the Settings toggle re-reads the real state, but somebody listening does not have Settings open. Since #577 there are **three** exclusive backends that can each fall back quietly, so the surface for this tripled while the reporting stayed at zero.

Native DoP has the same shape: the toggle can be on, the DAC can refuse the rate, and playback continues as ordinary DSD → PCM without a word.

## Shape

**A badge in the player showing what actually engaged**, not what was asked for. The states worth distinguishing are the ones that mean different things to a listener: shared mode, exclusive engaged, native DoP, exclusive requested but refused, and a device error that stopped playback. The audio-pipeline popover already computes most of this — `exclusive_active`, `dop_active`, the processing chips — so the badge is a summary of state that exists rather than new plumbing.

**And a real surfacing of `player:error`** — the message is already written for a human and reaches the frontend unused.

## Traps

- **Say what happened, not what failed.** "This device does not accept exclusive mode, playing in shared mode instead" is actionable; "AUDCLNT_E_UNSUPPORTED_FORMAT" is not.
- **A fallback is not an error.** Falling back to shared mode is normal operation and must not read as a fault; losing the device is a fault. The badge and the error path are different registers.
- **Do not make it nag.** Something a user has already chosen — they turned exclusive on, it does not work on their hardware — should be visible once and legible on demand, not repeated at every track.
- New strings go to all 17 locales.

Issue #594, opened 2026-09-09
feat: filter virtual ALSA aliases out of the device list


InstaZDLL:
The Linux output list shows entries that are not devices. ALSA's namespace carries `plughw:`, `surround40`, `iec958`, `sysdefault` and friends — aliases and routing layers over the same hardware — so one sound card appears several times under names that mean nothing to the person choosing.

Worse for us specifically: several of those aliases route through the mixer, so picking one silently defeats the exclusive output the user just turned on.

## Shape

Filter the aliases out of the presented list, matching on the driver field as well as the name — the same card reached through two different alias families would otherwise survive one check and fail the other. Two endpoints whose raw names collide should also stop being presented as one entry.

## The constraint that matters

**Do not solve this by opening devices.** Our enumeration deliberately uses ALSA's hint database rather than cpal's `output_devices()`, because probing every PCM costs a one-to-two second freeze and floods stderr with `pcm_dmix` / `pcm_route` noise. That shortcut is load-bearing and predates this issue. The filter has to work from what the hints already carry.

Related, and worth deciding together: what the picker shows for a device that is a real alias someone might legitimately want — `default` is not hardware, but it is the right choice for most users most of the time.

PR #638, opened 2026-09-13
feat(audio): output device batch — rebuild, device loss, capabilities and rate matching

### LINKED ISSUE #634: An output rebuild can stop a track the user just picked, then resume nothing

InstaZDLL:
## What this is

The first grouped batch of the v1.8.0 milestone: everything that touches the output device and the engine's rebuild path. Six issues, one commit each, because they all land on `audio/engine.rs` + `audio/output.rs` + the three exclusive backends and would otherwise conflict with each other.

Closes #634, #617, #597, #594, #593, #600.

## The rebuild puts back what the decoder holds (#634)

A rebuild owes the session something after the `Stop` it has to send, and it used to owe it a snapshot taken *before* that stop. A track picked during the open — tens to hundreds of milliseconds, exclusive — was delivered first, loaded, unloaded by the `Stop`, and the rebuild's own resume was then dropped on arrival for carrying an older intent. **Nothing played**, while the player bar named the track the user had just picked.

The decoder now records the load it *accepts* in `SharedPlayback::last_load` — `accept_load` is the single funnel every load passes through, the auto-advance included, which sends straight down the channel without going through `AudioEngine::send`. A rebuild re-dispatches that, **with the load's own intent** rather than a fresh one: an already-delivered intent is accepted again, while a newer pick still supersedes it. Minting a new one there would outrank a selection that had claimed but not yet reached the channel.

Two things fell out of it: the resume no longer goes to the database for a file path (so it is synchronous, and keeps the **source** the play came from instead of crediting it to `device-rebuild`), and the position — which has to be read before the stop, since opening the replacement writes its own sample rate into the shared block — is now stamped with the load it belongs to.

## A device that goes away no longer moves the music (#617)

`audio.pause_on_device_loss`, **on by default**, parks the session when the rebuild lands on a *different* endpoint: headphones off, Windows falls back to the built-in speakers, and the album no longer keeps playing out loud. A device that flaps and comes back is reopened and picked up where it was — the automatic recovery of #175 is untouched — and an endpoint neither side can name never counts as a move.

Parking now holds the load itself, which is what lets a radio station or a server track be parked at all: the persisted resume point is always a *library* track's, so a paused stream used to be resumed unconditionally rather than come back as the last local track. `resume_last` prefers the park over that resume point.

**The tension with #627 is arbitrated**, not left to chance: unplugging raises both signals at once — the stream breaks *and* the system default moves — so the loss is stamped at the moment of the error and the default-device follow stands down for three seconds. Without that, the pause landed or didn't on a coin toss between the two rebuilds.

## Playback says what it had to do (#597)

`player:error` reached a `console.error` and stopped there. Two registers now, deliberately different:

- **a fault** keeps its technical message (that is what a bug report needs) and gains a `kind` the UI turns into a sentence in the user's language, with the message in the tooltip;
- **`player:notice`** is the other one — exclusive refused, DoP refused, playback parked because the device went away. Nothing failed; playback simply contradicts a choice. Emitted **only on a transition**, so a DAC that never accepts exclusive says so once instead of at every track.

`AudioEngine::output_mode` is the same question as state (`shared` / `exclusive` / `dop` / `exclusive-refused`), computed under one acquisition of the output lock so the badge, the notice and the Settings card cannot disagree. `exclusive-refused` is precisely the state the UI could not name before.

## ALSA's routing layers leave the device list (#594)

One card came back as six lines — `hw`, `plughw`, `dmix`, `dsnoop`, `surround40`, `front` — and three of those route through the software mixer, so picking one silently defeated the exclusive output the user had just turned on. `present_alsa_hints` drops the routing families, drops a wrapper over hardware already on the list (paired by the **card token**, since `front:CARD=PCH,DEV=0` and `hw:CARD=PCH,DEV=0` share no text), and disambiguates two rows that would read identically. `default`, `pulse` and `pipewire` stay — not hardware, but the right answer for most people, and on a PipeWire desktop often the only one that works.

The enumeration itself is untouched: it still reads the hint database rather than opening every PCM, which is what keeps the menu from freezing for one to two seconds.

## Each device says what it accepts (#593)

`audio/capabilities.rs` asks one device at a time, **on demand and never during enumeration**, and memoises only real answers — a device another client was holding must be asked again rather than remembered as unavailable for the run. Each row in the picker carries its deepest format, its top rate and a Hi-Res marker; the sheet behind it adds the channels, the buffer, the formats as a set, and the rates **grouped into named tiers**, with the numbers on the tooltip.

The sheet says which question was asked, because a driver's declaration and an exclusive-mode acceptance are different things: WASAPI offers each pair to the driver (the same call the real open makes, and it initialises nothing, so it is safe even while we hold the device), ALSA reads the raw `hw:` parameters **non-blocking** (so the device we are playing on exclusively cannot answer, and says so), CoreAudio reports what the device declares to the HAL.

## Playing at the track's own rate (#600)

`audio.match_source_rate`, **off by default** — not because it is worse, but because reopening the device costs an audible gap and every rate change becomes one. With it on, and only while the output owns its device, the decoder asks for the track's rate after the container has said what it is and before the first packet.

`RequestedFormat` carries either demand, and a refusal means different things: DoP fails the open, a rate request falls back. The guard that keeps it from thrashing is `last_requested_rate` — the next track compares against what the last open **asked** for, never against what it got, because a device that refused 96 kHz has its own rate installed and comparing against that would rebuild the output for every single track, forever.

## Verification

- clippy `-D warnings` clean on **Windows** (the COM half, including the new capability probe) and on **macOS** (the CoreAudio halves of #593 and #600, which no CI job compiles);
- **550 app-crate tests** green on macOS, `cargo fmt --check`, `typecheck` and `lint` clean;
- the WASAPI candidate ordering of #600 is unit-tested, and since no CI job runs this crate's Windows-only tests, those two were also extracted into a throwaway crate and run there — where the first version of the dedupe rule turned out to be wrong;
- 34 new strings across all 17 locales.

## Not covered, on purpose

macOS re-clocks the device for #600 through the same physical-format machinery DoP uses, with the same restore guard; that path has been compiled and linted on the Mac but **not heard on real hardware**. Same caveat as DoP: no CI job compiles macOS at all.

## Summary by CodeRabbit

  - Affichage détaillé des capacités, formats, fréquences et niveaux de qualité des périphériques audio.
  - Indication du mode de sortie utilisé : partagé, exclusif, DoP ou refusé.
  - Options pour suivre la fréquence d’échantillonnage du morceau et mettre la lecture en pause lors d’une perte de périphérique.
  - Alertes de lecture traduites, avec détails techniques et notifications distinctes.

- **Améliorations**
  - Restauration plus fiable de la lecture après reconstruction de la sortie.
  - Sélection optimisée des formats et fréquences compatibles, avec repli automatique.
  - Filtrage et présentation améliorés des périphériques audio disponibles.


InstaZDLL:
@coderabbitai All ten are fixed in `eb45dd81`. Three of them changed behaviour rather than wording, and they are the ones worth writing down.

**`buffer_frames` had no defined quantity.** You are right, and the sheet was worse than imprecise: ALSA reported the largest period the card accepts, WASAPI the *default* period converted to frames, and the two sat under one label inviting a comparison that means nothing. Rather than pick one and convert the other, I picked the one quantity both platforms can actually answer — `min_period_frames`, the smallest period the device will take. That is the device's latency floor, which is what an audiophile sheet is being asked for, and CoreAudio still reports nothing there because a buffer size is the client's to choose rather than the device's to declare.

**The macOS re-clock found nothing on most DACs.** This was the sharpest catch of the review, and it is the whole feature: `find_matching_physical_format` matches the depth and the number type exactly — only the flags are ignored — so a request for 32-bit float alone returns `None` on any device whose physical formats are integer, which is most of them. The `None` branch kept the current format, so `match_source_rate` was a toggle that quietly did nothing on the platform where I could least easily test it. It now walks `F32`, `I32`, `I24`, `I16` and takes the first the device offers **at the requested rate** — the rate is the demand, the depth is whatever it comes in — the same shape the DoP path already had.

**The parked session was spent before the dispatch was claimed.** Also right, and your reasoning is the part I had missed: I had thought the loss was benign because a newer selection plays something else, but a producer can claim and then give up without loading — the queue step whose cursor write fails is a real example — and the park would be gone with nothing playing. The next Play then falls back to the persisted resume point, which is always a library track: the radio-comes-back-as-a-local-track case #617 exists to fix. It is now read without being consumed and cleared only after a successful `claim_dispatch`.

The rest, briefly:

- **Persistence errors were swallowed.** Both commands answered `Ok` on a missing profile or a failed write, so a setting the UI believed it had saved came back undone at the next launch. They persist first and propagate now, which also means nothing is applied when the write fails. The four neighbouring audio commands have the same shape; I left them alone rather than widen the batch, but they are wrong in the same way.
- **`pause_on_device_loss` did not follow the profile.** It lives on the engine rather than in `SharedPlayback`, and I only seeded it at boot, so the second profile of a session kept the first one's answer. It is re-read where `dsd_dop` and `match_source_rate` already are — the block that runs on a switch.
- **`max_channels` under Windows was the mix format's count**, which is stereo on plenty of endpoints that are not. It now takes the widest layout the device reports for itself, which is what the two other backends answer with. The probe still runs at the mix layout: that is the layout exclusive mode wants (#409), and it is a separate question from what the sheet publishes.
- **The ALSA disambiguation could still print one line twice** — two devices of the same card share the card token. It is used only where it tells the colliding rows apart, and the id otherwise, with a test for that case.
- **The device menu cached an unavailable answer.** The backend memo already refused to, and I had put the same cache back one level up in the component. Only real answers are remembered now.
- **The two toggles took the fire-and-forget shape** rather than the chained, profile-guarded write `handleToggleDsdDop` established for a profile-scoped setting. Fixed, queue and rollback guard included.
- **The subtitle promised "no resampling at all"**, where the device may refuse the rate and be met by the resampler exactly as before. Reworded in the source and the sixteen translations.

Declined, with the reason: the two spinners in the device menu use `animate-spin` with no reduced-motion override — and so do the fifty others in the app. Reduced motion is handled here by per-animation rules in `app.css`, so honouring it for spinners is one rule covering all of them, not a utility class on the two this branch happens to add.

Verified after the fixes: clippy `-D warnings` clean on Windows and on the macOS machine, **551** app-crate tests green there, `cargo fmt --check` / `typecheck` / `lint` clean.

InstaZDLL:
@coderabbitai review

InstaZDLL:
@coderabbitai Two of the four are fixed in `f7f37b14`; the third was already done and the fourth I am declining with a reason.

**The deliberate rebuilds asked the wrong question.** You are right, and it is the more interesting of the two: `set_output_device` and `set_exclusive_output` branched on "is something loaded" — `Playing` or `Paused` alike — and then re-dispatched it, so **picking a device or flipping the exclusive toggle while paused started the music**. The device-error path stopped doing that in #611, and this branch handed it `park_session`; the other two now run the same `rebuild_resume` and park what they should not start. Choosing an output is an act on the *output*, never a request to play. It predates this branch, but the branch is what makes the rule general, so leaving two of the three paths contradicting it would have been the worse choice.

**The settings panel could paint over a toggle.** Also right. `playerGetAudioSettings` is an IPC round trip, and a click during it was overwritten by the value it had just replaced — the write still landed, so the panel then disagreed with the database until the next hydration. The two toggles this branch adds record that they were touched while a read is in flight, and the read skips exactly those. The same window exists for the eight settings hydrated beside them; it predates this branch and I left it alone rather than restructure that effect here.

**The `ja` / `ko` / `nl` subtitle** was already reworded in `eb45dd81`, one commit before this review ran: all three now carry the condition (`デバイスが受け付ければ`, `장치가 받아들이면`, `als het apparaat die accepteert`), matching the `en` source.

**Declined: `start_ms` on `LoadUrlAndPlay`.** The observation is correct — a rebuild during a seekable server stream restarts it from zero rather than at its position — but it is not a regression: the snapshot this branch replaces ignored the position for URL loads too, with the same comment about a stream starting where the server is. Closing it means adding a position to the command contract, carrying it through the resume, and seeking an HTTP source in the middle of a device rebuild, with the range-request behaviour that implies. That is its own change with its own testing, not a line in this one, so it is stated in the commit rather than half-done here.

Verified after the fixes: clippy `-D warnings` clean on Windows and on the macOS machine, **551** app-crate tests green there, `cargo fmt --check` / `typecheck` / `lint` clean.

InstaZDLL:
@coderabbitai All three are fixed in `a5a275d6`.

**The resume decision was taken before the open.** Right, and it is #634 again from the other side: an exclusive open costs tens to hundreds of milliseconds, and a Pause the user hits during it reaches the decoder first — so a decision made from the pre-open state started a track they had just stopped. All three rebuild paths now re-read `rebuild_resume` immediately before acting on it, not only the two you named: the device-error path has the same window, widened by its own 300 ms backoff.

What makes the late read safe is worth stating, because it is the reason the decision was taken early in the first place: the `Stop` these paths send **does not clear either input**. `handle_playback_outcome` fires analytics for an interrupted track and transitions nothing, and `paused_output` is only cleared by a load. So the state and the pause flag still describe what the user last asked for, and reading them after the swap is reading their last word rather than something the rebuild invalidated.

**The panel could stay out of step after a failed write.** Also right: the rollback restores the value the click replaced, and after a failed write that value may never have been persisted either — the case is sharpest right after mount, where it is the default rather than anything the backend said. A failed write now re-runs the hydration read through the effect that already exists, guards included, so the panel is refilled from the backend instead of from a guess.

**And the subtitle still promised too much.** Correct, and in a direction I had not considered: even when the device accepts the track's rate, a playback speed other than 1× feeds rubato a deliberately false source rate, so the resampler is engaged whatever the output is opened at. It now says it *avoids* resampling rather than abolishing it, in the source, the sixteen translations, and the docs — where the speed interaction is spelled out.

Verified after the fixes: clippy `-D warnings` clean on Windows and on the macOS machine, **551** app-crate tests green there, `cargo fmt --check` / `typecheck` / `lint` clean.

InstaZDLL:
@coderabbitai These three landed while `a5a275d6` was being pushed, and they describe `f7f37b14` — all three are already fixed on the branch head. Verified rather than assumed:

- **the subtitle** now reads `リサンプリングを避け` / `리샘플링을 피하고` / `voorkomt herbemonstering` — "avoids" rather than "no resampling", in the `en` source and all sixteen translations, with the speed interaction spelled out in the docs;
- **`rebuild_resume` is re-read immediately before each `match`**, in all three rebuild paths rather than the two named — the device-error path has the same window, widened by its own 300 ms backoff. The pre-open decision is kept only for `was_playing`, i.e. for whether a `Stop` is owed;
- **a failed write re-runs the hydration read** through the effect that already exists, so the panel is refilled from the backend rather than from the value the click replaced.

CI is green on `a5a275d6`: 563 app-crate tests on Linux, Windows green, and clippy `-D warnings` plus 551 tests on the macOS machine, which is the only place the CoreAudio halves of #593 and #600 compile at all.

InstaZDLL:
@coderabbitai review

InstaZDLL:
@coderabbitai review

InstaZDLL:
@coderabbitai Four findings this pass, and I am declining all four. Each was checked against the branch head rather than dismissed on sight.

**The decoder acknowledgement after `SwapProducer`.** The ordering it would buy is already there, and the handshake would cost more than it fixes.

The command channel is a single FIFO queue, so `Stop` → `SwapProducer` → resume keep their order, and every interleaving with a track the user picks meanwhile already ends with the newest selection playing:

- the pick was accepted before the read — `last_load` names it, and the rebuild re-dispatches *it*, which is the whole point of #634;
- the pick is sent after our `Stop` but before our resume — the decoder accepts it, then drops our older intent on arrival;
- the pick arrives after our resume — it is newer, so it supersedes what we just put back.

The one interleaving that really loses audio is a producer that **claims and never sends** (a `Next` on an exhausted queue, a `track` row that has gone). An acknowledgement does not close that: the claim can land just as well after the ack as before it. Meanwhile the wait would be taken while holding the output mutex, so a decoder busy opening a slow source — a network share, a cold USB drive — would stall a device switch for as long as that open takes. That is a real regression against a case I cannot construct.

**The three i18n comments about playback speed.** The observation is factually right: the resampler stays `Passthrough` only when the rates match *and* the speed is 1×, so a speed other than 1× engages it whatever the device was opened at. I checked that rather than assume it.

It still does not belong in this string. The subtitle already says the toggle **avoids** resampling *when the device accepts the rate* — softened for this very finding one pass ago — and the resampler can engage for reasons this toggle does not own: the speed, a DoP stream, a rate the device refused. Enumerating them turns a one-line label under a switch into a specification, in seventeen languages. The interaction is written down where a reader can act on it, in `docs/features/playback.md` beside the feature itself.

**Disabling the two toggles until hydration resolves.** This is the alternative your previous pass offered — "if per-setting tracking is not used, disable both controls" — and I took the tracking: `audioSettingsTouched` records which of the two the user touched while a read is in flight, and the read skips exactly those. Doing both would grey out controls for the length of an IPC round trip without adding a guarantee.

One thing did come out of re-reading the rebuild for you, and it is in `6763d896`: the resume still carried the comment describing the asynchronous database lookup it did before #634 replaced it. A wrong comment in that function is worse than none.

InstaZDLL:
@coderabbitai One of the six is new and it is a real hole; it is fixed in `d9624c5e`. The other five are the previous pass repeated, and my reasons stand — restated briefly at the end so they are on this comment too.

**A pause the decoder has not reached yet now counts as paused.** You are right, and it is the half my last fix did not reach. Re-reading `rebuild_resume` after the device opens closed the wide window — the open itself, hundreds of milliseconds — but `paused_output` is the *decoder's* answer and it lags by design: it is raised when the `Pause` is **processed**, a decode cycle or a ring-poll interval after it was sent. A `Pause` sitting in the channel behind the rebuild's own `Stop` therefore still read as "playing", the rebuild resumed, and the resume cleared the flag and started the music. The pause disappeared.

Rather than wait for the decoder, the engine records the intent where it is already known: `pause_pending`, maintained at the `send` boundary that every `Pause`, `Resume` and user-facing load passes through, and the rebuilds now ask both sources. The two paths that hold the channel directly cannot run against a paused session anyway — a paused track never ends, so the auto-advance has nothing to advance, and a rebuild only resumes when it has just decided the session was not paused.

Three pure tests on the lifecycle. The one worth naming: **a `Stop` leaves the flag alone**. The rebuilds send one, and it says nothing about whether the user wants to hear anything afterwards; clearing on it would undo the very pause this carries.

The five repeats, unchanged:

- **the decoder acknowledgement** — the channel is a single FIFO queue, so the ordering is already there and the newest pick wins in every interleaving; the one case that really loses audio (a producer that claims and never sends) is not closed by an ack, and the wait would be taken while holding the output mutex, so a decoder busy opening a slow source would stall a device switch;
- **the three i18n comments about playback speed** — the fact is right and I verified it (`Passthrough` only survives when the rates match *and* the speed is 1×), but the subtitle already says the toggle **avoids** resampling *when the device accepts the rate*, and the resampler engages for reasons this toggle does not own. It is documented in `docs/features/playback.md` beside the feature;
- **disabling the two toggles while the audio settings hydrate** — I took the other half of the choice you offered two passes ago, the per-setting tracking, and it is in place. I also checked the neighbours before declining: of the twenty-three switches in that panel, none of the audio settings hydrated by the same effect is disabled during that read. Greying two of them out and not the eight beside them would be an inconsistency the user can see, for a window of one IPC round trip.

Verified after the fix: clippy `-D warnings` clean on Windows and on the macOS machine, **554** app-crate tests green there, `cargo fmt --check` / `typecheck` / `lint` clean.

InstaZDLL:
@coderabbitai Both inline findings were already fixed in `717e3cdb`, pushed just now along with a second commit. Answering them anyway so the reasoning is on the record, and reporting one place where following a finding uncovered more than it described.

**`AudioEngine::send` — the flag and the channel under one lock.** Done, with the restore on failure you asked for: a new leaf `dispatch: Mutex<()>`, deliberately *not* `publish` (that one is an async mutex a producer already holds when it calls `send`, so taking it here would deadlock on the spot). Nothing awaits inside it and no other lock is taken under it. On a failed send the previous value goes back, because an intent nothing was queued for is one the decoder will never see either.

**`PlayerState::Loading` counts as a session.** Done, including the `paused_output` distinction — `Loading` now answers `StayPaused` or `Play` exactly like `Playing` and `Paused`. Worth stating why, since the interesting part is not the resume: the decoder is inside `play_track` from the moment it accepts a load, and a `SwapProducer` that reaches it there is dropped, because both drains fall through to a catch-all on the stated assumption that the engine always sends a `Stop` first. Answering "nothing to do" for a loading track broke that assumption, so the swap was lost and the decoder went on writing into a ring whose consumer had just been torn down — silence until something else rebuilt the output. `Idle` and `Ended` stay outside, where the decoder is parked at the top-level loop that handles a `SwapProducer` properly. Covered by `a_track_still_loading_is_still_a_session`, both branches.

**Six more from a local review pass, in `c07a8cf7`.** Two of them went past what the finding described, and that is the part worth flagging:

- **WASAPI COM.** The finding named `probe_capabilities`, which initialised COM and never balanced it — worst there because the probe runs on a tokio blocking-pool thread, pooled and long-lived, so the apartment outlived the call and the reference count rose once per probe for the life of the process. Writing the RAII guard exposed the same hazard on the render thread: both event loops called `deinitialize()` by hand with their `client` and `render` interfaces still alive in scope, releasing them *after* `CoUninitialize`, and the early return a busy device takes every time never balanced the init at all. Both loops now hand COM to the guard, which is declared above everything it protects so it drops last.
- **ALSA probe.** The finding named the unpinned call, where `resolve_hw_device` answers `hw:0,0`. But it answers `hw:0,0` for the literal name `default` too, and the listing keeps a `default` row on purpose (#594) — so the sheet describing card 0 under a name the plug layer merely *routes* was reachable straight from the menu, not only from an internal call. Both are refused now.

The rest: `player_probe_output_device` documented that it cannot fail and then propagated its blocking task's join error (a panicked or cancelled task is one more way of not being able to ask, so it comes back as the same unavailable answer); `DeviceMenu` left a row on "reading capabilities" forever when the probe rejected, now one write site for both outcomes so the only-real-answers-are-memoised rule applies to either; `PlaybackAlertToast` carried `z-100` without a portal while mounting inside the skin's motion wrapper; and the `matchSourceRate` subtitle ended on an apposition with no verb in fr, en and hi.

One note on the `DeviceMenu` fix, because the first version was wrong in a way only the linter saw: writing the fallback inside the `catch` made `react-hooks/set-state-in-effect` stop reporting on the *whole component*, the only symptom being a neighbouring `eslint-disable` turning up as unused. Bisected to confirm the cause, then restructured so both outcomes compute a `caps` value and there is a single `setCapabilities` call outside the `try`.

Validated: `cargo fmt --check`, `clippy -D warnings` at zero on Windows **and** on macOS (CoreAudio is compiled nowhere else), 555 app-crate tests green on macOS, `typecheck` and `lint` clean.

InstaZDLL:
@coderabbitai Pushed `220c9be4`. One finding declined with reasons, one carried into its own issue, and one CI failure of my own fixed.

**Extending the dispatch lock over the rebuild sequences — declined, but it found a real defect.**

Declined as proposed, for three reasons. `AudioEngine::send` is called from async command handlers; holding its mutex across a rebuild would park tokio workers for the length of a device open, which in exclusive mode is tens to hundreds of milliseconds. It would also undo what the last three review rounds were spent on: a `Pause` blocked for the duration of the open cannot set `pause_pending`, so the rebuild would read a snapshot taken before the user acted — exactly the family #634, the pre-open resume decision, and #611 each turned out to be. And the field is documented as a leaf lock held across two statements with no await and no other lock under it; the rebuild paths take the output guard and reach `park_session`, which emits events and writes to the database.

The race it points at is real, though, and worse than the comment suggests: it is not only a pause that can interleave, it is a **load**, and the consequence is silence rather than a mistimed decision. The decoder discards a `SwapProducer` that reaches `drain_commands` while a track is decoding, on the stated assumption that the engine always sends a `Stop` first. The `pre_release` branch — taken whenever the current stream is exclusive, per #322 — sends that `Stop` and then opens the device before sending the swap. Pick a track in that window and the decoder leaves `play_track` on the `Stop`, accepts the load, re-enters `play_track`, and the swap arrives there and is dropped. The rebuild's own resume does not recover it: it re-dispatches `last_load` into the same dead ring.

Serialising is the wrong medicine for it. The fix is to stop discarding the swap — install the producer and end the current track, since the resampler is built against the output's sample rate and a producer swapped in under a running track would play at the wrong speed. That is a decoder change, not the shape of this PR, so it is #639.

**The fr/hi subtitle wording — declined, same reason as the previous round.** The string says "évite le rééchantillonnage quand il l'accepte" / "avoids resampling when the device accepts it"; it was already walked back from "supprime" to "évite" two rounds ago precisely because a speed other than 1× feeds rubato a deliberately false source rate whatever the device is opened at. My change this round touched only the trailing clause (the interruption now reads "au prix d'une coupure"), not the hedge. The speed caveat itself is in `docs/features/playback.md`, in those words: "It avoids the resampler rather than abolishing it". A toggle subtitle that already runs to two lines is not where that detail belongs.

**And one failure that was mine.** `Option::is_none_or` is stable since 1.82, the crate declares 1.80, and clippy's `incompatible-msrv` lint failed the Linux job on the ALSA guard I added last push. Windows could not catch it — the file is compiled only on Linux, so the lint never ran on it. Rewritten with `is_some_and`, stable since 1.70.

Validated this round on a Linux box rather than waiting on CI: `fmt --check` and `clippy -D warnings` clean, **567 app-crate tests green**. Windows and macOS clippy clean as before.

Issue #644, opened 2026-09-14
DSD: ratings and tag-saved lyrics still bypass the DSF writer


InstaZDLL:
#592 gave `.dsf` a tag writer, so the properties dialog can now correct a title, an artist or a year on a DSD file. Two other things that write into a tag did not come with it, because each reaches the file by its own route rather than through `patch_file`:

- **Star ratings.** `write_rating_to_file` (`commands/track.rs`) calls `lofty::read_from_path` directly to write the POPM frame. lofty has no `FileType` for DSD, so a rating on a `.dsf` stays in the database only. The next scan preserves it, since the `(mtime, size)` fast path never re-extracts an unchanged file — but the rating does not travel with the file.
- **Lyrics saved into the tag.** `write_lyrics_to_file` (`commands/lyrics.rs`) does the same, so choosing "save to tag" rather than a sidecar fails on a `.dsf` with a generic unknown-format error. The sidecar destination works.

Neither is a regression — both behaved this way before the DSF writer existed. What changed is that they are now the odd ones out: the same file whose title we can edit refuses its rating and its lyrics, with a message that describes lofty rather than the situation.

## What it needs

The container work is done: `waveflow_core::tagio::{read_dsf_layout, write_dsf_id3v2}` handle the layout, the metadata pointer and the header fields, and `commands/edit.rs::apply_patch_id3` is the worked example of applying an edit through the `id3` crate rather than lofty. What is missing is the same shape for two more payloads — POPM for the rating, USLT / SYLT for the lyrics — and a route to them for `.dsf` before each function reaches for lofty.

## Traps

- **Nothing may be lost.** The DSF write replaces the whole tag, so whatever the edit does not touch has to be carried through. That is what `id3::Tag` does by holding the frames it parsed, but only if the existing tag is read first rather than replaced by a fresh one — see the refusal in `patch_dsf` for the shape, and the commit that removed the `.ok()` which had quietly discarded unreadable tags.
- **Keep the version the file arrived with.** Rewriting a 2.3 tag as 2.4 is a file some hardware DSD players stop reading, and a 2.2 tag cannot carry `TDRC` at all — the writer refuses the whole tag rather than dropping the frame.
- **`.dff` stays refused**, per format, for the reason recorded in `UNTAGGABLE_EXTENSIONS`.

Found while answering a review comment on the #592 branch; kept out of it because each is a second applier over a different payload, not a change to the DSF writer.

InstaZDLL:
[...]

PR #645, opened 2026-09-14
feat(tags): write a tag into the padding, survive the interruption, and reach DSD

### LINKED ISSUE #590: feat: write tags in place when the padding block allows
Every tag edit rewrites the whole file. On a library sitting on a NAS, that is the difference between correcting an album and giving up on it.

## What we do today

`edit::patch_file` ends at lofty's `save_to_path`, which writes the file out in full. Correct, and safe — but it moves every byte even when only a few changed.

## What it should do

Most containers reserve space for exactly this. FLAC has a padding block; ID3v2 has a declared tag size with slack. When the new tag fits in the space already there, the bytes on disk can be patched where they sit, and the full rewrite becomes the fallback rather than the rule.

The gain is not marginal. A comparable player reports **2 613 ms → 20 ms** on a 34 MB FLAC across an SMB share — two orders of magnitude, and it lands hardest on exactly the libraries that hurt most today.

## Why this comes before the batch work

Every batch tag operation is this operation repeated. At 2.6 s per file, correcting a three-hundred-file folder is a ten-minute job with a progress bar; at 20 ms it is instant. This is the prerequisite that makes the rest of the tag tooling worth building, and it is also the "safe file writing" prerequisite already identified for online tag retrieval.

## Traps

- **Tag writes go through the concrete tag type**, never the generic one — the generic `Tag` drops non-standard Vorbis comments and TXXX frames ([invariant](https://github.com/InstaZDLL/WaveFlow/blob/main/docs/architecture/invariants.md#tag-writes-go-through-the-concrete-tag)). An in-place path must preserve that, and preserve untouched frames and blocks byte for byte.
- **Playback must be paused before rewriting the current track's file on Windows, and the hash re-taken afterwards** ([invariant](https://github.com/InstaZDLL/WaveFlow/blob/main/docs/architecture/invariants.md#file-write-safety-on-windows)). An in-place write does not escape this: it still holds a write handle.
- **`track.file_hash` is written by two different algorithms** — the scanner hashes head and tail, the tag editor hashes the whole file. Whichever path writes must keep its own convention rather than assume the column's meaning.
- **Decide what happens when the padding is exactly consumed.** Growing a tag into zero remaining slack must fall back rather than silently truncate, and the fallback needs the same atomicity guarantees the current path has.
- Worth measuring on a real share before and after, since the whole justification is a number.

### LINKED ISSUE #592: feat: write tags into DSD files (.dsf / .dff)
We index `.dsf` and `.dff` and we play them natively over DoP on three platforms, but we cannot correct a typo in one. `edit.rs` refuses them before touching the file:

> lofty 0.25's `FileType` carries no DSD variant, so `read_from_path` on a `.dsf` / `.dff` fails with a generic "unknown format" — accurate, but it reads as a corrupt file rather than as a format we never supported writing. DSD metadata is parsed by `waveflow_core::audio_format::dsd`, which is read-only, so there is no fallback to reach for: refusing before we touch the file is the whole of the honest answer.

That refusal was the right fix for the honesty problem. It is not an answer to the capability problem, and the gap is pointed: native DSD is the headline of our audio story, and it is the one format whose tags we cannot touch.

## What it needs

A writer of our own, because the dependency will not provide one. The reading side already exists — `waveflow_core::audio_format::dsd` parses both containers — so the container knowledge is in the tree; what is missing is its mirror image.

- **DSF** carries an ID3v2 tag at an offset declared in the header. Writing means rewriting that tag and updating the pointer and the file-size field with it.
- **DFF** has no ID3 by convention; metadata lives in its own chunk structure, and some taggers append an ID3 chunk. Decide which shape we write, and read both.

## Traps

- **Nothing may be lost.** The same rule the rest of our tag writing follows: frames and blocks we did not edit are copied through untouched, so correcting a title never costs a cover, a rating or a MusicBrainz identifier.
- **Cover art is addressed by content, not by index.** Removing "the third picture" is how the wrong picture gets removed.
- **Pause playback before writing the current track's file, and re-hash after** ([invariant](https://github.com/InstaZDLL/WaveFlow/blob/main/docs/architecture/invariants.md#file-write-safety-on-windows)).
- **The refusal must stay until the writer is complete per format.** Half-supporting DFF while claiming to support DSD would be worse than today's clear no.
- Worth pairing with in-place writing, since a DSD file is large and a full rewrite of one is expensive.

### LINKED ISSUE #598: feat: make file writes survive interruption, permissions and antivirus
Not to be confused with #590. That one makes tag writes **fast**; this one makes them **survive what can go wrong**. Both are prerequisites for batch tag work, for different reasons.

## What is missing

**No `sync_all()` before the rename.** The atomic write pattern is write-temp, then rename. Without an explicit flush before that rename, a power loss or a crash can leave the rename committed over a temp file whose contents never reached the platter — a file that is neither the old one nor the new one.

**Permissions are not carried across.** A rewritten file gets whatever the process default gives it, so a file that was read-only, or group-writable, or carried an ACL somebody set on purpose, quietly changes character on the way through.

**The Windows read-only attribute is a hard stop.** A file marked read-only cannot be replaced, and the operation fails with an error that describes the symptom rather than the cause. Lifting it for the write and putting it back is a few lines, and it is the difference between "cannot save" and saving.

**Antivirus holds files open, briefly.** On Windows, a freshly written file is frequently scanned before the process can rename over it, and the rename fails with a sharing violation that a retry a moment later would have avoided. One attempt turns a transient lock into a permanent failure the user sees as data loss.

## Why it comes before the batch work

Each of these is survivable once and unacceptable three hundred times. A single failed save is an annoyance somebody retries; the same failure rate applied to a folder-wide operation means every batch ends with a handful of files in an unknown state and no way to tell which.

It is also the stated prerequisite for online tag retrieval, which writes across a whole album at once.

## Traps

- **Pause playback before rewriting the current track's file on Windows, and re-hash blake3 into `track.file_hash` afterwards** ([invariant](https://github.com/InstaZDLL/WaveFlow/blob/main/docs/architecture/invariants.md#file-write-safety-on-windows)).
- **`track.file_hash` carries two different algorithms** — head-and-tail from the scanner, whole-file from the tag editor. Whichever path writes keeps its own convention.
- **A retry loop needs a ceiling and a distinction.** Retrying a sharing violation is right; retrying a permission error is a hang.
- **Cross-volume moves are copy, verify, then delete** — never delete-then-copy, and never a rename that silently degrades to a copy without the verify.

InstaZDLL:
Lot B of the v1.8.0 milestone: the three issues that all land on `edit::patch_file`. They are prerequisites for the batch tag work, for different reasons — one makes a tag write fast, one makes it survivable, one makes it possible at all on DSD.

Closes #598. Closes #590. Closes #592.

## What the code actually did, which is not what the issues assumed

Worth stating first, because it reshaped the work.

**There was no atomic write to fix.** #598 describes a write-temp-then-rename pattern missing its `sync_all`. lofty has neither: `save_to_path` opens the real file read-write and rewrites it where it lies. So "permissions are not carried across" was not true — writing in place keeps the inode and everything hanging off it — and the interruption failure was worse than described rather than milder.

**The ID3v2 path truncates the original.** lofty reads the whole file into memory, calls `truncate(0)` on the file it is rewriting, and writes it back. A crash, a full disk or a killed process anywhere in there left a truncated audio file, on every tag edit of an MP3.

**FLAC does not rewrite the file either — it moves it.** The writer builds the new block set and hands it to a `replace_range` that shifts every audio byte in 64 KB chunks whenever the new metadata differs in size from the old. Padding exists so that never has to happen, and lofty carries the open TODO for it (`lofty-rs#445`).

## What landed

**A tag edit that fits the padding touches only the padding (#590).** Both formats that matter: ID3v2 for MP3 and AAC, and FLAC. The tag is re-encoded, padded back out to exactly the length it replaces, and written over the old one. Not one audio byte moves and the file never changes length — which is what the tests assert, byte for byte, because the failure worth proving absent is a wrong-length prefix landing in the first audio frames. It is also the safe path, which was not the reason for building it: it never truncates and never shifts.

For FLAC, only the comment block is regenerated; the seek table, cue sheet, application and picture blocks are copied through untouched. That is faster and it is the only way to keep the rule that correcting a title never costs a cover.

**Two write modes, chosen explicitly (#598).** In place when it fits; otherwise a copy beside the original replaced by a single `rename`, with the original untouched until then. Around both: the Windows read-only attribute lifted and put back on every way out, a sharing violation waited out five times at 120 ms (and only that error class — retrying a permission error is a hang that ends in the same message), `sync_all` before reporting success, and a lock keyed on the file so the three commands that can write it never interleave. The trade the rename makes is stated where it is made: a new inode, so hard links keep the old content.

**A DSF can be tagged (#592).** Its metadata is a plain ID3v2 block *after* the audio at an offset the header declares, so the tag grows or shrinks without anything moving. The pointer is validated against the end of the `data` chunk — a header claiming its tag starts inside the audio is refused, because believing it would overwrite the track and then cut the file back. The file keeps the ID3 version it arrived with. `.dff` stays refused, and the refusal is per format now rather than for DSD as a whole.

## Scope left out, deliberately

- **`.dff`** — no ID3 by convention, metadata in its own chunk structure, and some taggers append an ID3 chunk anyway. Which shape to write is a real question with no obvious answer, and half an answer is worse than a clear no.
- **Ratings and tag-destination lyrics on DSD** — both reach the file through lofty directly rather than through `patch_file`, so neither sees the new writer. Not a regression, but now the odd ones out: #644.

## Validation

`cargo fmt --check`, `clippy -D warnings` at zero on Windows **and** Linux, **577 app-crate tests** and 278 core tests green on Linux. No frontend file is touched.

Twelve local review passes before this was pushed: **27 findings, 25 founded**, the last pass clean. Three of the founded ones were defects introduced by the fix for the previous one — version preservation made `TDRC` invisible to a 2.3 reader, the mirror that fixed that erased the year of every 2.3 tag it touched, and on 2.2 the same frame made the write fail outright. Two were declined with the reason written into the code, and one of those declines convicted the code it was defending: refusing to fall back onto a read-only file was right, and the rewrite path was bypassing that protection anyway.

## Summary by CodeRabbit

  - L’édition des métadonnées est désormais prise en charge pour les fichiers audio `.dsf`.
  - Les pochettes, dates et informations existantes sont mieux préservées lors des modifications.
  - Les dates ID3 sont enregistrées dans les champs appropriés selon la version des tags.

- **Améliorations**
  - Les écritures de tags sont plus fiables avec les fichiers verrouillés, protégés ou nécessitant une réécriture.
  - Les fichiers invalides ou tronqués sont détectés pour éviter leur corruption.
  - Les fichiers `.dff` restent refusés, avec une explication affichée dans la boîte de dialogue.


InstaZDLL:
@coderabbitai Two findings, one fixed in `d70255e1` and one declined with a reason.

**The DSF header ordering — founded, and it was a cheap fix for a real window.** You are right that the blocking case does not arise: a shrunk tag leaves the old `metadata_offset` below the new length, a removal makes it equal, and `read_dsf_layout` only rejects offsets *above* the length, after which `patch_dsf` reads EOF as `NoTag`. The window you actually named is the one that matters: `set_len` ran before the two header fields, so an interruption in between left `total_size` larger than the file physically was — and an oversized declared length is what makes a player call a file damaged.

An in-place update of three separate places cannot be made atomic, so the question is which side of the truth a stop leaves it on. Reordered to bytes, then header, then truncation. Every window now leaves the header *under*-declaring, and a reader that trusts the declared size simply ignores whatever sits past it. On a removal the tag is gone from the header before the bytes are cut, rather than the other way round. There is a test for the property rather than for the sequence.

The doc comment above it claimed the old order was the safe one — "the two header fields are written after the tag, so a file interrupted mid-write still declares the tag it had". True of the pointer, and beside the point for the size.

**Extracting the shared field semantics into `crates/core/` — declined here, on the trade rather than the merit.** The duplication is real and you have identified it precisely: `apply_patch` and the `TagPatch::Fields` branch of `apply_patch_id3` encode the same four decisions. What is already in place is the guard against them drifting — `patch_agreement_tests` runs the same `TrackEdit` through both appliers and compares the observable fields, including the clearing cases and the rule that a title-only save must not truncate a full release date to its year. A divergence fails the build rather than reaching a user.

Against that, a third abstraction spanning two crates is a refactor with its own review surface, and this branch is eighteen commits deep on a milestone that is full. Your own note that `TagPatch::Cover` cannot come along — it depends on `lofty::picture::MimeType` — is a good sign the extraction is less tidy than it first looks: it would split one enum's handling across two layers. Better as its own change, decided on its own merits, than appended here.

For the record on process: this branch went through twelve local review passes before it was pushed — 27 findings, 25 founded. Three of them were defects introduced by the fix for the previous one, which is the argument for the passes rather than against them.

PR #646, opened 2026-09-14
feat(playback): album gain, album shuffle, and Opus

### LINKED ISSUE #581: feat: support Opus playback
Requested in [discussion #519](https://github.com/InstaZDLL/WaveFlow/discussions/519) (first of its three asks — the `.lrc` sidecar detection it also asks for already works, and taskbar buttons are a separate matter).

## Current state

Opus is not playable. The scanner's extension list refuses it, so `.opus` files are not even indexed.

What already works, and is worth not rebuilding:

- **`symphonia-format-ogg` 0.6.1 ships a complete Opus mapper** — it parses `OpusHead`, wires the pre-skip through `with_delay()`, computes multi-frame durations, and maps `OpusTags` to Vorbis comments, so **tags already read correctly**. Only the decoder is missing: `symphonia-codec-opus` does not exist (upstream PR #398 has been WIP since October 2025).
- `isomp4` is already enabled, so Opus-in-`.m4a` would come along for free.

## Decoder choice: the C binding

**`opusic-sys` 0.7.5.** BSD-3, actively maintained, vendors **libopus 1.6.1**. Links dynamically against the system library by default — nothing to install on Linux, no cmake — with a `bundled` feature for a static cmake build on Windows and macOS. It exposes raw FFI only, so roughly 100 lines of safe wrapper are ours to write; no safe `opusic` crate exists.

**Not `opus` 0.3.1.** It depends on `audiopus_sys` **0.2.2, published in 2021**, which vendors libopus ~1.3.x. Harmless on Linux (system library), but Windows and macOS builds would ship a 2019 codec.

The C dependency is cheaper than it sounds: **`libopus.so.0` is already loaded in the WaveFlow process**, pulled in transitively — `libwebkit2gtk-4.1.so.0` links it directly. Flatpak gets it from `org.gnome.Platform`.

## Why not a pure-Rust decoder

Two candidates were measured rather than assessed from their READMEs. Both were rejected, for opposite reasons.

**`libopus-rs` 0.0.2** covers CELT-only fullband, single frame: 601/601 packets at 128k and 96k stereo, **9/601** at 64k stereo, **0/601** on voice. 87.7 dB inside its range, and it refuses honestly (`Error::Unimplemented`) outside it. Too narrow to ship, but it fails loudly.

**`opus-rs` 0.1.33** (re-measured 2026-09-07, on the release published that morning) decodes everything and **never returns an error**, including on the files it gets wrong. Reference decode is libopus, whose 16-bit output caps the comparison at ~79.5 dB:

| Regime | SNR vs libopus | Reading |
| --- | --- | --- |
| CELT | **79.5 dB** | at the reference ceiling → effectively exact |
| SILK wideband | **bit-identical** | 0 differing samples out of 575 997, at a constant 3-sample offset |
| SILK stereo | 63 dB | correct in practice |
| SILK narrowband | 11 dB | wrong |
| SILK mediumband | **−5.4 dB** | error louder than the signal, and no lag improves it |
| Hybrid | 30.6 dB at best | wrong |

Two things follow.

The trajectory is real — SILK wideband went from broken to bit-exact in two and a half weeks, and the crate is worth re-measuring when commits land on mediumband, narrowband and hybrid.

But it cannot ship today, and the reason is a case we would actually meet: a **48 kb/s stereo** file that is 94 % CELT and 6 % hybrid scores only **30 dB**. Six percent of packets being wrong costs about 50 dB, and 48 kb/s stereo is a plausible encode for someone fitting a large library on a small disk. Combined with `Ok` being returned on every packet, that is silent audio corruption with nothing for the player to catch.

Writing our own is not on the table. For scale: the DSD pipeline cost 1 688 lines without a single bit of entropy decoding — container, low-pass, decimation, no binary conformance required. Opus is a range coder plus CELT plus SILK plus hybrid plus mode switching, 15–20k lines in which every bug is a corrupted playback.

## Integration: three traps, all verified

1. **Pre-skip must be consumed.** Symphonia 0.6 removed `enable_gapless`; the delay is exposed as `Track::delay` and it is the application's job to drop it. We currently consume **neither `delay` nor `padding`** — `crossfade.rs` and `decoder.rs` have no reference to either. This is mandatory for Opus (312 frames) and would improve MP3/AAC gapless as a side effect.
2. **The `OpusHead` R128 gain** is not applied by the mapper. Either wire it into ReplayGain or ignore it deliberately, but decide rather than inherit.
3. **Seeking needs ~80 ms of pre-roll** — decode and discard — or the first frames after a jump are audibly wrong.

## Where the change lands

Three sites move from `symphonia::default::get_codecs()` to a custom registry:

- `app/src/audio/crossfade.rs` — `ActiveStream`, the single entry point for playback, files **and** web radio
- `core/src/analysis.rs`
- `core/src/scanner/extract.rs` — which also has to start accepting the extension

## Measurement method, for whoever re-runs it

Encode with `ffmpeg -c:a libopus -b:a <rate>`, parse the Ogg pages by hand (~30 lines), and classify each packet from its TOC byte: `config = toc >> 3`, where 0–11 is SILK, 12–15 hybrid, 16–31 CELT. Decode with the candidate, drop `pre_skip`, and compare against a reference decode in SNR.

Three things that are easy to get wrong:

- **The reference must be libopus**, i.e. `ffmpeg -c:a libopus -i file.opus -f f32le`. Without that flag ffmpeg uses its own native Opus decoder, which is itself **2.6 dB** away from libopus on SILK content while agreeing to 79.5 dB on CELT. An earlier round of this measurement used the native decoder and partly measured ffmpeg rather than the crate.
- **Reaching pure SILK takes a band-limited source.** A broadband signal makes libopus pick CELT even under `-application voip`. A 3.8 kHz low-passed source at 8–20 kb/s mono with `-cutoff 4000/6000/8000` gives 100 % SILK packets.
- **Find the offset by FFT cross-correlation**, not a naive sweep — ±2000 lags over 576k samples does not finish. And keep an internal control: if CELT does not reach the reference ceiling, the harness is wrong, not the crate.

### LINKED ISSUE #587: feat: ReplayGain album mode
Deferred to a second PR when #545 was arbitrated, on the grounds that it touches the playback context and the queue. Having looked at where things actually sit, that turns out to be much cheaper than it sounded.

## What it is for

Track gain normalises every track to the same loudness. On a record mastered as a whole — a hushed intro, a loud single, an ambient closer — that flattens what the mastering engineer intended: the quiet interlude gets pushed up to meet the single. Album gain applies **one gain across the whole record**, preserving the level relationships inside it, at the cost of two different albums no longer matching each other.

Which of the two is right depends on **why** a track is playing: album gain when the record is being played through, track gain when the same song comes up shuffled in a playlist between two unrelated things.

## The context is already there

This is the part that made it look expensive, and it is already solved. `ActiveStream` — the per-stream decoder state — carries these fields next to each other:

```rust
pub source_type: String,
pub source_id: Option<i64>,
/// What is known about this track's loudness — from its own tags
/// when it has them, from our analysis otherwise. Kept as metadata
/// rather than as a baked-in scalar so the pre-amp and clipping
/// settings apply the moment they change […]
pub replay_gain: super::replay_gain::TrackGain,
```

`queue_item.source_type` is a `CHECK`-constrained column over `'album'`, `'playlist'`, `'artist'`, `'library'`, `'liked'`, `'manual'`, `'radio'`, and it already travels all the way to the decoder. So "is this track playing as part of its album" is a field that has already arrived where the gain is applied — no queue change, no new plumbing.

And because the gain is deliberately kept as metadata rather than a baked scalar, a mode switch would take effect mid-track exactly as the pre-amp already does.

## What is actually missing

1. **The data does not reach the stream.** `rg_album_gain_db` and `rg_album_peak` are read from tags and stored on `track` by the scanner, but nothing in `audio/` mentions them: `TrackGain` carries one gain and one peak. It needs both pairs, and the loader needs to select the album columns it currently ignores.
2. **A mode**, persisted per profile like the other ReplayGain knobs: track, album, or automatic — where automatic means album gain when `source_type == "album"` and track gain otherwise. Whether automatic should be the default, or even whether the manual modes are worth offering at all given the context is known, is worth deciding rather than copying from elsewhere.
3. **The clipping cap must switch too, and this is the trap.** Clipping prevention caps the gain at `-20·log10(peak)`. In album mode it has to use the **album** peak; capping each track by its own peak would pull individual tracks down according to their own loudest sample, which reintroduces exactly the per-track variation album mode exists to remove.
4. **Strings for the mode, propagated to all 17 locales.**

## Two questions worth settling first

**Tagged-only, or measured too?** Our analysis pass is per track, so we have no album-level measurement of our own. Album mode would therefore work only on files carrying `REPLAYGAIN_ALBUM_*` tags, and fall back to track gain elsewhere — which is defensible, and is what most taggers produce anyway. Computing an album gain ourselves means analysing a whole album together and deciding what an "album" is for loose files, which is a much larger piece of work and probably a separate one.

**Freshness.** Where an album peak is measured rather than tagged, it inherits the problem in #586: analysis rows written before the BS.1770 switch carry a peak taken over a mono downmix, which under-reports it and makes the clipping cap too generous. Album mode leaning on peaks makes that more visible, not less.

### LINKED ISSUE #618: feat: album-based playback, with shuffle by album and an album mode for mood radio and daily mix
Requested by email:

> Make the album more central to the listening experience. I'd really like an option to shuffle by album, rather than only shuffling individual tracks.
>
> It would also be great if Mood Radio and Daily Mix had an option to choose between album-based and track-based playback.
>
> More generally, I think giving albums a bit more importance throughout the app would make the experience feel more like a proper music library rather than just a collection of individual tracks.

## Today

Everything is track by track:

- **Shuffle** is a boolean (`player.shuffle`). `queue::shuffle` reorders the queue's tracks individually, keeping the current track in position 0.
- **Mood Radio** returns 40 individual tracks.
- **Daily Mix** shuffles its tracks and truncates at 50.

The data needed is already there: `track` carries `album_id`, `disc_number` and `track_number`.

## Proposal

1. **Shuffle by album.** Shuffle becomes a three-way mode: off, tracks, albums. In album mode, whole albums are shuffled and each album plays in disc/track order.
2. **Album mode for Mood Radio and Daily Mix.** Pick albums that fit rather than scattered tracks, then play each one in order.

## Constraints to design around

- **The current track is pinned at position 0** when shuffling. In album mode, the rest of its album has to follow it.
- **`player_play_tracks` reshuffles the whole queue** when shuffle is on, which would scramble an album-ordered Mood Radio or Daily Mix. Their album mode must not go through that, or must set the mode explicitly.
- **Daily Mix's 50-track cap** would cut an album in the middle. An album mode needs a different limit (whole albums, or a duration).
- **MPD exposes shuffle as a boolean** (`random`). An album mode needs a mapping there.
- **Mood Radio's selection is per track** (#616). Deciding whether an *album* fits a mood needs an album-level measure, such as the median BPM of its analysed tracks.

The broader "albums first" direction is a UX question beyond these two options. Worth a discussion before any design.

InstaZDLL:
Lot C of the v1.8.0 milestone: the three issues about how a track gets played rather than how it gets written. They turned out to share a subject — the album as a unit rather than a bag of tracks — which is why two of them reinforce each other rather than just sharing a branch.

Closes #587. Closes #618. Closes #581.

## Album gain, chosen from context rather than from a setting (#587)

ReplayGain levels every track against every other one. That is right when a song comes up shuffled between two unrelated things, and wrong on a record mastered as a whole: the hushed interlude gets pushed up to meet the single, flattening exactly what the mastering engineer put there.

The feature looked expensive because it sounds like it touches the queue. It does not. `queue_item.source_type` already travels all the way to the decoder, sitting on `ActiveStream` right next to the gain — so "is this track playing as part of its album" was a field that had already arrived where the answer is needed. The default is therefore **automatic**: album gain while a record plays through, track gain otherwise, with nothing for the listener to manage. Album shuffle counts as a record playing through, which is the first place the two issues meet. `track` and `album` can still be forced.

**The clipping cap switches with the gain**, which is the part that is easy to get wrong and the one the issue flagged. In album mode it caps by the **album** peak — the same number for every track on the record, so the cap is uniform. Capping each track by its own peak would pull tracks down according to their own loudest sample, re-introducing precisely the per-track variation album mode exists to remove.

Album numbers are tag-only, because our analysis pass measures one track at a time and has no notion of a record. A file without them falls back to its track gain, since album mode would otherwise do nothing at all on a half-tagged library. The peak falls back with it, and that one is a deliberate trade in the other direction: an uneven cap is cosmetic, and the cap only ever binds where the alternative is audible clipping.

## Shuffle whole records, and build sessions out of them (#618)

Shuffle becomes three-way: off, tracks, albums. In album mode only the order of the records is randomised — inside each one the tracks are put back into disc and track order, so a record that arrived in the queue scrambled still plays the way it was pressed. The record you are in carries on rather than restarting: the current track keeps position 0, the rest of its album follows, and the tracks before it come round at the end. A track with no album is its own record, so loose files still shuffle like tracks instead of being welded into one block that always plays together.

Persisted as the existing on/off switch crossed with a grouping, rather than as one three-valued key. `player.shuffle` stays authoritative for "is shuffle on" — it is what MPD's `random` maps onto and what an older build reads — the grouping is remembered while shuffle is off so a preference for whole records survives an off/on, and a profile that predates this needs no migration. MPD's `random 1` therefore turns shuffle on **without** forcing tracks: a remote that cannot express the grouping should not quietly undo it.

Mood Radio and the Daily Mix get the same treatment behind one setting, because it is one preference. A record fits a mood when **most of what we have measured of it** sits inside the tempo window — a fraction rather than the median the issue suggested, since a median says nothing about spread: a record half ambient and half thrash lands in the middle and would be offered for a mood neither of its halves belongs to. The cap is then spent in whole albums; truncating at 40 or 50 tracks would end the session halfway through a record, which is the single thing this exists to prevent.

Album mode stays a preference rather than a contract: when no whole record fits a mood, the radio falls through to picking tracks rather than refusing to play. A library whose records are mostly unanalysed can satisfy a mood track by track while no album qualifies, and a "play something now" button should not answer with an error.

**Two things found in the queue on the way past.** Reordering wrote every row back as `source_type = 'manual'`, which threw away the source a `play_event` is attributed to *and* the boundary `fill_queue` uses to tell queued-up "play next" items from the source queue around them — so shuffling an album silently cost both, and unshuffling did not bring them back. Sources now survive a reorder, one per occurrence, so a queue holding the same track twice hands each copy back its own. That same column is what lets automatic ReplayGain tell a record playing through from a shuffled playlist, which is the second place the two issues meet.

## Opus (#581)

Opus was not playable, and the scanner refused the extension so the files were not even indexed — while the desktop entry and the AppStream metadata have been claiming `audio/opus` all along.

symphonia was never the problem for the container: its Ogg reader ships a complete Opus mapper, so tags and durations already came out right. The missing piece was the decoder, and `symphonia-codec-opus` does not exist. This one is **libopus 1.6.1** through `opusic-sys`.

**Why a C binding.** Two pure-Rust decoders were measured against a libopus reference rather than assessed from their READMEs, and both were rejected for opposite reasons. `libopus-rs` covers CELT only and refuses honestly outside that range — too narrow to ship, but it fails loudly. `opus-rs` decodes everything and **never returns an error**, including on the files it gets wrong: a 48 kb/s stereo file that is 94 % CELT and 6 % hybrid scores 30 dB. That is silent audio corruption with nothing for the player to catch. Writing our own is not on the table — for scale, the whole DSD pipeline cost 1 688 lines without a single bit of entropy decoding.

**Static on every platform**, which is a packaging decision rather than a technical one: the Linux packages repackage the release binary rather than building from source, so a system libopus would mean a runtime dependency added to three packaging manifests and a library bundled into the AppImage, to gain nothing. The cost is `cmake` at build time — see the validation note below, because that turned out not to be free.

**The pre-skip is the part that was nearly missed.** Every Opus stream opens with encoder priming that must never be played. The obvious reading is that the reader handles it — it sets `Track::delay` from `OpusHead` and hands each packet a `trim_start`, which is exactly how the Vorbis decoder next door disposes of its own priming. That reading is wrong: the Opus packet parser in `symphonia-format-ogg` 0.6.1 reports a discard of **zero** for every packet, so the trim is always empty. Measured against a real file rather than reasoned about — a 3-second encode decoded to 144 312 frames instead of 144 000, the difference being precisely the 312-frame priming. The decoder now counts it, credits whatever the reader did report against what is owed so a future upstream fix cannot make it drop twice, and does **not** re-apply it after a seek, where there is nothing to drop and doing so would eat real audio.

RFC 7845 also suggests decoding ~80 ms before a seek point and discarding it so the decoder has converged. We deliberately do not, and the reason is written where the decision is: that is audio the listener asked for, and the convergence artefact is brief and bounded where the loss would not be.

**One codec registry for the whole app.** Playback, the analysis pass and the scanner's probe each reached for `symphonia::default::get_codecs()`. That was harmless only while the app shipped no decoder of its own; now it is not. The three have to agree on what this build can decode, because that disagreement is exactly how an unplayable track gets into a library — the single thing `AUDIO_EXTENSIONS` exists to prevent. Written up as an invariant.

## Also here

**A sidecar the user placed beats a description we guessed at.** Reported on [discussion #519](https://github.com/InstaZDLL/WaveFlow/discussions/519) by the same person who asked for Opus, and not previously tracked. A `.m4a` pulled with `yt-dlp` carries the auto-generated "Provided to YouTube by…" credit in its `description` tag; that blurb is comfortably more than the three lines which were the whole test for treating a description as lyrics, so it won — and because the embedded tier ran before the sidecar, the `.lrc` placed next to the file was never read at all. Other players showed the lyrics; we showed the credit.

Two things were wrong at once. A description field is not a lyrics field, and reading it as one is a guess; and the guess outranked an explicit statement, which is what made it invisible. The tier survives, because some rips really do put lyrics there, but it moves behind the sidecar and refuses that one blurb — by a recogniser that stays deliberately narrow and only looks near the top of the text, because a false positive here costs someone their real lyrics.

The wrong text was also **cached**, and the waterfall never refetches once a row exists — so correcting the reader alone would have reached only tracks nobody had opened yet, which does not include the person who reported it. A cached `embedded` row that the same recogniser identifies as a service credit is dropped on read and re-resolved. Narrow, because the row is provably not lyrics, and governed by one predicate so what we refuse to store and what we refuse to serve cannot drift apart.

## Validation

`cargo fmt --check` and `clippy -D warnings` at zero across the workspace on all three platforms, and the full suite green on the two that can run it: **900 tests on macOS 26.6** and **912 on Fedora 44**. Windows runs the core crate only — the app crate's tests do not execute there (`STATUS_ENTRYPOINT_NOT_FOUND` from the Tauri DLLs), which is exactly how two of the new ReplayGain tests went from written to green-by-assumption; both platforms caught them, with the same two failures, and the defect was in the expectation rather than the code.

macOS matters more than usual here. No CI job compiles it, and this PR adds a cmake-built C dependency to every platform, so it is the first time libopus has been built there at all.

Nine local review passes before the first push to this repository: **31 founded findings fixed, 12 declined with the reason recorded**. Three of those declines were mine to correct — a search window that excluded the binding it was looking for, a consistency argument that had the ages of two conventions backwards, and a first reading of symphonia's pre-skip handling that the measurement then contradicted. The trend across passes was 8 → 6 → 4 → 2 → 1 → 2 → 4 → 2 → 0; the bump at seven is the interesting one, and it is the same shape three times over: a fix moved an invariant and what depended on it did not announce itself.

Opus was checked against real files rather than only against unit tests — encoded with `ffmpeg -c:a libopus`, probed and decoded through the app's own registry, frame counts compared against the expected duration. That is what turned up the pre-skip.

**cmake is a build dependency now**, and it is declared rather than assumed. The five Linux workflows install it; the macOS and Windows release jobs check for it and install it if the runner image ever stops shipping it; CONTRIBUTING lists it for all three platforms. That last part is not theoretical: a current Mac with Xcode installed turned out to have neither cmake nor Homebrew, so a contributor on a clean macOS cannot build this without installing one. The one place it is still taken on trust is the Flatpak SDK — **no CI job builds that manifest** — and the manifest now says so at the line that assumes it.

Two things found on the way, recorded in the commits for whoever hits them next. Running `generate-sources.sh` **on Windows** corrupts the Flatpak output in a way CI does not catch: the upstream generator builds paths with `os.path.join`, so git-sourced entries come out with backslash separators inside shell commands that run under Linux in the Flathub sandbox, and `check-sources.py` only verifies that every crate is covered, not that the paths are shaped like Linux paths. And MSVC fails the libopus build with `C1083: Cannot open compiler generated file` when the build directory path is long — a checkout nested deep under a temp directory is enough to trigger it.

## Summary by CodeRabbit

  - Ajout de la lecture audio au format Opus.
  - Le mode aléatoire propose trois options : désactivé, par morceaux ou par albums.
  - Mood Radio et Daily Mix peuvent générer des sessions à partir d’albums complets.
  - ReplayGain prend en charge les modes automatique, par morceau et par album.

- **Améliorations**
  - Récupération des paroles enrichie grâce aux descriptions locales, avec filtrage des crédits YouTube.
  - Meilleure prise en charge des métadonnées lors du réordonnancement de la file d’attente.

- **Documentation**
  - Documentation mise à jour sur Opus, ReplayGain, le shuffle et les formats audio pris en charge.


InstaZDLL:
@coderabbitai Seven findings: five fixed in `c3fb3f8f`, two deferred to #647 with the reasoning written down rather than waved at.

**The prefetch losing the description tier — founded, and it was a regression, not a gap.** You are right that `run_prefetch` never reaches `read_description_lyrics`. What makes it worth more than the fix: that tier used to be *inside* `read_embedded_lyrics`, which prefetch does call. Moving it out was my change; I updated `try_local_lyrics` and never looked for a second caller of the function whose contract I had just narrowed. So a description-only track silently stopped being filled locally and started going to the network. Both paths now go through one helper that owns the order, so they cannot drift on it again.

**`MIN(t.primary_artist)` — founded, and worse than a decode error.** `track.primary_artist` is `INTEGER REFERENCES artist(id) ON DELETE SET NULL`, so an album whose tracks have all lost theirs aggregates to NULL; decoding that into `i64` fails the row, which fails the query, which takes the whole radio down. It is `Option<i64>` now, and records with no artist share one bucket for the per-artist cap — the conservative reading, and it keeps a pile of untagged rips from filling the radio between them.

**The Daily Mix fallback — founded on consistency, and the inconsistency was mine.** Mood Radio gained exactly this fallback on an earlier review pass, with the argument that album mode is a preference rather than a contract. Leaving the Daily Mix to clear its slot meant the same setting behaved two ways, and the comment explaining why the Daily Mix did *not* fall back was contradicting its sibling three files away. Both fall back now.

**Both documentation lines — founded.** The cache repairs itself since a later pass, so the note telling users to refetch was describing the version before that; and LRCLIB became tier 6 when the description tier was numbered in, which I fixed in two places and missed in two more.

**The album-session ReplayGain context — deferred to #647, not declined.** You have identified something real, and it is the one place where the two features in this PR fail to meet: an album-mode Mood Radio or Daily Mix plays whole records in disc order and is enqueued as `'radio'` or `'playlist'`, so `listening_to` reads it as a track session.

The reason it is not here is that the signal does not exist. `listening_to` has two inputs — the queue row's `source_type` and the shuffle grouping mirrored into `SharedPlayback` — and a generator session is neither. Mood Radio's command returns a bare `Vec<i64>`, so the frontend cannot even tell whether the backend answered in albums or fell back to tracks, which it deliberately does. The Daily Mix is harder still: the session outlives its generation, so the flag has to be persisted in `smart_rules` and read back when the playlist is played weeks later. And `queue_item.source_type` is `CHECK`-constrained with each value meaning something to `play_event`, so an album-mode radio cannot simply claim to be `'album'` without lying to analytics. That is a new end-to-end signal plus a rule-schema change — coherent work, but not a fix to either feature on its own. #647 carries the shape and the traps, including the one that matters most: clearing the flag is as important as setting it, or the next ordinary playlist inherits album gain.

For the record on process: this branch went through **nine local review passes before it was pushed** — 31 findings founded, 12 declined. Three of those declines were mine to correct on a later pass, which is the honest argument for running the passes rather than for trusting the first answer.

InstaZDLL:
@coderabbitai Same subject as the earlier finding, and it stays in #647 rather than landing here. The reasoning is worth restating with what checking it again turned up, because you are right about the behaviour.

**Confirmed, again.** `MoodRadioGrid` calls `playerPlayTracks("radio", null, ids, 0)` whatever the backend answered with, so an album-mode radio queues as `'radio'`, `listening_to` sees neither `'album'` nor album grouping on shuffle, and the decoder gets `Listening::ToATrack`. Whole records, in disc order, on track gain.

**Two reasons it is not a narrow fix, and the second is new since last time.**

The clearing is harder than the setting. A per-queue "this session is album-ordered" flag has to be written on *every* path that replaces the queue, or a stale one hands album gain to the next unrelated playlist — a silent audio defect, which is worse than the one being fixed. I had assumed `player_play_tracks` was that single path. It is not: `queue::insert_after_current` also fills the queue outright when it finds it empty, so "Play next" on an empty queue is a second entry point that never passes through the command carrying the flag. Any fix has to cover both, and MPD's paths besides.

And doing Mood Radio alone would re-create exactly the inconsistency the previous review round asked me to remove. Those two generators are driven by one setting, and this same PR just made their empty-result behaviour match after you pointed out they disagreed. Giving album context to the radio and not to the Daily Mix would put them straight back to answering the same preference two different ways — and the Daily Mix cannot be fixed the same way, because its session outlives its generation: the mix is a stored playlist, played weeks later through the ordinary playlist path, so the flag has to be persisted in `smart_rules` and read back at play time.

So it is one change covering both, spanning the command layer, the queue, the decoder and the rule schema — not two independent ones. #647 carries the shape, the traps, and now the `insert_after_current` case as well.

What ships here is honest about itself: album mode picks and orders whole records, and ReplayGain treats the session the way it treats any radio. Nothing claims otherwise, and the follow-up says exactly what is missing.

Issue #595, opened 2026-09-09
feat: fall back to software rendering after a GPU start-up crash


InstaZDLL:
When GPU-accelerated rendering fails on Linux, the app starts and shows nothing. The window exists, the process is alive, and the user has a blank rectangle with no message and no way back. The AppImage `libwayland-client` fix (#525) removed one cause; it did not make the class of failure survivable.

## Shape

A sentinel written before the window is created and cleared once the interface has actually painted. A launch that finds the sentinel still there knows the previous one never got that far, and starts in software rendering instead — with an explanation, because a silent downgrade is its own kind of confusing.

Plus an escape hatch: an environment variable to force `gpu` / `software` / `auto`, so a user who has been told what to try can try it without editing anything.

## Why this is cheaper than it looks

**The signal already exists.** `lib.rs` emits `app://ready` when the frontend has painted — the splash handoff already depends on it. That is exactly the "we got far enough" event the sentinel needs to clear on, so no new instrumentation is required.

## The trap to write down before starting

**A WebKit process crash must block the disarming, not just a failure to start.** If the sentinel is cleared on window close as well as on paint, then the user closing the blank window clears the evidence, and the next launch tries the GPU path again — the loop repeats forever and the mechanism appears not to work. The sentinel is cleared by *paint* and by nothing else.

Worth considering alongside: a known-bad environment check, so a platform we already know cannot do GPU rendering starts in software on the first run rather than the second.

PR #648, opened 2026-09-14
feat(library): columns, an inventory, a task bar, high contrast and a cache location

### LINKED ISSUE #588: feat: choose, reorder and resize the columns in track lists
Track lists ship one fixed set of columns. Someone whose library is organised around a field we don't show — a composer, a catalogue number, a rip source kept in a custom tag — has no way to see it, and no way to sort by it.

## Shape

**Choose which columns show, and in what order.** Two sources feed the list: the fields the library already holds, and **the tags actually present in the user's files**, offered with a count of how many tracks carry each. Offering the theoretical tag list instead would bury the five useful ones under thirty-five that nobody has.

**Sort by any column, both directions.** With one rule that matters: sorting has to happen **in the database** wherever the list is paginated. Sorting the loaded page of a fourteen-thousand-track library in memory produces a wrong answer that looks right.

**Resize, and remember.** A drag handle on the right edge of each header, living outside the sort button — inside it, a press to widen would fire the sort. Double-click to fit the content. Widths persist per column, with a way to put them all back.

Two details worth stealing from anyone who has done this:

- **Measure the fit off-DOM.** Rows are virtualised, so the ones outside the viewport have no computed layout; measuring the DOM would size the column to whatever happens to be on screen. Measuring the text through a canvas gives the same answer regardless of scroll position — and the header label has to be in that measurement, or a column fitted to short content shows a truncated title.
- **Give the title column a floor.** If it is the only flexible column, it absorbs every pixel the others take and disappears once a few columns are added.

**Keep the header visible while scrolling.** A value with no label above it means nothing, and that goes double for a column the user chose themselves out of their own file tags.

## Our side

The tag reading already exists — `edit.rs` goes through the concrete tag type per format precisely so non-standard frames survive, so the values are reachable. What is missing is the plumbing to the list and the per-profile persistence of the choice.

Persistence goes through [`useProfileSetting`](https://github.com/InstaZDLL/WaveFlow/blob/main/src/hooks/useProfileSetting.ts) rather than a hand-rolled hook, per the settings invariant.

## Traps

- **Virtual scroll everywhere.** The table is `@tanstack/react-virtual` over `usePageScroll()` with no nested `overflow-y-auto` ([invariant](https://github.com/InstaZDLL/WaveFlow/blob/main/docs/architecture/invariants.md#virtual-scroll-everywhere)). A sticky header and resizable columns have to live inside that constraint, not around it.
- **Existing settings must survive.** Whatever shape the stored preference takes, a profile written before this change has to keep showing the columns it shows today rather than losing all of them at once.
- New strings go to all 17 locales.

### LINKED ISSUE #589: feat: a "needs attention" inventory of the library
There is no way to ask the library what is wrong with it. Someone who has just imported two thousand files has no starting point other than scrolling and noticing.

## Shape

An inventory that opens on **what needs attention**, as counted categories you can click into:

**Missing fields** — no title, no artist, no album, no year, no track number, no cover. Plus one that is not about the tags: **files in a format we cannot write tags into**, which today is `.dsf` and `.dff` ([`UNTAGGABLE_EXTENSIONS`](https://github.com/InstaZDLL/WaveFlow/blob/main/src-tauri/crates/app/src/commands/edit.rs)). Someone deciding how to fix their library deserves to know which files they cannot fix in place.

**Album inconsistencies** — years that disagree inside one album, a compilation with no album artist, duplicate track numbers, gaps in the numbering. These are the ones a per-track view can never surface, because each track looks fine on its own.

**Probable duplicates** — same title and artist, within a couple of seconds of each other. **Chain them transitively rather than comparing pairs**: 180 s, 181.5 s and 183 s are one recording, but the outer pair is 3 s apart and a naive pairwise tolerance would split them into two groups. This is the detail that decides whether the feature is useful or merely present.

Clicking a category loads exactly those tracks into a working view, so the inventory is an entry point and not a report.

## Our side

Everything here is an aggregate over columns we already have. `track` carries title, year, track number, album and artist links; duration is stored; `album` carries `is_compilation`; artwork presence is derivable. No new data, no migration — this is queries plus a view.

Two existing pieces to reuse rather than re-derive: the duplicates work already in the tree, and [`waveflow_core::metadata::name_match`](https://github.com/InstaZDLL/WaveFlow/blob/main/src-tauri/crates/core/src/metadata) for name comparison, which handles NFD combining marks properly.

## Traps

- **Multi-artist credits are `"; "` joined and rebuilt from `track_artist`** ([invariant](https://github.com/InstaZDLL/WaveFlow/blob/main/docs/architecture/invariants.md#multi-artist-queries)). A duplicate check that compares the display string will disagree with one that compares the credit list.
- **Album identity is `(canonical_title, album_artist_id)`** ([invariant](https://github.com/InstaZDLL/WaveFlow/blob/main/docs/architecture/invariants.md#album-grouping--canonical_title-album_artist_id)), so "same album" for the consistency checks means that pair, not the raw title.
- The counts run over the whole library; they belong on the read side and must not block behind a scan.
- New strings go to all 17 locales.

### LINKED ISSUE #596: feat: high-contrast mode
Our secondary text is grey on grey by design, and for someone with low vision that design is the problem. There is no way to ask the app to be legible instead of elegant.

## Shape

A mode that lifts contrast across the whole interface at once, rather than a pass over every component:

- **Secondary greys resolve to real contrast.** Because colour lives in theme tokens, this is a token set rather than a component rewrite — which is the reason it is worth doing as a mode and not as a hundred small fixes.
- **Background blurs off.** `backdrop-filter` behind text is the single biggest legibility cost in the current design, and it buys nothing to someone who needs the contrast.
- **Keyboard focus becomes unmistakable.** Not a subtle ring — a visible, high-contrast outline that survives on every surface.
- **Controls that rely on subtlety get redrawn.** Checkboxes and toggles that read as "filled versus not filled" at normal contrast need a real mark, not a shade change.

## Our side

The five skins and fourteen themes already run entirely on tokens, and the artifact-level rule in `app.css` is that colour is defined once at the root and consumed everywhere else. That is what makes a contrast mode a palette rather than a fork — and it also means the mode has to compose with skin and theme selection rather than replace it.

## Traps

- **Every skin must survive it.** A mode that only works on one of the five is a mode that breaks four.
- **`prefers-contrast: more` exists**, and someone who has set it system-wide should not have to find our toggle. Decide whether the OS preference selects the mode by default.
- Do not let the mode become an excuse to skip contrast in the normal themes — this is for people who need more, not a place to put the contrast the default palette should already have.
- New strings go to all 17 locales.

### LINKED ISSUE #601: feat: a status bar for long-running tasks, with cancellation
Long operations report themselves inconsistently and none of them can be stopped.

A library scan emits `scan:progress` and the Library view shows it. An analysis sweep, an artwork fetch, a lyrics prefetch, a remote mirror walk, a backup — each reports somewhere else or not at all, and none offers a way out once started. Somebody who launched an analysis over forty thousand tracks and wants their laptop back has to quit the application.

## Shape

**One place that lists what is running**, with a progress figure where one exists and an indeterminate state where it does not, and a cancel control per task. Something a user can glance at to answer "why is this machine busy", which is the question that actually gets asked.

**Cancellation that is honest about its granularity.** Most of these tasks are loops over tracks; stopping between two items is easy and stopping mid-item is not. A cancel that takes effect at the next item boundary is fine, provided the button does not claim otherwise.

## Our side

The pieces exist unevenly. Scan already has progress events. Reconciliation gained cancellation through an `AtomicU8` phase machine, which is the pattern to generalise rather than reinvent. The plugin host, the analysis sweep and the artwork pipeline each have their own notion of running or not.

Worth doing as one registry the tasks report into, rather than a fourth bespoke progress surface — the current state is the argument for that.

## Traps

- **SQLite takes one writer at a time**, and a scan is a writer ([invariant](https://github.com/InstaZDLL/WaveFlow/blob/main/docs/architecture/invariants.md#single-writer-to-sqlite)). A status bar that invites starting three things at once has to reflect that some of them will queue rather than run.
- **Cancelling must leave consistent state.** A half-applied batch is worse than a slow one; each task defines its own safe stopping point.
- **Do not let it become a notification centre.** This lists what is running now, not a history of what has run.
- New strings go to all 17 locales.

### LINKED ISSUE #619: feat: choose where artwork and caches are stored
Requested by email, in Russian:

> Приложение сохраняет artwork на системный диск даже при условии, если само установлено на другой. Но если на диске С критически мало места, возникает проблема в виде системных сбоев и тормозов. Хотелось бы иметь возможность выбора директории сохранения artwork

In English: the app stores artwork on the system drive even when WaveFlow itself is installed on another one. When `C:` is critically low on space, that causes system errors and slowdowns. They would like to choose where artwork is saved.

## Today

Everything lives under Tauri's app-data directory: `%APPDATA%\app.waveflow\waveflow\` on Windows (`paths.rs`). The installation directory plays no part, and there is no setting to move it.

| Path | What it holds | Grows with |
| --- | --- | --- |
| `metadata_artwork/` | shared remote artwork cache | library size |
| `motion_cache/` | animated album covers (`.mp4`, LRU) | usage |
| `canvas_cache/` | per-track Canvas clips (`.mp4`, LRU) | usage |
| `profiles/<id>/artwork/` | per-profile covers and thumbnails | library size |
| `profiles/<id>/motion/` | manual motion covers, never evicted | usage |
| `profiles/<id>/data.db` | the profile database | library size |
| `avatars/`, `app.db` | small | — |

## Options

1. **Move only the caches** (`metadata_artwork`, `motion_cache`, `canvas_cache`, per-profile `artwork`). These hold the bulk of the size and can be rebuilt, so a failed move can fall back to "clear and re-fetch".
2. **Move the whole data root.** Simpler to explain, but databases are involved, and moving those is a real migration, including a restart.

## Constraints

- **The asset protocol scope is static.** `tauri.conf.json` only allows `$APPDATA/…` and `$APPLOCALDATA/…` paths. Artwork served from another drive needs a runtime scope grant, or the images won't load.
- **Moving means copying existing files** and switching over only once the copy completes. An interrupted move must not leave a profile pointing at a half-filled directory.
- **A removable or network drive can be missing at launch.** That needs a defined fallback (caches: recreate at the default location; databases: refuse clearly).
- **The pre-flight schema check** resolves the root before the app exists (`root_for_identifier`). A custom root must be readable at that point too, so it can't live only in `app.db`.

InstaZDLL:
Lot D of the v1.8.0 milestone — the library interface.

Closes #588.
Closes #589.
Closes #596.
Closes #601.
Closes #619.

### #596 — high contrast

A third axis beside theme and skin: `auto` / `normal` / `high`, per profile, `auto` deferring to the OS `prefers-contrast: more`.

The issue assumed colour already lived in WaveFlow's theme tokens. It does not — only the accent does; the greys are ~1 900 raw `text-zinc-*` utilities. The conclusion survives for a different reason: Tailwind v4 defines its own palette as CSS variables, so the same `@theme inline` that lets a theme re-tint every `bg-emerald-*` can re-point the whole grey scale. Only shades 300-600 move, because measured across `src/` those are secondary *text* and almost nothing else, while 700-900 are each a background, a border **and** text at once.

Three things that would each have made the mode worse than no mode:

- The greys invert by ground, so the dark-mode selector must also match `[data-skin="lounge"]` and `[data-skin="pulse"]` — those two are dark by skin and never carry `.dark`. Without it the mode inverts on two of the five skins.
- Turning `backdrop-filter` off is half a fix. Pulse's sticky header is `transparent` and Liquid's chrome is an 8-16 % tint; unblurred and still translucent, their text sits directly on the content scrolling underneath. Each such surface is forced opaque in the same breath.
- Written with the standard property before the prefixed one, the minifier folds the two and emits only `-webkit-backdrop-filter`. Verified in the built CSS.

Also fixes a defect in the first-paint bootstrap next door: `VALID_SKINS` never listed `liquid`, so a Liquid user was stamped `data-skin="studio"` and got exactly the flicker that block exists to prevent.

### #619 — where the caches live

Reported by email: artwork landed on the system drive whichever drive WaveFlow was installed on.

The split is not "big things move" but **evictable moves, chosen stays**. Everything under the new `cache_root` is content-addressed or LRU-evicted, so the worst case of a failed move is a re-fetch; the databases, the hand-picked motion covers and Canvas clips, and the offline downloads stay put.

Four hazards, each silent if missed:

- `AppPaths` is captured once at boot and ~95 call sites read it directly, so a move cannot take effect in the running process. It copies, persists, and asks for a restart; the old tree is removed by a startup pass when nothing is reading it.
- The asset-protocol scope is static, so artwork on another drive fails to load with no error and no console message. It is widened at runtime — and **only to the four cache directories**, never to the chosen root: that root can be `D:\` or a home folder, and granting the webview recursive read access to all of it would be a privilege escalation.
- A `.waveflow-cache` marker guards every recursive delete. The layout is a set of ordinary names — `metadata_artwork`, `profiles` — under a folder the user picks in a dialog, so without it, pointing WaveFlow at a folder that already has a `profiles/` directory would have had it deleted on a reset.
- `reset_app` wiped `AppPaths::root` and called it done, which stopped being the whole story.

### #601 — one status bar for long-running tasks

The issue says none of these could be stopped. Not true of the current code: five of them already could, each with its own safe stopping point. So the registry deliberately does **not** implement cancellation — a task registers a callback and the registry routes to it. Re-deriving five stopping points would have got at least one wrong.

The genuine gaps were the single surface and the library scan, which had no mechanism at all. Adding one surfaced the trap the issue warns about in the abstract: **the scan ends with a pass that marks everything the walk did not reach as `is_available = 0`.** Stopping halfway and letting that run would mark a working library unavailable *because the user pressed stop*. A scan's safe stopping point is not between two files — it is before that pass.

Every other operation turned out to have an honest stopping point too, down to the thumbnail pass, which stops between two files.

### #589 — the "needs attention" inventory

Twelve counted categories in a new library tab, each clicking into the library's own table — same columns, sort, context menu and properties modal, so fixing a track from there is the ordinary edit flow.

The issue says the untaggable formats are `.dsf` and `.dff`. No longer true: lot B made `.dsf` writable, so only `.dff` remains.

Two things decide whether this is useful or merely present:

- **Probable duplicates chain, they do not compare pairs.** 180 s, 181.5 s and 183 s are one recording, but the outer pair is 3 s apart — a pairwise tolerance splits them, and *which* two it splits into depends on iteration order, so the same library answers differently between runs. The chaining lives in `waveflow-core` with tests.
- **"Gap in the numbering" measures holes inside the observed range.** The obvious spelling assumes every disc starts at 1 and reports a box set's third disc numbered 20-22 as missing nineteen tracks.

Every category was verified against a database built from this repo's own migrations, constraints and foreign keys on — twelve expected id sets. That is how `album.album_artist_id` turned out not to exist (the column is `album.artist_id`; `album.album_artist` beside it is free text).

### #588 — choose, reorder and resize the columns

Seventeen built-in columns, chosen and reordered from a picker, resized by dragging, fitted by double-clicking, sorted in the database in either direction, under a header that stays put while the list scrolls. Per profile, shared by every list that renders the same table.

Sorting had quietly stopped working at seven columns: the order clause spelled out two `match` arms per column. It is a whitelist now — which is also the only reason the caller may hand the result to `AssertSqlSafe` — with NULLs last in *both* directions and a total order so two identical queries cannot return two different rows first.

**Columns from the user's own tags cost a second read of each file, and the issue mispredicted that.** It assumed the reading existed because `edit.rs` goes through the concrete tag type per format — true of the *writer*. The scanner reads through `lofty::read_from_path`, whose generic `Tag` has no `Unknown` variant and keeps the remainder in a `pub(crate)` slot, or for Vorbis comments not at all. Both halves of the split are read now: the remainder **and** the generic half minus what WaveFlow already stores — because `Composer` has an `ItemKey` and never reaches the remainder, so a version reading only that missed the very field the issue opens with. The extra parse is timed on its own as `extra_tag_cpu_ms_total` rather than folded into the scan total.

A library that predates the feature has files whose mtime and size have not moved, so one forced pass per folder is marked in `profile_setting` and never repeats.

## Validation

| | |
| --- | --- |
| Windows | `clippy -D warnings` 0 · `typecheck` · `lint` · `prettier` on the touched files · 213 `waveflow-core` tests |
| Fedora 44 | `fmt --check` · `clippy -D warnings` 0 · **605** app tests + **316** core |
| macOS 26.6 | `fmt --check` · `clippy -D warnings` 0 (CoreAudio compiled) · **593** app tests + **316** core |

The inventory's twelve SQL categories are checked against a database built from the repo's migrations with `foreign_keys = ON`, one expected id set each.

## Local review

Nine passes before this branch was pushed, 24 → 16 → 8 → 9 → 17 → 11 → 5 → 9 → 23 findings. Several passes were made longer by the previous pass's own fixes. What the review caught that reading would not have:

- The scan that had just been made cancellable **could not be cancelled** — `cancel()` returned before setting the flag when a task registered no callback, so the button never even appeared.
- Choosing `D:\` as the cache folder gave the webview read access to the whole drive.
- A folder already containing `profiles/` would have been deleted by a reset.
- "Back to default" was blocked by the guard added against that.
- The data rows lost their grid: `gridCols` stopped being a Tailwind class and only the header was converted to an inline style.
- A column resize wrote to the database on every pointer move.
- The **ReplayGain** backfill marked itself done on a cancelled scan — a bug created by making the scan interruptible.

Declined findings carry their reason in the code.

## Notes

- `track_tag` is a new profile table; values appear after the next scan of each folder.
- Moving the caches asks for a restart, by construction.
- **#595** (software-rendering fallback after a GPU start-up crash) is in the milestone but belongs to no lot — it is backend start-up work, not library UI.

## Summary by CodeRabbit

  - Suivi centralisé des tâches longues avec progression, détails et annulation.
  - Nouvel onglet « À corriger » pour identifier les problèmes de bibliothèque et doublons potentiels.
  - Colonnes personnalisables avec tri, redimensionnement et tags personnalisés.
  - Emplacement des caches configurable et modes de contraste normal, élevé ou automatique.

- **Améliorations**
  - Annulation des opérations longues avec conservation des résultats déjà traités.
  - Résultats de sauvegarde plus précis, y compris les exécutions partielles.
  - Contraste, focus clavier, mini-lecteur et accessibilité améliorés.

- **Documentation**
  - Documentation actualisée sur les tâches, le stockage et l’inventaire.


InstaZDLL:
@coderabbitai review

InstaZDLL:
@coderabbitai All eleven acted on in `6394fe32`, and four of them were
better than they looked from the diff. Thank you.

**The cache location.** The first two findings are one defect seen from
two sides, and you named the thing I had missed: while a move waits for
its restart, the caches exist under *two* roots at once.

- `profile.rs` — deleting a profile removed its cache from the active
  root only. The part that makes it permanent is the part I had not
  followed through: after that command, nothing can enumerate a deleted
  profile again, because `cache_dirs` builds its list from the `profile`
  table. Centralised in `commands::storage` as
  `profile_cache_dirs_elsewhere`, ownership check applied per root. The
  same hole existed one level up in `wipe_targets_outside_root`, which a
  later pass caught: a reset staged right after a move wiped the old
  tree and left the new one — the copy the app was about to start
  reading.
- `set_cache_location` — mutex added to `AppState`, held from before the
  pending-move check through the last persistence write, on the model of
  the existing `drain_lock`. You are right that the card's `busy` flag
  guards a button and not the IPC surface behind it.
- `get_cache_location` — `cache_root_fallback` is the discriminator, as
  you proposed. This one is my favourite finding of the batch: the two
  states overlap *by design*, because `cleanup_moved_caches` keeps the
  marker precisely when the session fell back, since the tree it names
  is the one being read. Reading the marker alone left "restart
  WaveFlow" on the card for a restart that could never finish.

That last one led somewhere worse, which a local pass then found:
returning to the default folder in that state was refused **twice** —
by the staged-move guard, then by the containment check, which is right
to call a tree "inside the folder being moved" when it *is* that folder.
No way out at all. Choosing the root the caches already sit at now means
"cancel the staged move": nothing is copied, and the cleanup marker is
retargeted at the abandoned destination so the copy staged there is
removed rather than orphaned.

**`tasks.rs`.** Real, and the window is narrower than the fix. Each task
now carries its own liveness mutex, taken by `cancel` and by `Drop` —
never both at once with the registry lock, so the two can only order.
That keeps the callback out of the registry's mutex, which was the
reason it was called outside it in the first place.

**`extra_tags.rs`.** Both correct, and they are the same defect: the
same field offered as two columns. `COMPOSER` from a `TCOM` frame beside
`Composer` from a `TXXX` one, and a WAV's `IENG` beside a FLAC's
`ENGINEER`. Duplicate detection is case-insensitive now with the stored
casing untouched, and recognised RIFF chunk ids are named through
`map_key` the way `generic_into` names things.

You were also half right in a way the finding did not claim: folding
*within* a file is not enough, because the picker groups across files.
`track_tag.key` is `COLLATE NOCASE` now — edited into the original
migration rather than followed by a second one, since it has never been
merged and no profile carries it.

**The frontend four.** All applied: `motion-safe:animate-pulse`; an
emptied category closes itself rather than leaving a dead table under a
disabled chip; `Object.prototype.hasOwnProperty.call` (not
`Object.hasOwn` — the project targets ES2020); and `useContrastMode`
gates both effects on `ready`. That last one was the worst of the four
in practice: before the profile read landed, `value` was still the
default `auto`, so painting it wrote `normal` over the `high` the
bootstrap had just stamped — a flash of ordinary contrast at every
launch, for the one user who asked for the opposite. The cache write
moved into that effect, so a failed write now rolls back both.

**The out-of-diff `MiniPlayer` finding is the best one here.** That
window paints itself from a gradient sampled off the cover art, so a
pale album gives white text on a pale ground — in the one mode whose
whole promise is that it will not. `.wf-mini-player-surface`, forced to
an opaque dark panel under `[data-contrast="high"]`;
`background-image` as well as `background-color`, since the gradient is
an inline `background` shorthand, and dark rather than
`--wf-contrast-surface` because everything in that window is
`text-white`.

Validated on `6394fe32`: Windows clippy 0 / typecheck / lint / prettier,
Fedora and macOS `fmt --check` and clippy 0, 608 app tests and 317 core.

InstaZDLL:
@coderabbitai Two of the three are fixed in `8448b380`, the third is declined.

**`title_from_filename` on the hash-identical path — fixed, and it mattered more than the comment suggested.**

You were right, and the consequence is larger than "existing rows are updated from 0". The migration adds the column as `NOT NULL DEFAULT 0`, so every row in an already-scanned library starts at `0` while its title is in fact the file stem. A file whose bytes never change never reaches the rewrite branch — so without this write those rows would have stayed at `0` for the life of the install, and `missing_title` would have gone on reporting zero exactly as it did before this PR. That is the very fault the column was added to repair, surviving in the upgrade path. The migration's own comment promised that "the category under-reports until each folder is scanned again", which was untrue until this commit.

Assigned rather than `COALESCE`d, since the column is `NOT NULL` and never falls through.

**The fade under `prefers-reduced-motion` — real gap, fixed one level down.**

`animate-fade-in` carries an 8px rise and was covered by no reduced-motion rule, so the finding holds. I fixed it in `src/app.css` instead of at the two call sites: the utility has ~30 of them across the app, and `app.css` already has four `@media (prefers-reduced-motion: reduce)` blocks that null out animations exactly this way. A `motion-safe:` variant would have covered the two reported overlays and left the rest — and it appears only once in the repo, on `animate-pulse`, where it was the only available lever because that utility is Tailwind's rather than ours. `animation: none` lands the element on the `to` state, so content appears in place rather than not at all.

**Blocking the custom-tag marker when extraction itself failed — declined.**

The reason is already written above that branch. A file whose extraction failed got nothing at all from the scan, not merely its tags, and it is already counted in `summary.errors`. A permanently unreadable file fails every time, so holding the marker back for it would re-run the forced full-folder tag pass on every scan for the rest of the install — one corrupt file quietly costing a whole folder its fast path. The asymmetry with the ReplayGain backfill beside it is deliberate: that one only blocks for files the pass named in `rg_missing`, whereas the tag pass is whole-folder. The case that does block the marker is the one below, where extraction succeeded and only the tag read returned `None` — nothing else reports or retries that.

Validated on `8448b380`: Windows `cargo fmt --check`, clippy `-D warnings` at 0, prettier; Fedora fmt, clippy 0, **609 app + 318 core tests**. No platform-specific code is touched, so CoreAudio was not re-checked.

Issue #585, opened 2026-09-08
feat: a plugin world that can carry word-level lyrics


InstaZDLL:
A plugin cannot return word-level lyrics today, and that blocks every provider worth plugging in.

## What the interface can carry

`waveflow:metadata/v1` has a `lyrics` function, and this is its whole vocabulary:

```wit
record lyrics-line {
  /// Timecode in milliseconds from track start. `none` = the
  /// line is plain (un-synced) text.
  time-ms: option<u32>,
  text: string,
}

lyrics: func(artist: string, title: string) -> result<list<lyrics-line>, string>;
```

A line, an optional start, some text. There is no per-word timing, no end timestamp, no translation, no pronunciation, no per-line singer, no marking of background vocals.

## Why that is a problem rather than a limitation

The app on the other side of this interface is already richer than the interface. `commands/lyrics.rs` parses Enhanced LRC and TTML with per-word `begin` / `end`; `line.words` and `activeWordIndex` drive per-word highlighting in both the immersive column and the docked panel; and `useKaraokeWordFill` paints a continuous sweep across the word being sung, extrapolating between the 4 Hz `player:position` events.

So a plugin holding word-level lyrics has to throw the timing away to hand them over, and what arrives is line-synced text — indistinguishable from LRCLIB, which is already in the chain, costs nothing and needs no credentials. Any word-level provider is worth exactly as much as LRCLIB through this interface, which means it is not worth writing.

This is not about one provider. Musixmatch is word-level too and sits in-tree today, and moving the in-tree providers out to plugins is on the roadmap.

## The versioning constraint, which is the real work

The WIT file records its own rule, and it now cuts against us:

> **NOT an ABI-compatible bump**: adding a field to a record changes its Component Model type […] This is safe ONLY because there were zero `waveflow:metadata@1.0.0` consumers — the world was defined but never bound or shipped. A real 1.0.0 consumer would have required a fresh interface version instead.

There is now a real consumer: a shipped plugin declares `world = "waveflow:metadata/v1"` and is listed in the public registry. So the next change to this interface cannot be an in-place bump. It needs either **`waveflow:metadata/v2`** or a **dedicated `waveflow:lyrics/v1` world**.

Two shapes, and they differ in more than taste:

- **Extend metadata to `/v2`.** One world keeps artwork, bios and lyrics together. Existing metadata plugins must be rebuilt and re-declared to move.
[...]

Issue #584, opened 2026-09-08
feat: Apple Music lyrics — word-synced, with translations and pronunciations


InstaZDLL:
Apple Music serves its lyrics as **TTML with per-word `begin` / `end` timestamps**, and alongside the original it can return Apple's own translation and a Latin-script pronunciation for the same song. That is the richest lyrics format we could realistically consume.

**Shape: a plugin, not an in-tree provider.** It needs the user's own Apple Music session, which is exactly the kind of integration that belongs outside the app — the same call already made for Spotify Canvas.

**Blocked on #585.** No plugin world can carry word-level lyrics today: `waveflow:metadata/v1` returns a line and an optional start time, nothing more. Through that interface this provider would deliver line-synced text, indistinguishable from LRCLIB — which is already in the chain, costs nothing and needs no credentials. So #585 comes first, or this is not worth writing.

## Why it fits once the world exists

- **We already parse word-level TTML.** `commands/lyrics.rs` handles Apple-style `<span begin="…" end="…">` word timing, alongside Enhanced LRC.
- **We already render it.** `line.words` and `activeWordIndex` drive per-word highlighting in both the immersive column and the docked panel, and `useKaraokeWordFill` does the progressive sweep, extrapolating between the 4 Hz `player:position` events so the fill is continuous rather than stepping every 250 ms.
- **The lyrics chain is already tiered and cached** — DB cache, embedded tag, sidecar file, then online providers — so this is a new tier, not a new subsystem.

Today the only online source of word-level timing is Musixmatch, which is off by default and, as things stand, can only be switched on by editing a row in SQL.

## What it would add beyond word timing

Each of these is a separate decision rather than a free bonus:

- **Translations and pronunciations.** These land directly on the translated-lyrics rework, which is Musixmatch-only today. They have to be joined to lines by a stable key rather than by array position, because a localized document may omit a line.
- **A per-line singer**, for duets and multi-vocalist songs.
- **Background vocals**, marked separately from the main line.

The last two have no equivalent in `LyricsPayload` today. Dropping them on ingest is a fine answer — but it should be a decision, not an accident.

## Constraints to settle before any of this is drawn

[...]

PR #662, opened 2026-09-16
feat(lyrics): show apple romanization and translation under each line

InstaZDLL:
An Apple TTML document can carry two further readings of every line — a translation and a Latin-script transliteration — and the parser walked straight past both. They do not sit beside the lines but in `<head>`, under `<iTunesMetadata>`, as entries pointing back at a line through its `itunes:key`. Nothing downstream could show them because nothing upstream read them.

This is the **rendering half of #584**. The plugin that fetches such a document is the other half and lands separately; the format already arrives today through an imported `.ttml`, which is how every claim below was verified against a real Apple document.

## The measurement the whole design rests on

On a full document (54 lines): one localization entry per line for each reading, and the transliteration carries **one span per original span, with identical `begin` / `end` bounds** — 54 lines out of 54, no exceptions.

So `activeWordIndex` addresses both rows at once. A romanized word is driven by the clock of the word it reads out, which means **no second timing pass, no realignment, nothing to keep in step** — the existing progressive fill and the existing active-word index simply apply. The translation, by contrast, is line-level with no word split, which is also how it reads on screen.

A single localized request returns both readings: asking Apple for a translation language is what brings the transliteration along with it.

## The trap that only shows on a real song

**Apple localizes every line, including those already in the target script or language.** In a Korean song, an English line comes back transliterated as itself and translated as itself.

Printed under the line, that is the same line twice. The misleading case is not the exact duplicate but the near one: when the syllable split differs from the word split, the eye reads a discrepancy that is not there. Apple Music hides these, and so does this — both readings are dropped when they say what the line already says, ignoring spacing and case.

## Rules written into the parser, each guarding a silent failure

- **Join by key, never by index.** A localized document may omit a line; matching by position would shift every following entry onto the wrong line with nothing looking broken.
- **A word split that does not match costs the split, not the reading.** Measured, it always matches — but nothing upstream promises it, and pairing words as far as they go would put the highlight on the wrong one, which reads as a broken transliteration rather than a missing feature. The text still reads correctly, so it is kept unsplit.
- **A reading that adds nothing is dropped** (above).
- **Only the first `<translation>` / `<transliteration>` is read.** Apple returns one of each because the language is chosen in the request; several would need the user to pick, which a parser cannot ask.

A document that asked for neither still carries an empty `<translations>` container, so absence is the normal case rather than an error.

## Preference

`lyrics.localization_mode`, per profile: off / romanization / translation, one at a time as Apple Music presents it — the two answer different questions, and a third row under every line turns a lyric column into a table.

It is **resolved** against what the current document carries rather than written back, so turning it on for one song survives the next song that has neither and applies again on the one after. The toggle sits in the lyrics panel header and appears only when there is something to show; the immersive column reads the same preference, so opening one view over the other never shows two different readings of the same line.

## Known limitation, in the code and the docs

The immersive column gives the progressive fill to the original line only. `useKaraokeWordFill` returns one ref callback for one element, and the sweep belongs on the line the eye is following. The romanization takes the discrete highlight, which still marks the word being sung.

## Validation

Verified on a real Apple document, both readings, in the docked panel and the immersive column — including the line-level fallback path.

The localization join was additionally checked outside the browser against both the plain and the localized document: 54/54 lines paired for each reading on the localized one, zero dropped for a count mismatch, and nothing at all produced from the plain one despite its empty container. No frontend unit tests, per this repo's convention.

`typecheck`, `lint` and `prettier` green. Three local CodeRabbit passes, the last at zero; all three earlier findings were sound and are fixed:

- **A transliteration with no spans was being discarded** — a line-timed Apple document carries exactly that, and its romanization vanished entirely. Fixed by modelling a reading as `{ text, words? }`, so the text always survives.
- `aria-pressed` on a button that cycles three states implies a two-state toggle — removed.
- zh-TW said 列 (column) where zh-CN said 行 (row).

i18n: 3 keys across all 17 locales.

Part of #584. Builds on #585.

## Summary by CodeRabbit

  - Ajout de la romanisation et de la traduction des paroles Apple Music.
  - Choix entre les modes désactivé, romanisation et traduction.
  - Bouton de changement de mode disponible lorsque nécessaire.
  - Affichage synchronisé de la romanisation et des traductions sous les paroles.
  - Préférence de localisation conservée par profil.

- **Internationalisation**
  - Ajout des libellés dans les langues prises en charge.

- **Documentation**
  - Documentation des règles de prise en charge et d’alignement des localisations.


InstaZDLL:
@coderabbitai Both findings were correct and are fixed in `6426295e`.

**The stale documentation rule.** You were right, and it is the worst kind of wrong: the rule was accurate when written, then the behaviour changed under it during review (a mismatched split used to drop the whole romanization, and now costs only the split) while the doc kept the old sentence. It now describes what the code does — words discarded, text kept unsplit under the line, the same fallback a line-timed document lands on.

**The unnormalized bounds.** Also correct, and worth spelling out because the impact is not where it looks. `fillEndTimestamps` closes an open `endMs` on `line.words` and knows nothing about `romanization.words`, so a document that omits `end` would leave the two sets of bounds disagreeing — while `LyricsReading` promises they are identical.

Nothing reads those bounds today: the rendering follows `activeWordIndex`, computed from the line. But the documented limitation of this PR is that the romanization gets no progressive fill, and whoever lifts that limitation is exactly the reader a leftover `-1` would bite. The bounds are now copied from the matching original word after `fillEndTimestamps` has run, which makes the type's promise true by construction rather than by trusting the document. Measured on a real Apple document the two are identical anyway, so this changes no timing — it removes the way they could quietly stop being identical.

Two further findings came from the local CLI on the same code:

- **The accessible name of the mode button** now carries the action *and* the resolved state (`Change the extra reading. Showing romanization`). With three states there is no `aria-pressed` to hold the state, so neither half can be dropped — my first attempt traded one for the other.
- **A text node between two spans** is now normalized rather than replaced by a fixed space, so punctuation placed outside the spans survives.

One local finding I did not apply: ignoring a text node whenever it contains a newline, to keep XML indentation from being appended between words. `replace(/\s+/g, " ")` collapses any run of whitespace to a single space, so indentation cannot be appended — the case the suggestion guards against cannot occur. Applying it would instead remove the separator between words in an indented TTML, changing pre-existing behaviour in the wrong direction.

`typecheck`, `lint` and `prettier` are green, and the local review is at zero.

InstaZDLL:
@coderabbitai Correct, and fixed — the useful part of this finding is that it caught two opposite policies living in the same function.

The rule I had written and documented is "first container wins": a document carrying several translations would need the user to choose, which a parser cannot ask. But within that container, a duplicated `for` key fell to `Map.set`, which keeps the LAST entry. So the same function resolved ambiguity one way at the container level and the other way at the entry level, and only one of the two was written down.

Both now keep the first, and the function's doc comment says so, including why the per-key half exists at all — nothing should point two entries at one line, which is exactly the kind of assumption that stays true until it does not.

`typecheck`, `lint` and `prettier` green; local review at zero.

InstaZDLL:
@coderabbitai Right on both counts, and fixed — plus one I am declining, with the measurement behind it.

**Namespace-qualified containers.** Correct, and it exposed an inconsistency of mine: I already handle prefixes for the `itunes:key` attribute (`getAttributeNS` with a literal fallback) but looked the containers up by qualified name, so `<itunes:translation>` would have been missed silently — original lyrics fine, localizations quietly absent. Lookups now go through a `localName` helper, and I extended it to the `<p>` scan and the `<span>` test for the same reason: a parser that understood prefixes for the localizations but not for the lines they attach to is the kind of half-rule that reads as a bug later. No behaviour is lost — a document prefixing its TTML elements parsed to nothing before.

**Dead exports.** Also correct. `ttmlLocalizations` and its type were written to label the toggle with the actual language tags; the labels ended up describing the state generically instead, and I never deleted the helper. Gone, along with `langOf` and `XML_NS`, which existed only to feed it — the typecheck found that last one for me.

**Declined: recursing into a `<span>` container that has no `begin`.** Measured on the real Apple document first: 480 direct spans, **zero** without a `begin`, so zero untimed containers. The guard is also pre-existing — this PR moved that line into `ttmlWords`, it did not introduce it. And the one shape that would produce such a container is background vocals, which #584 explicitly lists as a separate decision ("Dropping them on ingest is a fine answer — but it should be a decision, not an accident"). Changing how they parse as a review fix would make that decision by accident, which is the thing the issue asks not to do.

`typecheck`, `lint` and `prettier` green.

PR #664, opened 2026-09-16
fix(settings): move the video cache controls to Settings > Data

InstaZDLL:
A lyrics-only plugin showed **"Save animated artwork locally"** in its ⚙️ panel — with the same 486 MB footprint and the same *Clear cache* button as the motion-artwork plugin listed just above it. Nothing about the lyrics plugin asked for that control.

## Why it was there

`PluginOptions` rendered the motion-cache control for **any** `waveflow:metadata` plugin, and the Canvas-cache control for any `waveflow:canvas` one. That rule was wrong in two ways:

- **The caches are app-wide.** Every plugin of the world showed the same switch and the same numbers, and the *Clear cache* under one plugin deleted files another plugin had produced.
- **A world does not say what a plugin produces.** `metadata/v2` covers artist info, album details and lyrics. The first plugin that ships lyrics without motion covers inherited a control unrelated to anything it does. The rule only ever looked correct because, until now, the only metadata plugin was the motion-artwork one.

## What changed

- Both controls now live in **Settings → Data**, directly under the *Cache location* card (#619), which already decides the drive both caches live on (`MediaCachesCard`).
- The plugin ⚙️ panel shows only the options the plugin's own manifest declares. The gear appears only when the manifest declares at least one.
- The two controls are always shown in Settings → Data, not only while a producing plugin is installed. The host has no record of which plugins produce what, and a cache that still holds files after its plugin was uninstalled is exactly the one a user wants to clear.

Smaller follow-ons:

- `isMetadataPlugin` and `isCanvasPlugin` had no callers left, so they are removed.
- The Canvas cache subtitle said "from the plugin", which means nothing once no plugin is shown next to it. It is reworded in all 17 locales.
- The size readout now uses the shared locale-aware `formatBytes`, as the *Cache location* card already does, instead of a private copy.
- `docs/features/plugins.md` and `docs/features/ui.md` no longer point to the ⚙️ panel for these controls.

## Validation

`typecheck`, `lint` and `prettier` are green on the changed files. The local CodeRabbit review returned one finding: the Hindi plural, where `hi.json` already uses `प्लगइन्स`. It is applied; the review now returns zero.

## Summary by CodeRabbit

  * Ajout, dans **Réglages → Données**, d’une carte dédiée aux caches vidéo Canvas et motion artwork.
  * Consultation de l’état, de la taille et du nombre de fichiers, avec activation, désactivation et effacement de chaque cache.
  * Configuration d’un emplacement commun et confirmation avant tout effacement.

* **Améliorations**
  * Les options des plugins affichent uniquement les réglages déclarés et apparaissent lorsqu’ils existent.
  * Les descriptions du cache Canvas ont été clarifiées dans les langues disponibles.
  * Les ressources supprimées ou périmées ne sont plus affichées après l’effacement d’un cache.


InstaZDLL:
@coderabbitai Correct, fixed in `58463dd1`.

Checked before changing anything: `set_cache_location` persists the chosen root, and `AppPaths::with_cache_root` rebuilds both `motion_cache_dir` and `canvas_cache_dir` under it at the next launch. So `<app-data>/waveflow/` is only where these caches start, and this PR — which puts their controls directly under the card that moves them — was a poor place to keep describing them as fixed.

All three occurrences now read as `motion_cache/` / `canvas_cache/` **under the cache root**, `<app-data>/waveflow/` by default or wherever *Settings → Data → Cache location* moved it. No hard-coded `<app-data>/waveflow/{motion,canvas}_cache` path remains under `docs/`. Local review at zero.

InstaZDLL:
@coderabbitai The clear finding was right, and fixed in `15d9dc1e` — with one change of scope.

With a cache on, the backend answers with the **local path** of the downloaded mp4, and `useAlbumMotionArtwork` / `useTrackCanvas` remember it. *Clear cache* deleted the files but kept those answers, so every surface already resolved kept pointing the webview at a missing file until restart.

Scope: the suggestion was to invalidate one album (`invalidateAlbumMotionArtwork`) or one track (`invalidateTrackCanvas`), but a clear empties the **whole** cache, so every other entry would stay stale. The card now calls `invalidateAllMotionArtwork()` after a successful motion clear, and a new `invalidateAllTrackCanvas()` after a successful Canvas clear. That one bumps every known track's generation before dropping the maps, so a lookup still in flight — possibly about to return one of the deleted paths — lands as stale instead of repopulating the cache. Nothing is invalidated when the clear fails: the files, and so the paths, are still there.

Declined, from the local review on the same code: also invalidating when the **toggle** changes. Neither direction leaves a broken source:

- **Turning the cache off does not delete anything** (`setMotionCacheEnabled` / `setCanvasCacheEnabled` never purge), so remembered local paths still point at existing files.
- **Turning it on** leaves remembered remote URLs, which still stream.

Invalidating there would only make every mounted video reload — a visible flicker, and a fresh download or re-stream — to fix nothing. The clear is different precisely because it deletes the files those paths name.

InstaZDLL:
@coderabbitai Right, and it went further than `useTrackCanvas` — fixed in `b8805d19`.

Invalidating cleared the shared maps and bumped the counter, but each hook's own state kept the previous answer until the re-resolve landed; after a cache clear that answer names a deleted mp4. Both ways in are closed:

- **The render** now requires the held state to carry the epoch (Canvas) / generation (motion) that is still current, and reads as "nothing" otherwise.
- **The in-flight apply** now also checks the live counter. `cancelled` alone could not catch it: the counter moves synchronously, but the effect is only cleaned up at the next commit, so a lookup started before the invalidation could resolve in between.

`useAlbumMotionArtwork` had the same shape — its reset to `null` goes through a microtask, so the render right after an invalidation still returned the old URL — and gets the same gate. Cost, accepted and stated in the commit: any invalidation briefly blanks every mounted surface of that kind until its answer returns, which is a microtask from the shared cache for everything the invalidation did not drop.

On invalidating after a **toggle**, raised again by the local review: still declined, and the reason now sits at the call site so it is read where the question comes up. Turning a cache off deletes nothing, so remembered local paths still name existing files; turning it on leaves remembered remote URLs, which still stream. The source on screen stays playable either way, and invalidating would only reload every mounted video to fix nothing. A clear is the opposite case — it deletes the files those paths name — which is why it, and only it, invalidates.

PR #671, opened 2026-09-17
feat(lyrics): a floating desktop lyrics window

### LINKED ISSUE #582: feat: floating desktop lyrics window, independent and always on top
Requested in [discussion #503](https://github.com/InstaZDLL/WaveFlow/discussions/503) by @JQuer-lab.

**Stalled on purpose**, not rejected — see the last section. The cheaper shape of the same need is tracked in #580 and is the one being pursued first.

## What is being asked

A **standalone** desktop lyrics window, distinct from anything in the player window:

1. A separate transparent window that stays on top of other applications
2. Draggable anywhere on the desktop
3. Independent of the main window — works while it is minimised
4. Customisable font size, colour, text outline and transparency
5. An optional click-through mode so the window does not intercept mouse input

The requester was explicit that this is not the immersive side panel (#328), which lives inside the player window.

## Where we stand

Lyrics render in two places today — the docked right-hand panel (`AppLayout`) and the immersive view (`ImmersiveSidePanel` → `ImmersiveLyricsColumn`). Neither is a desktop overlay, and the mini-player has no lyrics at all.

The data side is not the hard part. `useTrackLyrics()` already handles fetching, the current track, position and seeking, and `useKaraokeWordFill` already drives the progressive word sweep, extrapolating between the 4 Hz `player:position` events. A third window mounting `PlayerProvider` would get both for free, exactly as `MiniPlayerApp` does.

## What makes it expensive

The window itself, not the lyrics.

- **Transparency and click-through are per-platform.** A transparent, decorationless, always-on-top window behaves differently on Windows, on X11, and on Wayland — where a client cannot freely position its own surfaces, so "drag it anywhere on the desktop" and "stay above everything" are not guarantees the compositor is obliged to honour. This is the same class of platform unevenness that makes global hotkeys awkward (#572).
- **Click-through needs a real hit-test policy**, not just a flag: the window has to stop intercepting the mouse while still offering a way to reach its own settings and to move it.
- **Five customisation options is a settings surface** — font size, colour, outline, transparency, plus whatever positioning state has to persist — with strings across all 17 locales.
- **A third webview has a memory cost** that the mini-player already pays once.

## Why it is stalled rather than started

#580 addresses part of the same use case — keeping lyrics visible while working in another application — while reusing machinery already present in the mini-player: an existing panel-toggle pattern, a window that already resizes and remembers its size, and a provider tree already mounted.

For that reason, #580 is the better first step. It lets us ship a useful version of this workflow without committing immediately to the platform-specific complexity of a full desktop overlay.

This issue remains separate because #580 does not provide everything requested here — notably transparency, click-through, independent positioning and the additional appearance controls.

Once the mini-player lyrics view has shipped, we can revisit this issue with real-world feedback and decide what is still missing from the standalone overlay experience.

InstaZDLL:
A floating desktop lyrics window: transparent, always on top, independent of the main window. The five points of the issue:

1. **A separate transparent window above other apps.** Undecorated, no shadow, out of the taskbar, on every workspace. It opens without focus and stays up while the main window is hidden to the tray.
2. **Draggable anywhere.** When unlocked, hovering shows a frame with lock and close buttons, and dragging anywhere moves the window. Position and size are restored on the next open, but only while they still overlap a monitor.
3. **Independent of the main window.** A third webview, label `lyrics`, loaded with `?lyrics=1`.
4. **Font size, colour, outline, transparency.** They live in Settings → Appearance → Desktop lyrics, with a preview: text size, text colour, sung colour, outline, background opacity, and whether to show translations. Changes reach an open overlay live.
5. **Click-through.** Locking calls `set_ignore_cursor_events`.

## What it shows

The line being sung, with the progressive word fill on word-timed lyrics. Under it goes that line's translation, or else the next line. Unsynced lyrics cannot follow the song, so title and artist stand in, as they also do before the first line.

## Design points

- **The backend owns the window.** Three surfaces control it: the player bar's "⋯" menu, the Settings card, and two tray check items (show, lock). The tray has no webview, so the window cannot be created from JavaScript the way the mini-player is. All three call `open_desktop_lyrics` / `close_desktop_lyrics` / `set_desktop_lyrics_locked`, and all follow the `desktop-lyrics:state` broadcast rather than their own last click. The tray's check marks are also set from that state, because a check item toggles its own mark when clicked on some platforms.
- **Unlocking always happens from outside the window.** A locked window receives no mouse input, so each of the three surfaces offers unlock, and a new window always opens unlocked. For the same reason the "⋯" menu now always renders: in Spotify mode with everything pinned it used to be hidden, which would remove one of the ways to unlock.
- **Open and close are serialized.** `open` awaits the bounds read between checking for the window and building it, so open, close and the tray toggle decide under one lock. (Found in local review.)
- **Style sync across windows.** `useProfileSetting` broadcasts a window event, which stays inside its own document, while this setting is edited in the main window and read in the overlay. A write re-emits the notice as a Tauri event that carries the writer's label, so the writer does not re-read in the middle of a slider drag.
- **Transparent document.** `index.html` stamps `data-skin` in every window and the skins paint `body` from it. So `html.desktop-lyrics-window` forces the document transparent, and `main.tsx` adds that class before the first render.
- **Spotify.** `SpotifyContext`'s mini-only flag becomes `IS_SECONDARY_WINDOW` (new `lib/windowRole.ts`), so the Web Playback SDK stays out of the overlay as well.

## Platforms

- **macOS**: a transparent window needs the `macos-private-api` Tauri feature plus `app.macOSPrivateApi`. That rules out the Mac App Store, which WaveFlow does not ship through. `Cargo.lock` is unchanged.
- **Wayland**: the compositor may ignore always-on-top and the chosen position. The window still opens and drags. Settings shows a note when the backend detects a Wayland session; `GDK_BACKEND=x11` counts as X11.

## Checks

- `cargo check`, `cargo clippy -D warnings` (Windows, app crate, all targets), `cargo fmt --check`, `typecheck`, `lint`.
- 17 locales updated with identical key structure.
- `docs/features/ui.md` has a new Desktop lyrics section, and the tray, `CLAUDE.md` and `integrations.md` entries are updated.
- Local CodeRabbit review: 4 findings on the first commit, covering two issues, both fixed in the second commit, which the review then cleared.

**Not tested on screen yet**, on any OS. Transparency, click-through, drag and resize of an undecorated window are exactly the platform-dependent parts and still need a real run on Windows, macOS and Linux (X11 and Wayland).

Closes #582.

## Summary by CodeRabbit

  - Ajout d’une fenêtre flottante dédiée aux paroles de bureau, indépendante de la fenêtre principale.
  - Affichage des paroles synchronisées, des traductions et de la progression karaoké.
  - Personnalisation de la taille, des couleurs, du contour, de l’arrière-plan et de l’opacité.
  - Déplacement, redimensionnement, verrouillage avec clics traversants et mémorisation de la position.
  - Contrôle depuis les réglages, le menu d’actions du lecteur et la barre système.
  - Prise en charge de Wayland, avec indication de ses limitations.

- **Documentation**
  - Documentation mise à jour pour présenter la fenêtre de paroles et ses intégrations.


InstaZDLL:
@coderabbitai All three findings on fb991b71 are fixed:

- **Reopen while the old window is still closing** (511e9811, 8183a16c, f6a20936): `close()` only requests the close, so a close now marks the window as closing (from `close_locked` and from `WindowEvent::CloseRequested`, which covers Alt+F4), `Destroyed` clears the flag and wakes waiters through a `Notify`, and `open_locked` waits for it before deciding. The `Notified` is created before the flag is read, so a `Destroyed` landing in between is not missed. A failed `close()` drops the flag again, and an open that outlasts the 3 s wait now returns an error instead of showing the window that is on its way out.
- **Overlay buttons unreachable from the keyboard** (511e9811): they stay mounted whenever the window is unlocked and show on hover *or* focus-within (tracked on the root with `onFocus` / `onBlur` + `relatedTarget`). They are not mounted while locked, since a locked window takes no input at all.
- **Fieldset without an accessible name** (511e9811): visually hidden `legend` added, as suggested.

Also corrected in the docs: the tray is described as the Rust-only surface it is (`toggle_from_tray` / `toggle_lock_from_tray`, check marks set by `publish` → `tray::sync_desktop_lyrics`), separate from the two webview surfaces that use the hook.

Not applied from local review: `pointer-events-none` on the hidden chrome. The hover state is tracked on the window's root element, so the pointer cannot be over those buttons without the chrome already being visible; there is no click path onto them while they are transparent.

PR #672, opened 2026-09-17
fix: synced lyrics after an unsynced plugin answer, and a track table that fits its width

### LINKED ISSUE #668: fix(lyrics): an unsynced plugin answer stops the waterfall before synced providers
## Problem

In `fetch_lyrics` (`src-tauri/crates/app/src/commands/lyrics.rs`), `waveflow:metadata/v2` plugins run as tier 4, ahead of Musixmatch and LRCLIB. The first plugin answer is cached and returned at once, whether or not it is synced.

The Apple lyrics plugin returns TTML with `itunes:timing="None"` for songs Apple has no timing for. That untimed document ends the waterfall, even when LRCLIB or Musixmatch has synced lyrics for the same track. The user gets static text where karaoke was available.

It applies to any plugin that returns unsynced lyrics: an untimed TTML document, or `plain`.

## Why the plugin tier comes first

The comment at tier 4 explains it. A plugin is there because the user installed it, and it is the only tier that can return translations and a pronunciation with the lyrics (#585). A fix should keep that for synced answers.

## Direction

Keep an unsynced plugin answer as a fallback instead of returning it:

1. Run the plugin tier as today.
2. If its document is synced (LRC, Enhanced LRC, or TTML with at least one `begin`), cache it and return, as now.
3. Otherwise hold it, and keep going down the network tiers.
4. If one of them returns synced lyrics, use that. If none does, cache the held plugin answer so it still shows.

To check:
- The Rust side does not parse TTML (`document_is_what_it_claims` only sniffs). Telling timed from untimed TTML needs at least a check for `begin=` inside `<p`. `quick-xml` is already a dependency of the app crate.
- `refetch_lyrics` with a plugin chosen explicitly should still return that plugin's answer, synced or not.
- `lyrics.prefer_lrclib` and the local tiers must keep their current order.

## Context

Found alongside #667, which fixes the display side: untimed TTML showed as raw XML. That PR does not change which provider wins.

### LINKED ISSUE #669: feat(library): fit the track table to its width by hiding columns
## Problem

The library track table (Tracks, Folders, Tags) has a minimum width for every column. When the centre column is narrower than their sum, for example with the Now Playing or queue panel open or in a small window, the table no longer fits.

Since #667 the table widens as a whole and the page scrolls sideways. Before, it split: background and header stopped halfway along the rows. That fixes the broken look but not the fit: with a panel open you still scroll sideways to reach Rating or Length, and the heart and action buttons at the end of each row fall off-screen.

## Direction

Make the table fit the space it has, rather than overflow it:

- **Drop optional columns by priority** as the width shrinks. Title always stays; the thumbnail and the trailing heart and actions stay because they carry the row's interactions. The table's container width (not the viewport) decides, like the settings rows in #667.
- The columns come from the user's layout (#588: choose, reorder, resize). Hiding one for width must not change the saved layout. It only reappears once there is room.
- An alternative is to shrink resized columns back toward their minimum first. It helps, but the minimums alone already exceed the column with a panel open.

To check:
- `trackMinWidthFor` (`src/lib/trackColumns.ts`) already gives each column's floor, and the table's `min-width` is built from it in `LibraryView.tsx`. A container-width measurement could reuse it to pick which columns fit.
- Header and rows must drop the same columns: both use one `gridCols` string.
- Double-click-to-fit and keyboard resize on a column that is currently hidden.

InstaZDLL:
Two independent follow-ups from this week, grouped.

## #668: an unsynced plugin answer no longer ends the lyrics waterfall

Plugins (`waveflow:metadata/v2`) run ahead of Musixmatch and LRCLIB, and their first valid answer returned immediately. For songs Apple has no timing for, the Apple lyrics plugin returns TTML with `itunes:timing="None"`, so a track LRCLIB had synced lyrics for showed static text.

- **Order.** `fetch_lyrics` now returns a plugin answer at once only when it is synced. Otherwise it holds the answer and runs the tiers after it, split out as `fetch_after_plugins` (Musixmatch word-level, LRCLIB, the query-based chain, and the deferred local tiers under prefer-LRCLIB). Synced lyrics from those tiers win.
- **Fallback.** Anything short of synced — plain text, an instrumental verdict, a miss, a network error — leaves the plugin's answer standing. It is **cached again** from the kept bundle, because those tiers write their own row (a plain row, an empty miss row) over it. `upsert_lyrics` already clears associated documents on replace, so no stale translation survives in between.
- **What "synced" means** (`lyrics_are_synced`) mirrors the renderer, since an answer judged synced ends the waterfall:
  - LRC / Enhanced LRC: at least one complete line stamp, using `LRC_LINE_STAMP_RE`'s pattern.
  - TTML: a well-formed document with a `<p>` whose `begin` `parseTtmlTime` can read and that is not negative. It reads to the end and requires balanced elements, because quick-xml is lenient where `DOMParser` rejects a truncated document outright.

  This stricter rule came up in local review.
- **Explicit choice.** Picking the plugin by name in the provider picker (`refetch_lyrics` with its id) still returns its answer as-is.
- **Tests.** Added for untimed Apple-shaped TTML, span-only timing, prefixed elements, truncated documents, `begin` values the renderer rejects or accepts, and incomplete LRC stamps. They were checked against quick-xml 0.42 in a scratch crate (app-crate tests do not run on Windows); CI's ubuntu Rust job runs them here.
- **Docs.** The tier list in `docs/features/integrations.md` did not mention plugins at all; it now does, with this rule.

## #669: the track table hides columns that do not fit

With the right panel open or in a small window, the column floors added up to more than the view, so the table scrolled sideways and rating, length, heart and actions went off screen. Since #667 it no longer split in two, but it still overflowed.

- **Fitting.** The table measures its container (`ResizeObserver`), and `fitLayoutToWidth` drops columns **from the right end of the user's order** until the rest fit at their floors. The fixed chrome (index, thumbnail, heart, actions), gaps, padding and border are counted. The user's order is the priority, and the title is never dropped.
- **Render only.** The saved layout, the column picker and the stored widths are untouched, so a hidden column comes back in place and at its width once there is room.
- **Chrome sizes.** They are now defined once in rem, and the grid, its `min-width` floor and the fitting all read them, so the three cannot drift.
- **Docs.** `docs/features/library.md` is updated.

## Checks

`cargo fmt --check`, `cargo clippy -D warnings` (Windows, app crate, all targets), `typecheck`, `lint`, and a local CodeRabbit review with no findings on the final state. Not checked on screen yet.

Closes #668.
Closes #669.

## Summary by CodeRabbit

  - Prise en charge des plugins pour rechercher des paroles en parallèle des fournisseurs réseau.
  - Les réponses synchronisées sont privilégiées ; les réponses non synchronisées peuvent être conservées temporairement.
  - La sélection explicite d’un plugin permet de récupérer directement sa réponse.

- **Améliorations**
  - Le tableau de bibliothèque adapte automatiquement les colonnes à l’espace disponible.
  - Les colonnes excédentaires sont masquées depuis la droite sans modifier les préférences enregistrées.
  - Le titre reste toujours visible et le défilement horizontal est disponible si nécessaire.


InstaZDLL:
@coderabbitai Fixed in a130c618: `is_timed_line` now iterates the attributes explicitly and returns `None` on the first malformed one, which makes `ttml_has_timed_line` return false for the whole document. That matches the renderer, whose `DOMParser` rejects the document outright. The check applies to every element, not only `<p>`, since a duplicate attribute anywhere makes the document unparseable. A test covers a duplicated `begin` on a line and a duplicated attribute on a `<div>`.

PR #674, opened 2026-09-17
fix(lyrics): fit desktop lyrics lines, stop clipping descenders, and give the Wayland workaround

### LINKED ISSUE #673: fix(lyrics): desktop lyrics on Linux, not kept on top, and clipped text
Follow-up to #582 / #671, reported from a test on Fedora (Linux).

## 1. The window does not stay above other windows

`desktop_lyrics.rs` builds the window with `.always_on_top(true)`, but on this Fedora session other windows cover it.

This is most likely the Wayland limitation #671 already documents: a native Wayland client cannot hold "always on top" if the compositor declines, and GNOME's Mutter does not honour it for regular clients. Settings shows a note when `is_wayland_client()` detects the session (`XDG_SESSION_TYPE=wayland` or `WAYLAND_DISPLAY`, unless `GDK_BACKEND=x11`).

To confirm on the reporting machine:
- `echo $XDG_SESSION_TYPE` and the desktop (GNOME / KDE).
- Whether the Wayland note appears under Settings → Appearance → Desktop lyrics.
- Whether `GDK_BACKEND=x11 waveflow` (XWayland) keeps it on top. Mutter honours `_NET_WM_STATE_ABOVE` for X11 clients.

Possible directions, depending on the result:
- If XWayland works: document it as the workaround, or consider running the app under XWayland on GNOME when the overlay is used. `GDK_BACKEND` is process-wide, so it cannot be per-window; that is a trade-off to weigh, not a quick switch.
- KDE Plasma exposes window rules, and a layer-shell surface would be the proper Wayland answer, but Tauri/tao have no layer-shell support today.
- At minimum, make the Settings note say plainly that GNOME on Wayland will not keep it on top.

## 2. Text is clipped: descenders cut and long lines cut short

In the screenshot, the bottom of letters that go below the baseline (`g`, `y`, `p`) is cut off, and the current line ends in `…` even though the window is wide.

Causes, both in `src/components/views/DesktopLyrics.tsx` (lines 176 and 191):
- **Descenders.** Both lines are `truncate leading-tight`. `truncate` sets `overflow: hidden`, which clips to the line box, and `leading-tight` (line-height 1.25) makes that box shorter than the glyphs' descenders at large sizes. The text-shadow outline makes it worse, since it extends past the glyph too. Rendering differs by platform (font metrics on Fedora), which is likely why it was not visible elsewhere.
- **Ellipsis.** A line longer than the window at the chosen font size is truncated rather than fitted.

Direction:
- Give the lines room for descenders and the outline: a normal line-height, or vertical padding inside the clipped box sized to the outline.
- Fit long lines instead of cutting them. After the text changes, measure `scrollWidth` against `clientWidth` and scale that line's font size down to fit, with a floor (for example 60 % of the chosen size), so only extreme lines still truncate. It has to run for the current and the second line independently, and again when the window is resized.

InstaZDLL:
Two issues found testing the desktop lyrics overlay (#671) on Fedora.

## Clipped text

Both lines were `truncate leading-tight`. `truncate` clips at the element's box, and `leading-tight` made that box shorter than the glyphs, so the bottom of `g`, `y` and `p` was cut off. A line longer than the window ended in `…` at any size.

- **Line box.** A shared `LINE_CLASS` with `leading-snug` and `em` padding keeps descenders and the text-shadow outline inside the box at any font size.
- **Long lines.** A new `FitLine` shrinks a line too long for the window, down to 60 % of its size; only past that floor does `truncate` cut it. The natural width is measured at the chosen size, not the size last applied, so the scale cannot feed back on itself; the inline size is restored right after measuring. Refitting runs whenever the line's text changes and when the window is resized (`ResizeObserver`).
- **Local review.** It caught a 0.01 tolerance that left a slight overflow cut, and a refit key that ignored the line's text. Both are fixed.

## Not kept on top on Wayland

Confirmed on the reporting machine: the session is Wayland (GNOME), and `GDK_BACKEND=x11` keeps the window on top. Mutter ignores always-on-top for a native Wayland client.

- **Settings note.** It now says so and gives the workaround, `GDK_BACKEND=x11`, in all 17 locales.
- **Forcing XWayland from the app was ruled out.** `GDK_BACKEND` is process-wide, so it would affect every Linux user, overlay or not.

`docs/features/ui.md` is updated. `typecheck`, `lint`, and a local CodeRabbit review with no findings on the final state. Not checked on screen yet.

Closes #673.

## Summary by CodeRabbit

* **Améliorations**
  * Les paroles du bureau s’adaptent automatiquement à la taille de la fenêtre afin d’améliorer leur lisibilité.
  * Les lignes sont réduites jusqu’à 60 % si nécessaire, puis tronquées avec des points de suspension lorsqu’elles ne peuvent pas tenir entièrement.

* **Documentation**
  * Les limitations des fenêtres sous Wayland, notamment avec GNOME, sont précisées.
  * Une solution de contournement avec `GDK_BACKEND=x11` est désormais indiquée.


InstaZDLL:
@coderabbitai Fixed in cb386796: `FitLine` now refits when `document.fonts.ready` resolves and on every `loadingdone` event, so a first fit taken in the fallback face is corrected once the `@fontsource` font arrives. The listener is removed and a late `ready` ignored on cleanup, alongside disconnecting the observer.

Issue #682, opened 2026-09-18
nowPlaying.noBio sends the user to configure a key TheAudioDB does not need


InstaZDLL:
Splitting this out of the #670 round: it is a content bug, not a
translation one, and it needs one pass over all 17 files.

`nowPlaying.noBio` reads:

> No biography available. **Set your Last.fm key in the settings.**

But a Last.fm key is not required for biographies. The app ships two
sources and says so itself, two screens away:

- `settings.artistBio.lastfmHint` — *English biographies. Requires your
  Last.fm API key.*
- `settings.artistBio.theaudiodbHint` — *Community biographies in several
  languages. **No key required.***

So a user on TheAudioDB who sees the empty state is sent to configure a
key they do not need, and told nothing about the setting that would
actually help — `settings.artistBio.title`, **Biography source**.

The sentence should point at that setting rather than at one provider.
All 17 locales carry the same wording, so the fix is a values-only pass
across the set, and each locale already has `settings.artistBio.title`
translated to build the phrase from.

Found by CodeRabbit during the locale review; the English and the French
say the same thing, which is why no native reviewer flagged it as a
translation defect.

InstaZDLL:
Adding one more string to this pass, from the review on #683.

`settings.plugins.emptyHint` names a path the reader cannot use:

> Install one from the store above, or drop a folder under `<data>/waveflow/plugins/`.

#683 gives the empty state a button that opens the real directory, so the
path is now redundant rather than the only instruction. Removing it from
the text is what #679 asked for and what this pass is for.

It is in **all seventeen** files, and each phrases the clause around the
path in its own grammar — `ضَع مجلدًا في …`, `lege einen Ordner unter … ab`,
`… にフォルダを配置してください` — so it is prose, not a substitution.

The English should stop naming a path at all; the button below it is the
path. Something like *"Install one from the store above, or drop a folder
into the plugins folder:"* — final wording to settle with the same
reviewers, since sixteen locales follow it.

Issue #681, opened 2026-09-18
two counters render "1 covers retrieved", and HomeView carries dead french fallbacks


InstaZDLL:
A Dutch reviewer, reading `nl.json` for #670, noticed that some keys are
named in French. Chasing that turned up three separate problems, only one
of which is visible to users — but that one is visible to all of them.

## 1. Every language shows French under the Daily Mix cards

`src/components/views/HomeView.tsx:633`

```tsx
{t("home.dailyMix.trackCount", {
  defaultValue: "{{count}} morceaux",
  count: pl.track_count,
})}
```

**`home.dailyMix.trackCount` exists in 0 of the 17 locale files.** Not in
`en`, so `fallbackLng` cannot save it either. Every user, in every
language, reads `12 morceaux` under each Daily Mix card on the Home
screen.

No new key is needed: `queue.count_one` / `_other` already say
`{{count}} track` / `{{count}} tracks` and are translated everywhere.

Four other French `defaultValue` strings sit in the same file
(`HomeView.tsx:381`, `:384`, `:393`, `:400`, `:404`). Those keys *do*
exist in all 17 locales, so the French never renders — but it should not
be in the tree either.

## 2. Three counters read "1 covers retrieved"

These keys interpolate `{{count}}` with no plural forms anywhere,
including in `en` and `fr`:

- `library.fetchCoversResult` — `{{count}} covers retrieved`
- `settings.regenerateThumbnailsDone` — `{{count}} thumbnails regenerated`
- `settings.pruneAlbumCoversDone` — `Deleted {{files}} files · freed {{mb}} MB`

The first two need `_one` / `_other` in all 17 files. The third
interpolates `{{files}}`, not `{{count}}`, so i18next has nothing to
select on — it needs `count` passed at the call site before plurals can
work at all.

## 3. French identifiers in the library tab ids

`LibraryView.tsx:176-192` and `AppLayout.tsx:198` type the tab as
`"morceaux" | "artistes" | "dossiers" | "a-corriger"`, and those ids build
the i18n keys, so all 17 files carry `library.tabs.morceaux`,
`library.empty.a-corriger.title`, `library.header.subtext.artistes_one`
and so on.

Nothing is broken and no user sees it. It is renaming work across the TS
types and 17 files, so it is listed here rather than fixed in passing.

## Order

1 is a user-visible bug and stands alone. 2 is a source defect worth its
own pass over the 17 files. 3 is tidying and can wait for a quiet moment.

[...]

Issue #670, opened 2026-09-17
fix(locales): native-speaker review of the strings added since #438


InstaZDLL:
## Why

The last native-speaker pass over the locales was #438, on 2026-07-26. Since then `en.json` has gained **631 keys** (1,463 → 2,091 leaf strings). Every one of them was propagated to the other 16 locales in the PR that added it, written by the author rather than by a native speaker. None of those strings has been reviewed.

Key parity is fine: every locale has every `en` key, plural forms included. What needs checking is the wording.

## Scope

All 16 non-English locales: `fr`, `es`, `de`, `it`, `nl`, `pt`, `pt-BR`, `ru`, `tr`, `id`, `ja`, `ko`, `zh-CN`, `zh-TW`, `ar`, `hi`.

Priority goes to the new keys, grouped by feature (counts are new `en` keys):

| Area | Keys | Where it shows |
| --- | --- | --- |
| `remote.*` (reconciliation, server, catalogue, upload, import, playlist, tags, queue) | ~200 | Settings → Connections → Remote server |
| `settings.organization.*` | 53 | Settings navigation, category and group names, search (#666) |
| `smartRuleSummary.*`, `smartPlaylistEditor.*` | ~60 | Smart playlist editor and rule summaries |
| `library.inventory`, `library.columns`, `library.folderBrowser`, `library.header`, … | ~80 | Library: needs-attention view, column picker, folders |
| `deviceMenu.capabilities`, `playerBar.outputMode`, `player.alert` | ~32 | Output device menu, player alerts |
| `settings.replayGain`, `settings.contrast`, `settings.cacheLocation`, `settings.remoteTranscode`, `settings.canvasCache`, `settings.visualizer` | ~56 | Settings rows |
| `tasks.kinds`, `rendering.*`, `tagFetch.fields`, `lyrics.localization` | ~26 | Task bar, software-rendering banner, tag fetch, lyrics |

The older strings can be reviewed too, but #438 already covered them.

## What to look for

- **Terminology** consistent with the rest of each file: the same word for the same thing (library, track, profile, server, scan), and consistent with the OS vocabulary in that language.
- **Register**: each locale already has one (e.g. informal `tú` in es/pt, `kamu` in id, informal German). New strings should match it.
- **Strings left in English.** Excluding the brand tokens, each locale has some values identical to `en` (from ~36 in ja/ko/ar/hi to ~155 in fr, where many are cognates). Only the ones that should have been translated need fixing.
[...]

PR #683, opened 2026-09-18
fix(plugins): open the sideload folder instead of naming a path nobody can type

### LINKED ISSUE #679: the plugins empty state points at a path nobody can open
## What the empty plugins list tells you to do

`settings.plugins.emptyHint`, shown by [`PluginsCard`](src/components/views/settings/PluginsCard.tsx#L173) when no plugin is installed:

> Install one from the store above, or drop a folder under `<data>/waveflow/plugins/`.

`<data>` is never defined anywhere in the UI. It resolves to a different place on each platform:

| | |
|---|---|
| Windows | `%APPDATA%` |
| Linux | `~/.local/share` |
| macOS | `~/Library/Application Support` |

A reader who follows the sentence literally looks for a folder called `<data>`, or gives up. The string is identical in all seventeen locales, so nothing is lost in translation — the English is simply not actionable.

(The code's own comment writes it as `<app-data>`, in [`commands/plugins.rs:342`](src-tauri/crates/app/src/commands/plugins.rs#L342), which is closer but still not a path anyone can type.)

## Why this is not a copy fix

Rewriting the sentence to spell out three platforms would make it long and still leave the user to navigate there by hand. The app already knows the real path at runtime and already has the pieces to use it:

- `AppInfo.data_dir` carries the resolved root — [`commands/app_info.rs:30`](src-tauri/crates/app/src/commands/app_info.rs#L30)
- `open_data_folder` opens a path in the system file manager — [`commands/app_info.rs:37`](src-tauri/crates/app/src/commands/app_info.rs#L37)
- **Settings → Data folder** already pairs a description with an *Open* button, [`SettingsView.tsx:4272`](src/components/views/SettingsView.tsx#L4272)

So the plugins card can do what the data-folder row does.

## Suggested shape

Give the empty state the same treatment as the data-folder row: the sentence, then a button that opens the sideload directory. Two details worth deciding while implementing:

- **The folder may not exist yet.** Opening a missing path fails silently or errors depending on the platform, so it should be created on demand — which also removes the "where do I even put it" step entirely.
- The hint then no longer needs a path in its text at all, which drops one translated string's worth of platform detail from all seventeen locales.

## How this surfaced

Found while rewriting `settings.localArtistImages.subtitle` (commit `33e3d0be`), where three native reviewers in a row had each read `<name>.jpg` as a code token that should not have been translated. `<data>` is the same notation and the last one left in the locale files — but unlike `<name>.jpg`, nobody reported it, because it is consistent everywhere. Consistently unusable.

InstaZDLL:
Two small things, both found while closing out the #670 locale round.

## The plugins empty state named a path nobody can type — #679

> Install one from the store above, or drop a folder under `<data>/waveflow/plugins/`.

`<data>` is defined nowhere in the UI and resolves to `%APPDATA%`,
`~/.local/share` or `~/Library/Application Support` depending on the
platform. A reader who follows the sentence goes looking for a folder
called `<data>`. The string is identical in all seventeen locales — the
English is simply not actionable, which is why no native reviewer reported
it.

The card now does what **Settings → Data folder** already does: a button
that opens the real directory.

- `open_plugins_folder` creates the sideload root before opening it. A
  missing path makes the opener a silent no-op, and the boot-time
  `ensure_dirs` is not a guarantee — a user who deleted the folder, or an
  install whose creation failed on a permission error since repaired,
  would otherwise press the button and get nothing.
- The create runs on a blocking thread, like every other filesystem walk
  in that module.
- The failure path reaches the card's existing error line rather than
  being swallowed.

**No new translation was written.** Every locale already says *Open folder*
in `settings.diagnostics.openFolder`, for this same action on another card,
and each of those values was read by a native during #670.
`settings.plugins.openFolder` takes those words under its own key rather
than reaching into the diagnostics namespace.

The hint still names `<data>`. Dropping the path from seventeen files
belongs in the locale pass #681 and #682 already need — the button is what
makes the folder reachable, whatever the sentence says.

## Five French sentences in an English source file — part of #681

`HomeView.tsx` carried French i18next `defaultValue`s: *"Pas encore de
Daily Mix"*, *"Fig 1. — Sélection du jour."*, *"{{count}} morceaux"* and two
more.

I opened #681 claiming the last one rendered in every language. **That was
wrong, and I have corrected the issue.** `home.dailyMix.trackCount_one` and
`_other` exist in all seventeen locales — Arabic with all six categories,
Russian with its four — and those suffixed forms are what i18next resolves
when `count` is passed. I had checked for the bare key, which is of course
absent, because that is how plural keys look.

So none of the five can render, and `fallbackLng: "en"` means a key missing
from one locale falls back to English long before a `defaultValue` is
consulted. A `defaultValue` only fires when the key is absent from
`en.json` too — a build mistake that should be loud rather than quietly
French. They are removed.

## Checks

`typecheck` · `lint` · `cargo fmt --check` · `cargo check --workspace
--all-targets` all clean. Local review at zero findings; its two earlier
findings (the swallowed rejection, the blocking call on the async
executor) are the two fixes described above.

Closes #679

## Summary by CodeRabbit

  - Ajout d’un bouton **Ouvrir le dossier** dans les paramètres des plugins lorsqu’aucun plugin n’est installé.
  - Le dossier est créé automatiquement s’il n’existe pas, puis ouvert dans le gestionnaire de fichiers.
  - Les erreurs d’ouverture sont affichées dans l’interface.

- **Traductions**
  - Libellé disponible dans toutes les langues prises en charge.


InstaZDLL:
@coderabbitai Both verified against the tree. Skipping both here, with reasons.

**The loading state.** `handleOpenDataFolder` in
[`SettingsView.tsx:2057`](src/components/views/SettingsView.tsx#L2057) is
the same action on the data folder, and the button it drives at
[`:4281`](src/components/views/SettingsView.tsx#L4281) has no pending
state, no `disabled`, and does not surface the failure at all — it only
`console.error`s. The diagnostics card's *Open folder* is the same shape.
Adding a spinner to this one would make it the only open-folder button in
the app that behaves differently, and the worst outcome it prevents is two
file-manager windows. The version in this PR is already stricter than its
two siblings, since the rejection reaches the card's error line.

**The `<data>` path in `emptyHint`.** Deliberate, and stated in the commit
and the PR body — but your list is worth correcting first: the path is in
**all seventeen** locale files, not the eleven named. `id`, `it`, `ja`,
`ko`, `nl` and `pt-BR` carry it too.

Which is the reason for deferring. Removing it is not a find-and-replace:
every locale phrases the clause around the path in its own grammar
(`ضَع مجلدًا في …`, `lege einen Ordner unter … ab`, `… にフォルダを配置してください`),
so it means writing prose in seventeen languages, six of which I cannot
verify. #682 already needs a values pass over the same seventeen files and
now carries this string too, so it gets done there by someone who can read
the result.

This PR does not introduce the path — it makes it unnecessary. The button
resolves `<data>` for the reader, which is the part that was not
actionable.

PR #703, opened 2026-09-19
fix(player): stop a profile switch inheriting the previous profile's audio settings

### LINKED ISSUE #698: fix(player): switching profiles inherits the previous profile's visualizer, EQ, speed and crossfade
## What happens

Switching profiles carries some audio settings over from the previous profile.

Reported case: profile **Default** has the spectrum visualizer turned on, profile **ABTest** never touched it. Switch Default → ABTest: the visualizer is **on** in ABTest's immersive view. The other direction looks fine, because the profile you land on (Default) has a stored value.

## Why

On a profile change, `PlayerContext` calls `player_get_state` (`src/contexts/PlayerContext.tsx`, effect keyed on `activeProfile`), which re-reads the new profile's settings into the engine's process-wide atomics (`commands/player.rs`, `player_get_state`).

Half of those reads only write the atomic **when the profile has a row**:

```rust
if let Ok(Some(v)) = sqlx::query_scalar::<_, String>(
    "SELECT value FROM profile_setting WHERE key = 'ui.visualizer'",
)
.fetch_optional(&*pool)
.await
{
    engine.shared().visualizer_enabled.store(v == "true", Ordering::Release);
}
```

A profile that never changed the setting has no row, so the atomic keeps whatever the **previous** profile left in it. The other half of the function already does the right thing — `.unwrap_or(default)` and an unconditional store (`audio.dsd_dop`, `audio.match_source_rate`, `audio.pause_on_device_loss`, `audio.replaygain`, `audio.replaygain_prevent_clipping`, `audio.replaygain_mode`, `audio.dsd_precision`).

## Affected keys

Every read in `player_get_state` that uses the `if let Ok(Some(v))` form:

- `ui.visualizer` — the reported one
- `audio.normalize`
- `audio.mono`
- `audio.crossfade_ms`
- `audio.gapless`
- `audio.playback_speed`
- `audio.eq_enabled`, `audio.eq_bands`
- `audio.smart_crossfade`
- `audio.dynamic_crossfade`

So a fresh profile can also inherit the previous profile's EQ curve, playback speed, crossfade or mono — the visualizer is just the visible one.

## Fix direction

Store the engine's default when the row is missing, for each key above, matching the `.unwrap_or(...)` reads already in the function. The defaults must be the same ones the engine starts with (`audio/state.rs`) and the Settings UI shows for an unset key, or the fix trades one inconsistency for another. A DB error (`Err`) should probably keep the current value rather than reset it — decide per key.

Also check:

- `ImmersiveNowPlaying` reads `getVisualizerEnabled()` **once on mount** (`useEffect(…, [])`). If the view can stay mounted across a profile switch, its colour button would keep the old profile's state even once the backend is fixed.
- The frontend state for the same settings (EQ panel, speed, crossfade controls) — confirm it re-reads on switch and doesn't hold the previous profile's values either.

## To reproduce

1. Profile A: turn the spectrum visualizer on.
2. Create profile B (or use one that never changed the visualizer setting).
3. Switch A → B, open the immersive view: the visualizer is drawing.
4. Same with a non-flat EQ, a playback speed ≠ 1.0, or crossfade on in A.

InstaZDLL:
Closes #698.

Reported on the 1.8.0-beta.1: switching from a profile with the spectrum visualizer on to a profile that never enabled it left the visualizer running.

## The leak

`player_get_state` restores the active profile's settings into the engine's process-wide atomics, and it runs again on every profile switch. Ten of those reads were conditional on the row existing:

```rust
if let Ok(Some(v)) = sqlx::query_scalar::<_, String>(
    "SELECT value FROM profile_setting WHERE key = 'ui.visualizer'",
) … { store(v == "true") }
```

No row, no store — so the atomic kept the previous profile's value. A profile that never touched a setting inherited it: `ui.visualizer`, `audio.eq_enabled`, `audio.eq_bands`, `audio.playback_speed`, `audio.crossfade_ms`, `audio.smart_crossfade`, `audio.dynamic_crossfade`, `audio.normalize`, `audio.mono`, `audio.gapless`.

Seven other reads in the same function already resolved to a default and stored unconditionally — `audio.dsd_dop` even carries the comment explaining why. This makes the remaining ten match.

## What changed

- **`read_bool_setting(pool, key, default)`** — the shape the four hand-rolled copies already had, now shared by them and by the six booleans that were missing it.
- **The EQ curve needed more than a default.** `set_all_bands_db` writes only the bands its slice covers, so a short or malformed row left the bands it did not reach holding the previous profile's gains. The stored row is padded into a full flat curve before it is applied.
- **`player_get_visualizer` read the atomic**, which only `player_get_state` refreshes — and nothing orders those two calls against each other after a switch, so the view could read the old value even with the restore fixed. It reads the profile's row now, and falls back to the atomic while no profile pool is leased.
- **Both frontend readers of that toggle were mount-only.** Settings re-reads on the active profile, with a `touched` ref so a read already in flight cannot land on top of a switch the user just flipped. The immersive view tags its answer with the profile it describes — the id gate `useArtistImage` uses — rather than resetting to `false` and blinking the colour button away.

## What I did not change

**Playback speed keeps its raw atomic store** instead of `set_playback_speed`. The review suggested rebasing through the setter when a stream is live; `switch_profile` stops playback before swapping the pool, so the engine is idle at this point, and `player_set_speed` persists a row on every change — no row means the engine is already at 1.0. The setter would also rebase `samples_played` against a position not yet loaded, which is what the existing comment warns about.

## Checks

`cargo fmt --check`, `cargo check`, `cargo clippy --all-targets` (app crate, Windows) and `typecheck` / `lint` all clean. The new test does not run locally — the app crate's tests hit `STATUS_ENTRYPOINT_NOT_FOUND` on this machine — so the Ubuntu Rust job is the judge.

### Test

`an_unset_toggle_reads_as_its_default`: against the real migrated profile schema, an unset key reads as its default in both directions (off for the visualizer, on for gapless), an explicit `true` row reads as on, and a row holding something else (`yes`) reads as the default rather than as on.

### By hand

Profile A with the visualizer on, EQ curved and speed at 1.5, profile B untouched: switching A → B now lands on a flat EQ, 1.0 and no visualizer, and the Settings switch and the immersive view agree.

## Summary by CodeRabbit

* **Améliorations**
  * La restauration des réglages audio respecte désormais le profil actif.
  * Les préférences absentes ou invalides utilisent des valeurs par défaut fiables.
  * La vitesse de lecture est limitée à une plage valide et l’égaliseur est réinitialisé si nécessaire.
  * Les options booléennes sont gérées individuellement pour chaque profil.

* **Correctifs**
  * Le visualiseur et les options de fondu se resynchronisent lors d’un changement de profil.
  * Les réponses obsolètes ne peuvent plus écraser les préférences actuelles.
  * Les réglages audio sont correctement appliqués après le changement de profil.


InstaZDLL:
@coderabbitai I pushed 0d5dad83 with the review applied, plus a root cause the findings pointed at.

**Moved the restore instead of patching the reader.** Every getter (`player_get_visualizer`, `player_get_smart_crossfade`, `player_get_speed`, `player_get_audio_settings`) reads the engine's process-wide atomics, and those were only refreshed by whichever `player_get_state` the frontend made next — nothing orders that against a view re-reading its own setting. The block is now `restore_profile_audio_settings`, called from `switch_profile` as the pool is swapped and from startup, as well as from `player_get_state`. The `switch_profile` call pins the pool to the profile it is switching to (`require_profile_pool_for`), which covers the serialization concern without a new mutex: a first switch's restore cannot land after a second's.

**Boolean parsing.** Right that a non-boolean value should read as the default, and there was more to it: `"1"` / `"0"` is the older spelling, still on disk and still accepted by seven other readers (`plugins.rs`, `canvas.rs`, `profile_io.rs`, …). Reading `"1"` as false would have turned settings off. Both spellings are accepted now, anything else is the default, and the test covers all three cases. The pause-on-device-loss block uses the helper too.

**Frontend.** The three per-profile toggles hold their value tagged with the profile that answered, are disabled until the active profile has answered, and roll a failed write back only while that profile is still active. The null-state case you flagged is explicit, so no active profile no longer reads as hydrated.

**Two I did not apply:**

- *Propagating SQL errors out of `read_bool_setting`.* The whole function swallows read errors by design — the seven pre-existing reads around it already resolve to a default on `Err`, and `player_get_state` is the frontend's hydration path, so failing it would leave the player without a snapshot rather than without one setting. sqlx's busy timeout is 5 s, so a read that fails here is not transient contention.
- *Passing the profile id into `player_set_visualizer`.* No `player_set_*` command pins its write today; pinning one of twenty would not establish the invariant. Worth its own change across the family, and I'd rather not smuggle it into this fix.

Local review is at zero findings on the branch.

InstaZDLL:
@coderabbitai d91663cc — I took half of this one.

**Logging: yes.** A failed query was indistinguishable from a missing row, both swallowed by `.ok().flatten()`. Every read in the restore now goes through `read_setting`, which logs the key and the SQL error before the caller falls back to the setting's default. That is the part that was genuinely missing: silence made the failure undiagnosable.

**Leaving the atomic unchanged on an error: no.** That is the leak this PR is about. The atomic holds the *previous profile's* value at that point, so skipping the store on a failed read means a profile whose database cannot be read inherits the settings of the profile before it — EQ curve, speed, visualizer. Neutral defaults plus a warning is the safer failure, and it keeps every read in the function to one rule.

Two more from the same pass, both fair:
- **Volume** had the same conditional shape as the ten and is just as profile-scoped. It moved into the restore and resolves to `1.0` — what `SharedPlayback::new` starts from — instead of keeping the previous profile's level.
- **Startup** no longer reads `audio.pause_on_device_loss` by hand; the restore that runs there now covers it, so the duplicate is gone.

Also dropped a duplicated UPDATE in the test. Local review is at zero.

InstaZDLL:
@coderabbitai 061770a0.

**Volume logging: applied.** `queue::read_player_volume` swallowed its own query error through `.await.ok()??`, so a failed read fell back to 1.0 silently — you're right that it was the one setting left outside the rule. Its only caller was this restore and the row is written next to it, so the read moved in through `read_setting` and the helper is gone.

**The three toggles: extracted.** `useProfileScopedToggle` now carries the tagged value, the profile-keyed read, the touched guard, the disabled-until-hydrated state and the profile-checked rollback once instead of three times; the call sites are five lines each. Its refs are synced from an effect, not during render, which the linter rejects.

**Two I am not taking:**

- *A mutex around `activate_profile` + the restore.* Already serialized, by construction. The restore holds a leased `ProfilePool`; a second `switch_profile` swaps the guard and then awaits `previous.close_when_idle()`, which cannot complete while that lease is outstanding — so the second switch's restore starts strictly after the first one's has finished. The `require_profile_pool_for` pin covers the other direction: a restore whose profile is no longer active never gets a pool. A mutex would add a lock order to a path that already has one.
- *Serializing writes in the hook.* Every toggle in Settings is fire-and-forget today, and each write carries the whole value rather than a delta, so an out-of-order pair converges on the next hydration. Chaining three of about twenty would not make the guarantee real, and it is not what #698 is about.

Local review is at zero on the branch.

InstaZDLL:
@coderabbitai 0c4e1979.

**The rollback fix is right and I took it.** A write that failed after the user had clicked again reverted on top of the newer decision. It now only reverts while its own value is still the one on screen, which subsumes the profile check it replaces — a switch has already re-tagged the value. I also added the guard at the top of `toggle`: the control is disabled until the active profile has answered, but a caller that renders it enabled anyway should not write from a profile we left.

**Two findings in this round describe code that is not there.** Both flagged a duplicated trailing `useCallback` / leftover `setState` in `useProfileScopedToggle.ts` (once as critical), and one restated the guard and `activeProfileId` capture as missing. The file is 105 lines, the `toggle` callback appears once, and `tsc` / `eslint` are clean — a duplicated closure there would not parse. Flagging it for calibration rather than as a complaint.

**Not taking, with reasons:**

- *Dropping the restore from `player_get_state`.* It is the net that covers any path that activates a profile without going through `switch_profile`; the reads are primary-key lookups on an open pooled connection.
- *One batched `SELECT … WHERE key IN (…)` instead of the per-key reads.* Same reason — the cost is not where the suggestion assumes, and a key/value map would trade the per-setting comments, which are the part worth keeping.
- *Reusing the hook in `ImmersiveNowPlaying`.* That view only reads the toggle; it would have to pass a writer it never calls, and the hook exists for the write/rollback lifecycle.

Local review is back to zero on the parts that exist.

InstaZDLL:
@coderabbitai b9d91798 — applied, and you were right that the value comparison left a hole.

Three clicks bring the switch back to what the failed write was setting, so `cur.on === next` reads as "still mine" when it is not, and the rollback undoes the click the user actually ended on. Each click and each profile change now bumps a revision, the write captures it, and a failure from an older revision is dropped. The profile check stays for the tagging, but the revision is what decides.

Local review is at zero on the branch.

Issue #693, opened 2026-09-19
bug: the artist backdrop is low quality, crops faces, and cannot be changed


InstaZDLL:
## What happens

The full-bleed backdrop behind the artist page header (#482) has three problems:

1. **It is often low quality** — soft, visibly upscaled.
2. **It is badly framed, and the framing changes with the window size.** Faces get cut: in the screenshot the head sits against the top edge.
3. **It cannot be changed.** The photo has a picker; the backdrop has nothing.

## Why — what the code does today

`ArtistHeroBackdrop.tsx` paints one of two tiers, chosen by the caller:

| Tier | Source | Treatment |
| --- | --- | --- |
| Real fanart | TheAudioDB, cached in `metadata_artwork/` | `blur(2px)`, `scale(1.06)` |
| Fallback | the **square** artist photo | `blur(56px)`, `saturate(190%)`, `scale(1.35)` |

Both are painted with `bg-center bg-cover`.

**Framing.** `bg-cover` fills a very wide, fairly short banner, so the image is cropped top and bottom — and `bg-center` crops **equally** from both. Faces are nearly always in the upper third of a photo, so the crop takes the head first. As the window widens, the banner gets relatively shorter and the crop deeper, which is why the result shifts with the window size.

**Quality.** Two measured causes, possibly more:

- The TheAudioDB payload is read in order of preference: `strArtistFanart` (1920×1080) first, but when it is missing the code falls back to **`strArtistWideThumb` and `strArtistBanner`, which are 1000×185 strips** — stretched across a banner that is often wider than 1000 px, and much taller than 185 px relative to its width.
- The square-photo fallback is, by design, a blurred colour field: a 1:1 image upscaled 1.35× behind a 3:1+ banner. It is meant to read as ambient colour, but on an artist with no fanart it is the whole hero.

**No manual choice**, although the data is already there: TheAudioDB returns up to **four** wide backdrops (`strArtistFanart`, `Fanart2`, `Fanart3`, `Fanart4`), of which the code uses the first it finds.

## Proposal

- **Anchor the crop high**, not in the middle — `background-position` around the upper third rather than `center` — so a shrinking banner loses sky and floor before it loses a face. A stored focal point per artist would be the fuller answer.
- **Don't stretch a 1000×185 strip across a full-width hero.** Below a width threshold, prefer the blurred colour-field tier (which is honest about being ambient) over a sharp-but-pixelated banner.
[...]

Issue #692, opened 2026-09-19
bug: picking another Deezer artist in the image picker changes the photo, not the artist


InstaZDLL:
## What happens

On an artist page, **Change artist image → Search Deezer** lists every Deezer artist matching the name. Picking a different one than the automatic match — here the real **NAYEON** (63K fans) instead of the "Nayeon" with 7 fans that the automatic lookup had latched onto — changes the round photo in the header, and nothing else.

After the pick:

- the header still reads **7 fans**, the wrong artist's count;
- the **About the artist** block in the Now Playing panel still shows the other artist's photo (the blue sphere in the screenshot);
- **Similar artists** are still the other artist's — the first tile in the screenshot looks like Madcon, which sits right next to the 7-fan Nayeon in the Deezer search results.

## Why

The picker changes the picture, not the artist.

`set_artist_artwork_from_deezer` (`commands/deezer.rs`) downloads the chosen artist's photo and writes one column:

```sql
UPDATE artist SET artwork_id = ? WHERE id = ?
```

It never touches `artist.deezer_id`. And `artist.deezer_id` is the key everything else resolves through:

| Surface | Reads |
| --- | --- |
| Fan count, Now Playing photo, bio, hero backdrop | `app.metadata_artist` **via `artist.deezer_id`** (`WHERE deezer_id = ?`) |
| Similar artists | `resolve_similar_by_name(…, source_deezer_id)` with the same id (`commands/similar.rs`) |

So the user explicitly tells us which Deezer artist this is, we use the answer for one image, and every other surface keeps asking about the artist the automatic match got wrong.

It also emits nothing. The artist page reloads its own photo through `onSuccess`, but the Now Playing panel caches its enrichment per `artist_id` and has no reason to look again.

## Expected

Choosing a Deezer artist in the picker **re-links** the local artist to that Deezer id: photo, fan count, similar artists and the enrichment cache all follow, on every surface, without a restart.

## Things to settle

- **Re-link, or only re-picture?** The two are different intents. The photo tab is the natural place to fix a wrong match, and it is the only place a user can see the candidates side by side with their fan counts — but some users will want a different photo of the *right* artist. Re-linking on a Deezer pick, and leaving **Local file** as "photo only", keeps both.
- **A manual link must survive the next enrichment pass.** The background sweep that fills `metadata_artist` must not re-run the name search and put the wrong artist back. It needs a marker that the id was chosen by hand — the same shape as the other user overrides.
[...]

PR #704, opened 2026-09-20
fix(artist): local images, Deezer re-link, and a pickable banner

### LINKED ISSUE #701: fix(slideshow): the cover slideshow ignores local artist images and only uses Deezer
## What happens

The cover slideshow (#466) alternates the album cover with the artist's photo — but the photo it uses always comes from Deezer. A library with **local artist images** gets the worse of two outcomes:

- when Deezer has a photo, the slideshow shows Deezer's instead of the local one, which is often lower quality than what the user curated;
- when Deezer has none, the slideshow has nothing to alternate with and falls back to the static cover — even though the artist page right next to it is showing a local photo.

Reported by a beta tester on 1.8.0-beta.1 (#700), who has a local image for every artist.

## Why

[`useArtistImage`](../../src/hooks/useArtistImage.ts) — the hook that feeds the slideshow — resolves the photo from `enrichArtistDeezer` alone: `picture_path` / `picture_path_1x` / `picture_path_2x` / `picture_url`, all of them Deezer-enrichment fields cached under `metadata_artwork/`.

The local artist image lives somewhere else entirely: the scanner links an `artist.jpg` sidecar onto `artist.artwork_id` (`commands/scan.rs::extract_artist_image`, see library.md), and that is what `ArtistDetailView` renders through `artwork_hash`. The hook never looks at it.

So the hook's own doc comment — "local `artist.jpg` sidecar or the Deezer-enriched picture" — describes an ordering the code does not implement. Same family as #350, where the similar-artists path also resolved Deezer-only and ignored the local sidecar.

## Fix direction

In `useArtistImage`, resolve in this order:

1. the artist's local artwork (`artist.artwork_hash`), when the row has one;
2. the Deezer enrichment, as today.

Points to settle while doing it:

- **Where the local hash comes from.** The hook only receives an `artistId`; the current track already carries artist ids, but not the artwork hash. Either a small command that returns the artist's artwork fields, or extend what enrichment returns so one round-trip covers both.
- **Skip the network when a local image exists.** Today the hook always calls `enrichArtistDeezer`. With a local hit, the enrichment fetch is pointless for the slideshow (it may still be wanted for the bio elsewhere — check before removing it).
- **Resolution.** The slideshow asks for `"full"` because it paints into the large cover slot; the local artwork must be resolved at the same size, not a thumbnail.
- **The other surfaces.** `NowPlayingPanel` resolves the artist photo inline on the same enrichment path — it needs the same ordering, or the panel and the immersive view will disagree about which photo the artist has.
- **An explicit override** (#692 / #693 territory) should still win over both.

## Related

- #466 cover slideshow
- #350 — the same Deezer-only resolution, fixed in two other call sites
- #693 — the artist backdrop, which has its own sourcing problem

InstaZDLL:
Closes #701, #692, #693 — the three artist-image issues, which all turned out to be the same seam: an artist's identity and an artist's pictures were resolved by paths that disagreed with each other, and nothing told anyone when one of them changed.

## #701 — the curated image was ignored

The artist page and the library grid show `artist.artwork_id`: an `artist.jpg` sidecar the scanner linked, or a picture set by hand. Everything reached through `enrich_artist_deezer` showed Deezer's picture instead — the now-playing panel and the cover slideshow. A user with local images saw Deezer's photo where the page beside it showed theirs, and saw *nothing* where Deezer has no photo, while the page showed the local one.

The enrichment carries the local image beside the Deezer one now (`artwork_path*`), and both consumers prefer it — per source rather than per size, so a Deezer thumbnail can never fill in for a local original.

## #692 — the picker changed the photo, not the artist

It lists every Deezer artist matching the name *with its fan count*, so it is the one place a user can say which artist this is. It wrote `artist.artwork_id` alone, and `artist.deezer_id` is what the fan count, the bio, the fanart and similar artists all join on: the page ended up showing one artist's photo above another artist's numbers. Photo and id are written in one statement now. The hand-picked id sticks, because the automatic link only fills that column when it is `NULL`. **Picking a local file deliberately does not re-link** — a file on disk says nothing about which Deezer artist this is.

Nothing was told, either. The three picker commands emit **`artist:updated`**; `useArtistUpdated` is the listener side, the sibling of `useTrackUpdated`. Before it, only the page the picker was opened from refreshed — a photo is resolved once per artist id and cached per artist id, so while the same track kept playing nothing else would look again. The library grid takes it on its own refetch counter: a new artist photo says nothing about tracks, albums, genres or folders.

## #693 — the banner

Three causes.

- **It was often not a backdrop at all.** With no fanart, the code fell back to TheAudioDB's wide thumb and banner — 1000×185 strips by its own spec, stretched across a header routinely wider than 1000 px. They are not deserialized any more; an artist with no fanart falls through to the blurred-photo tier, which is honest about being an ambient colour field.
- **It cropped the face off.** `bg-cover` crops a wide, short banner top and bottom, and `bg-center` crops *equally* from both — so the head goes first, and the framing visibly shifted as the window resized (a wider window makes the banner relatively shorter, so the crop deepens). The fanart tier is anchored at 30 % now; the blurred tier keeps `center`, having no subject left to protect.
- **It could not be changed**, although TheAudioDB returns up to four fanarts and we kept the first. All four are cached (`metadata_artist.background_urls`), and a picker shows them in the hero's own 16:6 proportions, cropped the way the hero crops — so what the thumbnail cuts off is what the header will cut off — plus a local file, plus "back to automatic".

The choice lands on `artist.background_artwork_id`, in the profile artwork dir where a metadata-cache prune cannot reach it, and it wins over the cached fanart in **both** paths that resolve the hero: `get_artist_detail` paints the first frame and `enrich_artist_deezer` refines it a moment later, so resolving it in one would let the automatic one win the race.

`set_artist_background_from_url` accepts only a URL from that artist's cached candidate list. The picker never passes anything else, but a command is an API, and without the check it fetches any URL the caller names from inside the user's network.

## Migrations

- `app/20260920120000_metadata_artist_background_candidates.sql` — `background_urls` (JSON list), NULL meaning "not looked up since", backfilled by the next enrichment.
- `profile/20260920120000_artist_background_artwork.sql` — `artist.background_artwork_id`, `ON DELETE SET NULL` so a future artwork GC cannot take the artist row with it.

Both are `ADD COLUMN` on existing tables.

## Checks

`cargo fmt --check`, `cargo check --workspace --all-targets`, `cargo clippy --workspace --all-targets`, `typecheck`, `lint` — all clean. Seventeen locales carry the six new `artistBackdropPicker.*` keys, key parity verified after folding plural suffixes (1957 logical keys). Local CodeRabbit review at zero across four passes.

### Tests

`theaudiodb`'s two fanart tests were rewritten for the new shape: `fanart_urls_keep_every_image_in_order` (blanks dropped, order kept) and `a_strip_is_not_a_backdrop`, which decodes a payload carrying *only* a wide thumb and a banner and asserts the result is empty — the regression that made the backdrop blurry in the first place.

### By hand

An artist with a local `artist.jpg` and no Deezer photo: slideshow and now-playing panel show the local one. Picking the right Deezer artist in the picker: fan count, bio and similar artists follow, and the now-playing panel changes without a track change. Backdrop picker on an artist with four fanarts: each is offered, the chosen one survives a page reload, and "back to automatic" returns to the first.

## Summary by CodeRabbit

  - Personnalisez la bannière d’un artiste local avec des suggestions de fanarts ou une image locale.
  - Revenez à la bannière automatique à tout moment.
  - Le sélecteur de bannière propose des états hors ligne et est disponible dans plusieurs langues.

- **Améliorations**
  - Les images locales et personnalisées sont prioritaires sur les images distantes.
  - Les vues se mettent automatiquement à jour après une modification d’image.
  - Les couvertures et héros utilisent l’image d’artiste la plus pertinente disponible.


InstaZDLL:
@coderabbitai 42d4ef79 — the cache-erasure finding was the best catch of the round, and it needed a sharper fix than COALESCE.

**Preserving the background columns.** You are right that writing `excluded` unconditionally let a failed lookup erase a good cached fanart for every profile. But COALESCE would keep a backdrop TheAudioDB has *removed* for ever, since "no fanart" and "could not ask" both arrive as NULL. The three columns are now written only when the API answered, which `background_fetched_at` already records — so an empty answer still wins, and a transport error changes nothing.

**Also applied:**

- `download_image_bytes` refuses a URL naming a host on our side of the network — loopback, link-local, the private ranges — and applies the same rule to each redirect hop. Covered by a unit test over the ranges (it does not run on this Windows box; the Ubuntu job is the judge).
- `set_artist_background_from_file` checks the file size before reading it.
- The backdrop button no longer requires an existing hero: an artist with no image at all is precisely the one who needs to pick one.
- The picker carries offline-ness in the same tagged answer as the candidate list, so the two cannot come from different cycles.
- The now-playing panel blanks its photo and bio only when the artist really changes.

**Not applied, with reasons:**

- *DNS rebinding.* Defending it properly means a custom resolver that pins and re-checks every resolved address at connect time. The threat needs a compromised Deezer or TheAudioDB *and* attacker-controlled DNS, and the literal-address and per-hop checks already close the reachable part. I would rather not carry a bespoke resolver for it.
- *A `ready` flag on `useArtistUpdated`.* Its sibling `useTrackUpdated` has none, and the consumers refetch on their own key anyway; an event missed in the millisecond before `listen()` resolves would be one emitted by a command the same user just invoked from another window.
- *A test that the command rejects a URL outside the candidate list.* The check is one `any()` over a list read two lines above; a test of it would restate the line. The URL predicate, which does have edge cases, is the one under test.

Local review is at zero on the parts I kept.

InstaZDLL:
@coderabbitai bb621339 — two of the three applied, and the first was a real defect my own earlier change introduced.

**The removed picture.** Right: once the panel stopped blanking on a same-artist refresh, `if (resolved) setPictureSrc(resolved)` meant a picture the user had just deleted stayed on screen for as long as the track played. A resolution that *succeeded* is the answer now, including when the answer is nothing; the `catch` still keeps what is on screen, since a failed call is a network problem rather than an answer. The bio follows the same rule — after a re-link it belongs to the artist we stopped pointing at.

**One call for the picker.** Done: `get_artist_backdrop_candidates` returns `{ urls, offline }`. Two calls could straddle a change of the setting and pair an empty list with "we are online", which reads as "this artist has no fanart".

**DNS rebinding — still not taking it**, and this is the second time it comes up, so let me be concrete about why rather than repeat myself. Doing it properly means resolving every hostname ourselves, rejecting on any non-public address, pinning the chosen address into the client with `resolve()`, and repeating all of it per redirect hop — a bespoke resolver in the artwork downloader. What it buys: protection in the case where TheAudioDB or Deezer is already compromised *and* the attacker controls DNS responses, against a desktop player fetching a JPEG. The literal-address check and the per-hop check already close the part an attacker gets for free. I would rather leave the remaining risk visible in the PR than carry a resolver that has to stay correct on three platforms.

Local review is at zero.

InstaZDLL:
@coderabbitai applied — good catch on the reopen path. Tagging by artist covered a *different* artist, not a second visit to the same one: the previous candidates painted while the new call was in flight, and since they are remote URLs, enabling offline mode between two visits would still have fetched them. The list is dropped in the effect's cleanup now, which keeps it out of the render pass.

On DNS rebinding, which is the only thing left: my answer is unchanged and I have given the reasoning twice, so I will let it stand as a known, documented limit of the artwork downloader rather than build a pinning resolver here.

Issue #695, opened 2026-09-19
feat: keep lyrics, Canvas clips and animated covers next to the music


InstaZDLL:
## The request

Let the user choose, for **lyrics**, **Canvas clips** and **animated covers**, whether WaveFlow keeps them in its **own data folder** (today) or **next to the music, in the library folder** — so they travel with the files, survive a reinstall, and are there for any other player or device that reads the same folder.

## Where things live today

**Lyrics are split in two, and only one half has a choice.** The editor already offers three destinations — embedded tag, **sidecar `.lrc`/`.txt` next to the audio file**, or WaveFlow database only — chosen during onboarding and in **Settings → Images and lyrics**. But that choice applies to lyrics **you edit**; the onboarding says so ("When you edit lyrics, where should they go?"). Lyrics **fetched automatically** from LRCLIB and the other providers always go into `app.lyrics`, in the app database, whatever the setting says.

So the concrete gap for lyrics is: **let fetched lyrics follow the destination the user already picked**, and optionally write sidecars into a dedicated subfolder rather than beside every file.

**Canvas and animated covers have no choice at all.** Per #619's inventory:

| Path | Holds |
| --- | --- |
| `canvas_cache/` | Canvas clips (`.mp4`, LRU-evicted) |
| `motion_cache/` | animated album covers (`.mp4`, LRU-evicted) |
| `profiles/<id>/canvas/`, `profiles/<id>/motion/` | manual ones, never evicted |

#619 made the cache **root** movable — useful for a full system drive — but it is still one app-side location, not per-library.

## A tree to decide on

Something like this, per library root:

```text
<library root>/
  .waveflow/
    lyrics/<relative path of the track>.lrc
    canvas/<relative path of the track>.mp4
    motion/<album artist>/<album>.mp4
```

The two real choices are **one hidden folder per library root** (above) versus **files beside each track** (`Song.lrc`, the convention other players read). Beside-the-track is what makes lyrics portable to other software; a single folder is tidier and far easier to exclude from scanning. They may deserve different answers — lyrics beside the track, videos in the hidden folder.

## Constraints that shape the design

- **The scanner would index the videos.** `scanner/extract.rs` accepts `mp4` as an **audio** extension. A Canvas or motion cover written into the library tree would be picked up as a track on the next scan. Whatever folder holds them must be excluded from scanning explicitly — and the watcher must not treat our own writes as library changes and trigger rescans.
[...]

PR #706, opened 2026-09-20
feat: fetched lyrics follow your destination, and a window-frame choice

### LINKED ISSUE #696: feat: a native-looking mode on Linux and macOS
## The request

Chrome on Linux has an appearance choice between **GTK**, **Qt** and its own **classic** look, so the browser can blend into whichever desktop it runs on. WaveFlow would benefit from the same kind of option on Linux, and a native-looking mode on macOS.

## What "native" can and cannot mean here

This needs framing first, because the Chrome option does not translate one-to-one.

Chrome draws its own window chrome and paints it from the GTK or Qt theme. WaveFlow's interface is a **web view**: everything inside the window is ours already, and is themed by our skins. What the desktop owns is the **frame around it** and the **system dialogs** it opens.

And on Linux that frame is GTK, not by choice: wry, the web view layer under Tauri, is bound to WebKitGTK. **There is no Qt backend**, so a "Qt mode" cannot mean Qt widgets. On KDE Plasma it has to mean *looking right under Plasma* — which is reachable through the levers below.

## The levers that exist

**1. File dialogs — the most concrete win.** `rfd` is compiled with the **`gtk3`** feature (`crates/app/Cargo.toml`), not **`xdg-portal`**. So on KDE Plasma, *Add folder* and every other file picker opens the **GTK3** chooser, not the KDE one. With the portal backend, the desktop supplies its own dialog: KDE's on Plasma, the GTK4 / libadwaita one on current GNOME. That is exactly the "Qt on Plasma" behaviour, obtained without any Qt code.

   Cost to check: the `Cargo.toml` comment pins `rfd`'s features to match `tauri-plugin-dialog` so Cargo unifies them into one copy, with no new crate and **no Flatpak source regeneration**. Switching to `xdg-portal` breaks that on purpose — it brings `ashpd`, and the Flatpak sources must be regenerated. Flatpak itself strongly prefers portals, so it is the right direction there anyway.

**2. Window decorations.** The main window uses **native decorations** (only the splash sets `decorations: false`), which on Linux means GTK3 client-side decorations. `docs/upstream-blockers.md` (B1) already records that these look dated next to libadwaita apps on GNOME 49+, and why: wry cannot use WebKitGTK 6.0 / GTK4 yet. An option could offer:
   - **System** — today's native frame, which KWin can draw server-side on Plasma;
   - **WaveFlow** — `decorations: false` and a titlebar drawn in-app, the route B1 already names as "what we can do today".

**3. System accent and colour scheme.** A "System" theme entry that follows the desktop's accent colour and light/dark preference (the XDG settings portal on Linux, `NSColor.controlAccentColor` on macOS), rather than one of our fixed presets.

**4. macOS.** A native-feeling mode there is mostly the titlebar: the traffic lights inset into our top bar (`titleBarStyle: "Overlay"`, hidden title) instead of a separate title strip, and vibrancy behind the sidebar where the skin allows it.

## Suggested scope

(1) is small, self-contained and fixes something users see on every KDE install. (2) and (4) are the substantive "native look" work and share one component, the in-app titlebar. (3) is independent and can land on its own. Worth taking in that order rather than as one switch.

InstaZDLL:
Two issues that both answer "make WaveFlow fit in where it runs" — the lyrics half of #695, and the window frame of #696.

## Fetched lyrics follow the destination you chose (#695)

The destination picked in onboarding and in Settings governed only the lyrics you **typed**. Anything LRCLIB, Musixmatch or a plugin returned stayed in `app.lyrics`, so a library deliberately set to "sidecar file" still kept everything fetched locked inside WaveFlow's database. Now it travels with the music.

Four refusals shape it, each written at the site:

- **Only what came off the network.** Embedded and sidecar lyrics are already on disk; a manual edit goes through `save_lyrics`, which asks per edit.
- **`tag` is deliberately not honoured here.** Writing a tag rewrites the audio file, and the file a fetch is about is usually the one playing — that write pauses playback and re-hashes, which is the editor's job on the user's command, never a side effect of a track starting.
- **Never overwrites, and never shadows.** The guard asks `read_sidecar_lyrics`, the waterfall's own reader, rather than probing one exact path: it matches the stem case-insensitively and looks in a sibling `Lyrics/` folder, so on a case-sensitive filesystem we cannot write `Song.lrc` beside a `song.LRC` the user made.
- **A failure is not an error.** A read-only library folder or a NAS that went away is logged at debug; the database row is the source of truth and the lyrics still show.

It hangs off `upsert_lyrics` — the funnel every primary write already passes through, whose own comment says so — plus the plugin bundle path that writes its own transaction, and runs **after** the commit. The editor's write and this one now share `sidecar_path`, so the existence check cannot drift from where the file would go. Five tests against the real migrations (`pool_with_app_schema`), including the two refusals that matter: a `tag` destination leaves the audio file byte-identical, and a differently-spelled sidecar still counts.

**The "Found on the way" in the issue no longer applies**: `onboarding.lyrics.description` already points at *Settings → Images and lyrics* in all seventeen locales, each using its own name for that category. What did need changing is that the copy promised this setting was about lyrics *you edit* — in seventeen locales, and in two of the card titles.

**The clips half of #695 is not in here**, per your call: Canvas and animated covers still live in the app folder. That is the part that needs scanner exclusion (`mp4` is an **audio** extension), a runtime asset-protocol scope per library root, read-only fallback, the end of LRU eviction and a migration — a lot of surface for one PR to share with anything else.

## Window frame: the desktop's, or WaveFlow's (#696)

**Settings → Appearance → Player and window → Window frame.** Default `system`, so nothing changes for anyone who does not go looking.

The request was the appearance choice Chrome offers on Linux, and it does not translate one-to-one: Chrome paints its own chrome from the GTK or Qt theme, while everything inside our window is a web view we already theme. What the desktop owns is the frame. A Qt mode in particular cannot mean Qt widgets — wry is bound to WebKitGTK, there is no Qt backend.

The platforms differ in kind, not in degree, so `preferences.rs` resolves the stored choice into **one instruction** the interface follows, instead of three platform tests in the layout:

| Platform | `app` means | What the frontend draws |
| --- | --- | --- |
| Linux | `set_decorations(false)` | `AppTitleBar` — drag region, title, minimise / maximise / close |
| macOS | `TitleBarStyle::Overlay` + an emptied title | a 28 px drag strip, so the real traffic lights float over our top bar |
| Windows | — | nothing: the option is not offered |

- **macOS keeps the system's buttons.** Dropping the decorations would take the traffic lights with them, and drawing our own is the one thing a macOS user would call *not* native. The overlay style paints the window title over our content and `hiddenTitle` is creation-time only, so the title is what we empty — and restore with the frame; the app is named by the menu bar either way. That style also provides no drag region of its own, which is what the strip is for.
- **Windows is excluded on purpose**: a frame we drew would lose Snap Layouts and the system menu. `supported: false` hides the card, and the resolver refuses `app` there even if the row says otherwise — a database carried over from a Linux install.
- **Applied before the reveal**, while the window is still hidden, so nobody watches the frame they turned off appear and vanish. `set_window_chrome` persists *and* applies in one call: on macOS the style and the title are two calls that must move together.

This closes the "what we can do today" line of [B1](docs/upstream-blockers.md) — as a choice rather than a default, because the GTK3 frame is the *right* answer on a GTK desktop and only looks dated beside libadwaita.

## Checks

`cargo fmt --check`, `cargo check --workspace --all-targets`, `typecheck`, `lint`, `prettier --check`, production build: green. Local review taken to zero over three passes — it caught a Korean verb one jamo from its opposite, a Turkish participle that mixed the singular with the plural suffix, and the case-sensitive sidecar guard above.

**Two things I could not verify:**

- **The macOS branch is compiled nowhere** — not on this machine, not in CI, and the Mac was unreachable. I compiled it here by temporarily removing its `cfg` gate (every call in it exists on all platforms), so the paths, types and syntax are checked; what nobody has seen is how it *looks*.
- **Neither frame has been seen on screen.** The Linux title bar and the macOS traffic lights both want your eyes before this ships.

Closes #696.
Part of #695.

## Summary by CodeRabbit

  - Ajout d’un réglage permettant de choisir entre les décorations système et un cadre dessiné par l’application, disponible sur Linux et macOS.
  - Le cadre personnalisé prend en charge les contrôles de fenêtre, le déplacement et le double-clic pour maximiser ou restaurer la fenêtre.
  - Les paroles récupérées peuvent être exportées vers un fichier sidecar selon la destination configurée, sans écraser les fichiers existants ni modifier automatiquement les tags.

- **Documentation et traductions**
  - Documentation et libellés mis à jour dans les langues prises en charge.


InstaZDLL:
@coderabbitai All five actionable findings applied in 5484323a. Thank you — the window-chrome one was the serious one.

**`apply_window_chrome` swallowing failures** — right, and the failure mode is worse than cosmetic: what the interface draws is decided by what is *stored*, so a stored `app` whose `set_decorations` never landed takes the in-app title bar away from a window that still has no frame of its own. No way to move, resize or close it. It now returns the error, `set_window_chrome` applies **before** it persists, and on macOS the title is set first — it is invisible either way, so a failure there stops before anything the user can see has moved. Startup still only logs: the window is whatever it was created as, and a frame is not worth failing a launch over. I did not add a rollback of the native state: with the apply-then-persist order there is nothing recorded to roll back to, and on Linux there is a single call.

**The empty sidecar** — correct, and it contradicted the rule written two lines above it. `read_sidecar_lyrics` reports an empty or whitespace-only file as a miss, so `fs::write` would have truncated a file we did not create. Both that and the check-then-write window are closed by `OpenOptions::create_new(true)`: the kernel decides, `AlreadyExists` means skip. Test added (`an_empty_sidecar_is_still_someone_else_s_file`).

**Serialising the writes** — added, as a ref guard (it has to hold from inside the click handler, before React re-renders) plus a `busy` flag the card uses to disable the options.

**The four option strings in the search catalog** — added.

**The German subtitle** — you are right about the sense: `überschreibbar` says the *text* can be overwritten, where what is per-edit is the *destination*. Now `Den Speicherort kannst du bei jeder Bearbeitung einzeln festlegen.` That half of the sentence predates this PR, so thank you for catching it on the way through.

One note: the finding on `useWindowChrome.ts:16` arrived with an empty body (severity header only), so there was nothing to act on. If it was about `INITIAL` assuming `supported: false` — that is deliberate: until the backend answers, the card renders nothing rather than offering a choice that may not exist on this platform.

InstaZDLL:
@coderabbitai Second round, in 17f54b3b and 00a4b240.

**The session flag left armed by a retry** — founded, and it was mine: `restore_bounds_and_reveal` retries the reveal up to ten times, so a first attempt that failed and a second that worked left `WINDOW_CHROME_UNAPPLIED` set. The interface would then stop drawing a title bar for a window that really has no frame of its own — the first defect with its sign flipped, and the worse of the two, since it leaves no way to move or close the window. Cleared inside `apply_window_chrome` on any success now, rather than at one caller.

**Restoring the macOS title when the style refuses** — founded. The title goes first precisely so a failure there stops before anything visible has moved; that left the other half unhandled, where the style refuses and the window stays nameless. Previous title captured and put back, the original error preserved. (Verified by compiling the `cfg(target_os = "macos")` branch here with its gate temporarily removed — every call in it exists on all platforms — since nothing else compiles that branch: not this machine, not CI.)

**Reported `app` after a failed apply** — this was your earlier finding and it was right; the fix is the session flag above.

Two I am not applying, with reasons:

- **`role="presentation"` on the drag region.** A `div` has no implicit ARIA role, so the attribute claims nothing and removes nothing; it silences a lint rule this repo does not run. The gestures it carries are mouse-only by nature (drag, double-click), and the action that matters — maximise — is a real `<button>` next to it, in the tab order and labelled.
- **Folding `data-tauri-drag-region` and `onDoubleClick` into one `onMouseDown` with `e.detail === 2`.** The attribute is Tauri's own mechanism and the explicit `startDragging()` beside it is a deliberate belt-and-braces: the attribute only fires when the mousedown target itself carries it, and pointer-events has lost the race against the Windows hit-test before — the mini-player carries the same pair with that comment. Emulating a double click from `detail` would replace a native event with an approximation and drop the platform's own gesture handling.

Also worth noting for calibration: the local CLI gave 0 findings on one pass and 2 on the very next, on the same commit. I read the zero as "this pass found nothing", not "there is nothing".

InstaZDLL:
@coderabbitai Both applied, in 29d27248 and da1ba966.

**A stale initial read overtaking a choice** — founded. Two instances are mounted (the layout draws the bar, the Settings card offers the choice), so a read still in flight when a choice lands would put the old frame back on screen and write it to the cache. A revision ref now, bumped by a write of our own and by a broadcast from the other instance; a read applies only while its revision is still current. Same shape as the counter in `useProfileScopedToggle` (#698), and for the same reason — comparing the *value* is not enough, only "is my answer still the current one" is. The bump sits after the two refusals in `choose`, so a click that changes nothing does not throw away a read that is legitimately in flight.

**Rolling back the frame when the write fails** — founded, and it was the mirror of the order I had already fixed. The apply comes before the write so a frame that refused to change is never recorded as in use; that left the other case open, where the apply worked and the write did not, leaving the window wearing a frame the store knows nothing about while the caller is told the change failed. The previous frame is reapplied on that path. A rollback that itself refuses is logged at `error`: the two states really have parted then, and only a restart — which reads the store — settles it. I did not invent a third "the window has something the store does not" state for that doubly-failed case; it would need two independent failures and a state nothing else reads.

Worth recording for whoever reads this thread later: the local CLI returned **0 findings on one pass and 2 on the next, at the same commit**, several times over this PR. I have been treating a zero as "this pass found nothing", not "there is nothing" — and it was a late pass, after two clean ones, that surfaced the retry leaving the session flag armed, which was the worst defect in the whole change.

InstaZDLL:
@coderabbitai Founded, applied in 594bb957 — and it is the third time on this PR that a fix of mine created the next defect, which is worth naming.

The marker says "the stored choice could not be applied this session", and the read then reports the fallback the window is actually wearing. `apply_window_chrome` clears it on success, which is right when a choice lands and wrong when the call is the rollback after a failed write: that call has just put back the very situation the marker described. The store would hold a choice nobody applied, with nothing saying so, and the interface would draw a title bar for a window that kept the desktop's frame — the defect the marker exists to prevent, reintroduced by the fix for the one before it.

The marker value is now captured before the previous frame is read, and restored when the rollback succeeds. When the initial value was `false` it is left alone, as you said. A rollback that itself fails is unchanged and still logged at `error`: the window wears the new frame, the store holds the old one, no marker describes that, and only a restart settles it — I am not adding a third state for a case that needs two independent failures.

No test for the marker's lifecycle: it needs a window that refuses `set_decorations`, which nothing here can produce. The decision it feeds, `effective_chrome`, is pure and is the part under test.

PR #714, opened 2026-09-21
fix(library): a split reaches the running window, and a 4K clip fits

### LINKED ISSUE #713: splitting an artist does not reach the running window: the panel keeps resolving the deleted id
## What happens

Split a comma-joined artist from **Edit metadata → Split into separate artists** while one of its tracks is playing. The split itself works — the new artist pages exist and are correct — but the running window does not take it into account. The **Now playing** panel's *About the artist* section keeps showing the old credit with *No biography available*, and the library views keep the old entry.

Reloading the window (Ctrl+R) fixes it. In a release build there is no Ctrl+R, so there is no way out short of restarting the app.

Reproduced with `TZUYU, PENIEL` and again with `Ice Spice, Central Cee`.

## Why

Two defects stacked, and fixing either one alone leaves the symptom.

**1. `split_artist` emits nothing.** Every other command that mutates the library announces it — [`edit.rs`](src-tauri/crates/app/src/commands/edit.rs) emits `track:updated` + `library:rescanned` + `player:queue-changed`, and `duplicates.rs`, `library.rs` and the filesystem watcher all emit `library:rescanned`. [`artist_split.rs`](src-tauri/crates/app/src/commands/artist_split.rs) has no `emit` at all, so nothing in the interface learns that anything changed.

**2. The panel resolves an artist that no longer exists.** The split runs `UPDATE track SET primary_artist = ?` and then `DELETE FROM artist WHERE id = ?` on the phantom. The enrichment effect in [`NowPlayingPanel.tsx`](src/components/layout/NowPlayingPanel.tsx) is keyed on `currentTrack.artist_id` — the copy `PlayerContext` holds, which still points at the deleted row. That is the *No biography available*: it is asking about a dead id, not failing to find a biography.

**3. (found while checking) `PlayerContext` never refreshes `currentTrack` from the library.** It only assigns it from `player:track-changed`. The comment in `edit.rs` says emitting `player:queue-changed` makes "the QueuePanel / PlayerBar reflect the new title / artist / album immediately" — no listener refreshes `currentTrack`, so editing the tags of the currently playing track has the same staleness. It is just less visible, because a title usually still reads correctly.

## Fix

- `split_artist` announces itself like every other library mutation.
- `PlayerContext` re-reads the playing track's row when the library changes, guarded for radio and remote tracks, which have no library row. This is what actually clears the symptom, and it fixes the tag-edit case the `edit.rs` comment already claims to cover.

InstaZDLL:
Two things reported in the same session. Grouped because both turn out to be a promise the code had already written down and never kept.

## Splitting an artist changed nothing on screen (#713)

`split_artist` emitted **no event at all** — the only command that rewrites library rows and does not announce it. `edit.rs`, `duplicates.rs`, `library.rs` and the filesystem watcher all emit `library:rescanned`.

That alone would have been a stale list. What made it worse is what the split does: `UPDATE track SET primary_artist = ?`, then `DELETE FROM artist` on the phantom. The Now Playing panel's enrichment is keyed on `currentTrack.artist_id`, so it kept asking about an id that no longer existed — the "no biography available" in the report is not a missing biography, it is a dead id. Reloading the window fixed it, and a release build has no reload.

**`currentTrack` was only ever assigned from `player:track-changed`.** Nothing refreshed it from the library, so every library edit to the playing track left it stale — invisibly for a title, glaringly here. It now re-reads its row on `library:rescanned`:

- skipped for radio and remote-queue tracks, which carry a negative sentinel id and have no library row;
- a generation counter, the shape `radioArtworkTokenRef` already uses here, so two rescans in quick succession cannot let the older read land last;
- object identity rather than the id alone, so a `player:track-changed` arriving in between — carrying engine-side fields fresher than the row — is not rolled back by it;
- a row that has gone keeps the copy we have, which is still what the engine is decoding.

That also delivers what `edit.rs` has claimed since it was written: that emitting `player:queue-changed` makes the player bar show the new title / artist / album at once. Nothing refreshed `currentTrack`, so it did not.

## A hand-picked 4K motion cover was refused

The comment above the manual cap says a deliberately-chosen file *"deserves a more generous limit than an automated fetch"*. Both caps were **64 MiB** — the automated one sized for *"Apple's 1080 H.264 renditions are a few MB"*. The intent was written and never turned into a number, so the feature refused the files it exists to accept.

Manual motion cover, manual Canvas clip and the plugin download all move to **256 MiB**.

**The download cap moves too, and that is the part worth arguing about.** Going over it is not a visible failure: the caller degrades to streaming the remote URL, so a 4K cover played perfectly and could simply never be cached — every time the album was opened, with a `warn!` as the only trace. Raising only the two manual caps would have fixed the reported symptom and left *Save motion covers locally* silently meaningless for exactly the covers people want saved.

**Left alone deliberately:** `DEFAULT_MAX_CACHE_BYTES` stays at 1 GB. A 256 MiB file in a 1 GB LRU is a real tension, but quadrupling how much disk the app takes by default is a separate decision and not one to make inside a bug fix.

## The refusal reads like a size

> file too large: 85.0 MB (max 256 MB)

instead of `file too large (max 67108864 bytes)`. The capped read only ever knows "over the limit" and nothing about how far over, so the real size comes from one `stat` on the path that is already failing — the number is what tells someone whether to re-encode or give up. The image picker in `browse.rs` had the same byte-speak and gets the same treatment.

## Checks

`cargo fmt --check`, `cargo clippy --workspace --all-targets -D warnings` (locally on Windows, where CI never lints), `cargo check`, `typecheck`, `lint`, production build: green. Two unit cases on the size formatter, whose seven expected strings I verified by compiling the function standalone rather than predicting them — app-crate tests only run in the Ubuntu job.

`PlayerContext.tsx` is one of the files that are not prettier-clean on main, so the block was written to prettier's shape by hand and verified against a formatted copy; running the formatter over the file would have reflowed code this change has no business touching.

Local review over three passes. It caught the stale-response window above. One finding was refused with proof: it asked for the `playerSources` import to collapse onto one line for a 120-column config, and `.prettierrc` says `printWidth: 80` — the collapsed line is 82 characters, and prettier's own output is the split form in the diff.

**Not verified on screen**: the split refreshing the panel live, and an 85 MB 4K cover being accepted.

Closes #713.

## Summary by CodeRabbit

  * Les fichiers MP4 d’artwork animé, Canvas et cache peuvent désormais atteindre 256 Mio, y compris pour les rendus 4K.
  * Les couvertures animées manuelles sont stockées dans un espace partagé et restent prioritaires sur celles fournies par les plugins.
  * Après une séparation d’artiste réussie, la bibliothèque et la file de lecture sont automatiquement actualisées.
  * Le morceau en cours se recharge après une nouvelle analyse, sans interrompre les flux distants ou radio.

* **Améliorations**
  * Les erreurs de fichiers trop volumineux indiquent la taille réelle et la limite autorisée dans un format lisible.
  * La documentation des limites et du stockage a été mise à jour.


InstaZDLL:
@coderabbitai Thanks — I've taken the ref-lag seriously and traced it, but I'm not applying the shared-setter refactor. Here is the reasoning, and what I did change instead.

**The lag is real; its only consequence is a skipped refresh, never a stale write.** The window is between `setCurrentTrack(...)` and the effect that mirrors it into `currentTrackRef`. If `library:rescanned` fires inside it, `refresh()` reads the previous track and fetches *its* row — and then the apply is guarded by object identity:

```ts
setCurrentTrack((prev) => (prev === playing ? fresh : prev));
```

By the time it resolves, `prev` is the new track, so `prev === playing` is false and nothing is written. The stale read cannot land. What is lost is the refresh itself — and that window opens only immediately after `player:track-changed`, which has just delivered the track straight from the engine. There is nothing for a library re-read to add at that instant.

**The proposed remedy costs more than the defect.** A shared setter has to be used by all eight `setCurrentTrack` call sites in this file, and it becomes an unwritten rule that the ninth must remember — the kind of invariant that rots quietly, which is precisely the class of bug this PR exists to fix. It would also mean editing regions of `PlayerContext.tsx` unrelated to this change, in a file that is not Prettier-clean on `main`, so the diff would carry reformatting noise.

**What the finding did surface, indirectly, is a real defect I had introduced.** Both new blocks were anchored on `const fetchRadioArtworkInto = useCallback(`, so they were inserted *between* that function and the seven-line comment describing it — the comment ended up documenting my ref sync. Both blocks now sit above it (`efbbc431`).

Two documentation findings from the same pass were founded and are fixed in that commit: the manual motion cover's "256 MiB" read as a cap on the `motion/` directory rather than on each file (the directory has no aggregate ceiling — nothing evicts it), and the Canvas sourcing paragraph still claimed "v1 is manual mp4 only, core-only, no plugin" directly above the section documenting the plugin fallback.

Happy to revisit the setter if you can show a sequence where the identity guard lets a stale row through — that would change the trade-off entirely.

InstaZDLL:
@coderabbitai Verified, and the substance you asked for is already in the file — ten lines below the one you flagged.

`docs/features/ui.md:155` is a paragraph titled **Where the clip is kept (#695)** that documents exactly this: the profile's `canvas/` directory normally, `<library folder>/.waveflow/canvas/` when *Settings → Storage and backups → Folders and caches → Keep clips with the music* is on, reads that look in both with the library first, and the fallback when a library folder refuses the write. Adding that account to line 145 would duplicate it, and two copies of a rule is how one of them goes stale.

What was fair in the finding is that line 145 read as exhaustive: it named `AppPaths::profile_canvas_dir` as *the* destination with no hint that anything else exists. That is now "by default, or next to the music when the user asked for that (see **Where the clip is kept** below)" — `1923ce3c`.

One note in passing: my first attempt linked that cross-reference as `#where-the-clip-is-kept`. That paragraph is bold text, not a heading, so there is no such anchor and the link would have been dead. It is a plain reference instead.

